PR #2, opened 2015-07-29
Improve build system

tskisner:
This changeset replaces the inconsistent makefiles currently in use.  It has a simple top level Makefile which include a platform-specific Makefile snippet that sets the compile options.  If no platform is specified, it defaults to generic linux.  platform files for OSX and HPCPorts included.  All comments welcome.


tskisner:
This update would make it possible to run on any NERSC machine, and since edison is down today Bob was hoping to use hopper...

rncahn:
OK. There's the answer.  So the new makefile should be able to handle all
the fitsio stuff?

rncahn:
so if I just say "make" it will know what to do?

tskisner:
See the updated README for build instructions...

weaverba137:
I notice there is also a product called fiberassign_sqlite.  Does that have similar problems, or is that a completely different code base?

tskisner:
That is some other code that I know nothing about.

-T

PR #16, opened 2016-01-09
Newio

forero:


sbailey:
I tried running this on cori without success - it couldn't ready any targets.  Could @forero update the scripts/ examples to work on cori?  It might be a matter of changing the features files to point to fits file input, or maybe providing the fits files, or?

Also: I merged Bob's script name cleanup PR into master, and then tried merging that into this newio branch.  There was a conflict on pipeline_features.txt .  It looks like both of you moved it to different locations, and perhaps made edits, and we've ended up in a situation where we have copies of the features files in both the top level directory and in the scripts directory.

Suggestion: put *example* features files into etc/, while having the batch scripts in scripts/ reference them from some place else.  Users should make a *copy* of the examples features files and then edit them with their own output directories, etc. rather than having continually updating the git copy to switch back and forth between different user's output directories, whether or not to output fits files, etc.

rncahn:
Stephen,

So would you like me to move all the features files to ./etc  ?  Then we
should update the README.  Also, I should send you an updated version of my
documentation.

Bob

PR #24, opened 2016-02-24
New truth format

### LINKED ISSUE #17: features files should support comments
I tried commenting a line in a features file in order to replace it with a different one, which resulted in a segfault after ~20 minutes of calculations.  e.g. this doesn't work:
```
galFile /project/projectdirs/desi/mocks/preliminary/objects_ss_sf0.rdzipn
tileFile /project/projectdirs/desi/software/edison/desimodel/0.3.1/data/footprin
t/desi-tiles.par
fibFile /project/projectdirs/desi/software/edison/desimodel/0.3.1/data/focalplan
e/fiberpos.txt
# outDir /project/projectdirs/desi/users/rncahn/fa_output/
outDir /global/cscratch1/sd/sjbailey/fiberassign
SStarsfile /project/projectdirs/desi/mocks/preliminary/SStarsfile.txt
SkyFfile  /project/projectdirs/desi/mocks/preliminary/SkyFfile.txt
...
```

1. Features files should support comments
2. If the code can't parse a features file, it should stop immediately rather than leave uninitialized variables or whatever caused the much later crash.  Fail early, fail often.

### LINKED ISSUE #21: Make numerically higher priorities = more important
Change the sense of priority: higher numbers should be higher priority.

(lower numbers = higher priority made sense when the priorities are 1, 2, 3, but gets confusing when they are 2000, 3000, 4000).

### LINKED ISSUE #22: RA,dec output should be double not float
The fiber assignment RA,dec output should use double precision instead of single precision floats.

Note: As of the newio merge, the inputs are double (even though the preliminary input mocks kept them as single precision floats).  This doesn't really matter for mocks, but floats are borderline in terms of accuracy needed for real data calculations.

Verify that internal calculations are done in terms of double too.

sbailey:
This PR refactors the input file formats, with examples at NERSC in /project/projectdirs/desi/mocks/preliminary/mtl/v2/

### File format descriptions ###

**target.fits** has columns:
```
    TARGETID            i8  -- unique target ID
    BRICKNAME           S8  -- brickname from tractor catalogs
    RA                  f8  
    DEC                 f8  
    DESI_TARGET         i8  -- target selection mask bits
    BGS_TARGET          i8  
    MWS_TARGET          i8  
    SUBPRIORITY         f8  -- [0-1] (not yet used)
```
A target file and a truth file (see below) can be generated from the preliminary mock "rdzipn" format using the following (no wrapper; this only has to be done once).
```
from desitarget.targets import desi_mask
from fiberassign.mock import rdzipn2targets

targets, truth = rdzipn2targets('../../objects_ss_sf0.rdzipn')

iisky = (targets['DESI_TARGET'] & desi_mask.SKY) != 0
iistd = (targets['DESI_TARGET'] & desi_mask.STD_FSTAR) != 0
iiobj = (~iisky & ~iistd)
print np.count_nonzero(iisky), np.count_nonzero(iistd), np.count_nonzero(iiobj)

targets[iiobj].write('targets.fits', overwrite=True)
targets[iisky].write('sky.fits', overwrite=True)
targets[iistd].write('stdstars.fits', overwrite=True)
truth[iiobj].write('truth.fits', overwrite=True)
```

This is converted into the **mtl** (merged target list) file using desitarget.mtl.make_mtl() -- another PR for that is upcoming (desitarget branch mtl).  MTL files have the target columns plus:
```
    NUMOBS_MORE         i4  -- number of additional observations requested
    PRIORITY            i8  -- larger priority == more important
    GRAYLAYER           i4  -- can this be observed during gray time?
```
"GRAYLAYER" was formerly known as "LASTPASS" or "Layer 5" even though it is interleaved with all the other dark and bright time layers.  This column will be refactored in a later update.

Standard star and sky fiber files have the same format as the target file.

**MTL + stdstar + sky are the inputs needed for fiberassign**

fiberassign_surveysim also uses a **truth** file, which is row-matched to targets.fits and has quantities like the spectro pipeline z catalog output:
```
    TARGETID            i8
    BRICKNAME           S8  
    RA                  f8  
    DEC                 f8  
    TRUEZ               f4  
    TRUETYPE           S10
    CATEGORY            i4  
```
"CATEGORY" is the old "ID" integer used by fiberassign_surveysim to look up the categories of before/after priorities in the features file.  It is not used by the full mtl+fiberassign+quickcat simulation loop.

#### Option 1 : classic fiberassign_surveysim ####

Use the mtl + stdstar + sky + truth (secret) file as input.  The CATEGORY column of the truth file is used as an index into the kind/type/prio/goal/priopost/goalpost arrays in the input features file.  This maintains compatibility with how fiberassign simulations have been previously run.

#### Option 2 : mtl + fiberassign + quickcat loop ####

This is more work in the short term, but more closely matches how the final data flow will occur.  Iterate:
  * mtl = desitarget.mtl.make_mtl(targets, zcatalog)
  * run fiberassign on that mtl
  * run quickcat to mimic spectro pipeline results:
    * new_zcatalog = quickcat(observed_tile_files, targets, truth, zcat=previous_zcatalog)
    * a pull request from desisim quickcat2 branch will enable this
  * repeat, using the new_zcatalog to update the mtl file...

That is just an overview; I've given @forero a tour of the actual commands and now he will attempt to reproduce this loop as part of reviewing these pull requests.

### Other improvements in this PR ###
  * features files now supports comments with \# ( fixes #17 )
  * RA,dec are now double instead of float ( fixes #22 )
  * numerically larger priorities are now higher priority (fixes #21 )
  * split_targ_secret_sky_std and split_targ_secret executables no longer included
    * if we need to reformat input files, let's do this with simpler python scripts
  * renamed the integer [0-7] "id" to "category" to distinguish it from the targetid use of "id"


sbailey:
I added a commit to this PR to initialize the number of categories and the min/max RA/dec so that you don't have to specify those in the features file if you are running fiberassign .  In particular, fiberassign doesn't (shouldn't!) need the truth category definitions, but it was initializing some dummy arrays using those, and if they weren't defined then it was crashing.  Now it works without needing to define them.

forero:
Tested ```option2``` / ```fiberassing``` as shown [in this python script] (https://github.com/forero/SandboxDESI/blob/master/PR_test/20150226_test.py) together with [this features file](https://github.com/forero/SandboxDESI/blob/master/PR_test/fa_features.txt). It works as I expected.

PR #25, opened 2016-03-07
Development2

rncahn:
This merge will enable fiber assignment to simulate different survey strategies through a file surveyFile, which lists, in order, the tiles to be used.  The basic list of tiles remains static.  The code also will check to use only tiles that are populated with target galaxies.


sbailey:
Please undo the changes to test/template_fiberassign.txt .  That is used by test/test_fiberassign.py to generate a real template file to use for running tests.  It should remain the smaller ra,dec range, use the lite files, have {outfile}, etc.  If you want an example for how to run over the full survey, that should be a different file.

The python scripts you added in py/fiberassign should be in bin/ .  If you would run it from the command line, it goes in bin/; if you "import fiberassign.blah" from the python prompt or from within a script, then it goes in py/fiberassign.

I've only taken a quick look at the C++ code but it looks ok, presuming that you've tested it with both fiberassign and fiberassign_surveysim.

sbailey:
More comments

* default_survey_list.txt and do_stripe_survey_list.txt should not be in py/fiberassign (that is for python code).  Are these redundant with the files in test/*_survey_list.txt now?
* What is the difference between bin/default_survey_tiling_order.py and bin/generate_survey_tiling_order.py ?
* please update your editor settings to not use tabs

rncahn:
On Tue, Mar 8, 2016 at 9:42 AM, sbailey <notifications@github.com> wrote:

> More comments
>
>    - default_survey_list.txt and do_stripe_survey_list.txt should not be
>    in py/fiberassign (that is for python code). Are these redundant with the
>    files in test/*_survey_list.txt now?
>
>
fixed

>
>    - What is the difference between bin/default_survey_tiling_order.py
>    and bin/generate_survey_tiling_order.py ?
>
>
generate renamed to indicate it covers an equatorial stripe 0<dec<10 of
elgs from layer 5 within the first year (or so)

>
>    - please update your editor settings to not use tabs
>
>
trying to address this

sbailey:
I just pushed some typo fixes to test/template_fiberassign.txt .

Running "python test_fiberassign.py" in development2 produces one less output file than master does.  Tile 13125 is missing.  Please track down why that file went missing.

sbailey:
Bob assures me that tile 13125 only had 2 targets assigned in master and those randomly were assigned to other tiles in development2, thus it doesn't get written out.  Will merge in a moment.

PR #26, opened 2016-03-14
Development3

rncahn:
This add some python code in /bin for making plots of the time development of the galaxies observed and figures showing coverage of the sky.  The latter can be made into a gif movie with imagemagick


moustakas:
Awesome!  But note https://github.com/desihub/desispec/issues/65

PR #33, opened 2016-03-28
omp master and die on missing surveyFile

rainwoodman:
I don't yet understand the job partition scheme in global.cpp.
But this shall allow us to compile without openmp in a clean way.

If @rncahn  or someone would like to say a few overall words about how OpenMP is used in the code it would be great.

### DIFF
--- a/platforms/fedora
+++ b/platforms/fedora
@@ -3,10 +3,10 @@
 export CXX := g++
 
 # Normal compile flags
-export CXXFLAGS := -O3
+export CXXFLAGS := -O0 -g
 
 # OpenMP flags (used for both compiling and linking)
-export OMPFLAGS := -fopenmp
+export OMPFLAGS := #-fopenmp
 
 # CFITSIO include and linking commands
 export CFITSIO_CPPFLAGS := -I/usr/include/cfitsio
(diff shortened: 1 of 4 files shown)


rainwoodman:
My bad. The threading is trivial. I missed `#pragma omp for ` below. 
Still, this PR gets rid of the requirement of an openmp enabled compiler.

rncahn:
Open MP is used only in "collect_galaxies_for_all"

it provides list of galaxies available to fiber k on tile j: P[j].av_gals[k]

this is in global.cpp

rainwoodman:
This PR are two usability improvements:

- pragma omp master is the proper way of instructing master-only execution in omp parallel sections. This allows the code to compile without -fopenmp.

- the code shall die instead of writing zero output to minimize surprise factor when surveyFile could not be properly opened.

sbailey:
Reassigning to Ted since he is much more knowledgable about omp, C++ compiling, etc. than I am.

tskisner:
This looks fine.  If you ever need access to the exact thread ID (rather than just checking if master with a pragma), then one would have to have a compile time define (e.g. -DHAVE_OPENMP) that is set in the specific platform file, and then guard the "#include <omp.h>" with a #ifdef HAVE_OPENMP.  Anyway, that is not necessary for this simple use of openmp.

PR #34, opened 2016-03-29
Reproduce

rncahn:
The change in kdTree.h removes a drand48, which was preventing us from getting reproducible results.


sbailey:
Bob now has reproducible results on edison after this change.  We'll merge now and continue testing on edison vs. cori when it is available.

PR #36, opened 2016-04-05
Fixes to use of tileid values read from surveyFile in structs.cpp:read_plate_centers()

apcooper:
surveyFile is supposed to contain a list of tileid values that can be used to filter and sort the list of plates read from tileFile (specifically, only those plates with in_desi=1). The implementation of this in structs.cpp:read_plate_centers() had a couple of array index bugs that were showing up as mysterious segfaults on my machine but could easily pass with invalid values.

Problems fixed here:
- no checks that values were within the ranges corresponding to their use as indices (now throw range_error exceptions).
- related to that, no check that tileids specified in surveyFile were included and flagged with in_desi=1 in tileFile; such cases ended up being mapped to a negative index in the list of plate structs (now check for the negative index).
- implicit use of 1-based tileid as 0-based array index causing (mostly harmless?) off-by-one (now have explicit comments and tileid-1 when used in loops).

I also added a few more comments in read_plate_centers() and moved the definition of the magic number `total_tiles = 28810` to the start of the function for clarity.

### DIFF
--- a/src/global.cpp
+++ b/src/global.cpp
@@ -1006,7 +1006,7 @@ void fa_write (int j, str outdir, const MTL & M, const Plates & P, const PP & pp
     float x_focal[optimal];
     float y_focal[optimal];
     //new
-    float t_priority[optimal];
+    int t_priority[optimal];
     
     std::vector <long long> potentialtargetid;
     
--- a/src/structs.cpp
+++ b/src/structs.cpp
@@ -10,6 +10,7 @@
 #include    <exception>
 #include    <stdexcept>
 #include    <sys/time.h>
+#include    <map>
 #include        <map>
 #include        <stdlib.h>     /* srand, rand */
 #include        "misc.h"
@@ -17,7 +18,6 @@
 #include        "structs.h"
 #include        "collision.h"
 #include        "fitsio.h"
-
 
 
 std::vector<int> count_galaxies(const Gals& G){
@@ -525,11 +525,12 @@ Plates read_plate_centers(const Feat& F) {
     Plates P,PP;
 
     // Maximum possible value of tileid, equal to total number of tiles
-    // (minimum value is 1).
-    int total_tiles = 28810;
+    // (minimum value is 1). No longer used, since tileid is not treated as an
+    // index now.
+    // int total_tiles = 28810;
 
     // read the strategy file
-    // survey_list is list of tiles specified by tileid (starting from 1) in order of survey
+    // survey_list is list of tiles specified by tileid (arbitrary int) in order of survey
     std::ifstream fsurvey(F.surveyFile.c_str());
     if (!fsurvey) {  // An error occurred opening the file.
         std::cerr << "Unable to open file " << F.surveyFile << std::endl;
@@ -569,7 +570,6 @@ Plates read_plate_centers(const Feat& F) {
     int ipass,in_desi,tileid;
     int l = 0;
 
-
     while (fs.eof()==0) {
         getline(fs,buf);
                 if(buf.compare(0, 7, "STRUCT1") != 0) {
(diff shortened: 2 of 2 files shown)


sbailey:
Thanks for these updates, @apcooper .

Please review this @rncahn .  The usage of tileid or tileid-1 as an index is concerning: the tileid from the input files should be treated as an arbitrary integer, not some sequential list that starts at either 0 or 1.  e.g. the current bright time time list starts with larger numbers so that they have different tileids than the dark time survey.  A deeper fix may be needed here.

rncahn:
I'm on it

rncahn:
This doesn't seem right to me.  Let's suppose that the first 50 tiles in
the complete list of 28810 aren't used but the 51st is.  Then
P[0].tileid=51 and invert_tile[50]=0.  But the tile with tileid=50 isn't
used.

apcooper:
Instead of a `std::vector`, how about doing the 'inversion' with `std::map`? This seems to be the only place that tileid is treated as an index. We then can forget about the sanity checks of tileid values (unless we want to impose that for some other reason). Something like:

```c++
#include <map>

    // ....
 
    // Map each valid tileid in order to an index in P[].
    // Tileid is an arbitrary int
    std::map<int,int> invert_tile;
    std::map<int,int>::iterator tileid_to_idx;
    std::pair<std::map<int,int>::iterator,bool> ret;

    for(unsigned i=0;i<P.size();++i)
    {
        ret = invert_tile.insert(std::make_pair(P[i].tileid,i));
        // Check for duplicates (std::map.insert only creates keys, fails on duplicate keys)
        if ( ret.second == false ) {
            std::ostringstream o;
            tileid_to_idx = invert_tile.find(P[i].tileid); // Get the existing entry.
            o << "Duplicate tileid " << P[i].tileid << " in tile file! (Indices " << tileid_to_idx->second << " and " << i << ")";
            throw std::logic_error(o.str().c_str());
        }
    }

    // Create PP, a subset of P containing those tileids specified in the
    // surveyFile, in the order in which they are specified.
    for(unsigned i=0;i<survey_list.size();++i){
        tileid        = survey_list[i];
        tileid_to_idx = invert_tile.find(tileid);

        if (tileid_to_idx == invert_tile.end()){
            // Can end up with no mapping if surveyFile contains valid tileids that have in_desi = 0 in the tileFile.
            std::ostringstream o;
            o << "surveyFile contains tileid " << tileid << ", which is not included (or has in_desi = 0) in tileFile.";
            throw std::range_error(o.str().c_str());
        }

        // Found a valid index, push the tile to the ordered list.
        PP.push_back(P[tileid_to_idx->second]);
    }
```

rncahn:
I'm happy to see the code improved.  Have you checked to see this gives the
same answer in fiberassign_surveysim?

apcooper:
Not yet, no -- I'm only running fiberassign for the tests I'm doing, so I only verified that the new code doesn't change the results there. I'll try and take a look at fiberassign_surveysim tomorrow, but I might not be able to get to it until next week.

rncahn:
After a discussion with Stephen Bailey I now understand better the issue.
Since tileid might not be a number between, say, 0 and 28810, we really
need to use map.  Just go ahead a check that you get the right answer with
fiberassign_surveysim and I will approve the merge.

apcooper:
I've confirmed that the output from fiberassign_surveysim for the test I'm using doesn't change between my commits https://github.com/desihub/fiberassign/pull/36/commits/aa8c0c6407102eb7576b85ad978508b8645d02ab and https://github.com/desihub/fiberassign/pull/36/commits/c525647a27a2c2c68de0c34918fcf1f8f4612a5c.

I've added three further commits to the PR. The first two are cosmetic. The last one fixes a problem in the writing of the FITS output spotted while I was doing the above check -- PRIORITY values in the output tile_xxxxx.FITS were mangled (huge nonsense ints rather than the input priorities). The type of `t_priority` in `global.cpp:fa_write` should have been `int` rather than float.

rncahn:
I believe that the huge integers are correct.  They are random numbers,
assigned to the individual galaxies.  Stephen Bailey should respond.

sbailey:
Priorities are medium-ish integers, e.g 2000, 3000, 3200.  @apcooper's description sounds correct.  I think the huge integers that Bob is remembering are the targetids (random 64-bit integers).

These changes look good to me.  Are we ready to merge?

Issue #38, opened 2016-05-04
Invalid SS and SF input files silently ignored, giving zero SS and SF in output?


apcooper:
I'm confused about how the sky and standard inputs are specified for fiberassign. 

I thought it was via the SStarsfile and SkyFfile paths in the features file -- I've been using some ascii files for this (don't know where these came from, I must have inherited them from an earlier version). Everything seemed to be working OK. However, I've just noticed that these files were not actually being read, and I was unknowingly ending up with no skys or standards. 

The reason for this is that at the top of fiberassign.cpp the files are being read with structs.cpp:read_MTL_file(), which passes through silently when the target path isn't readable by cfitsio.

I have a partial fix for this already that adds some sanity checks for the existence and reading of targFile,  SStarsfile and SkyFfile (the current version runs to completion  even if it can't read the target file -- was that intended?). Can someone please clarify whether or not the SStarsfile and SkyFfile are supposed to be FITS? And if so does anyone know where I can get FITS versions of those files?

rncahn:
Can you send me the features file you are using?

Bob Cahn

apcooper:
Attached. 

[bt_features_apc.txt](https://github.com/desihub/fiberassign/files/249185/bt_features_apc.txt)

rncahn:
Well, I think the SS and SF files should be in fits.  Here is a features
file from master on github:

Targfile /project/projectdirs/desi/mocks/preliminary/mtl/v2/mtl-lite.fits
SStarsfile
/project/projectdirs/desi/mocks/preliminary/mtl/v2/stdstars-lite.fits
SkyFfile  /project/projectdirs/desi/mocks/preliminary/mtl/v2/sky-lite.fits
Secretfile
/project/projectdirs/desi/mocks/preliminary/mtl/v2/truth-lite.fits
surveyFile test/default_survey_list.txt
tileFile
/project/projectdirs/desi/software/edison/desimodel/0.3.1/data/footprint/desi-tiles.par
fibFile
/project/projectdirs/desi/software/edison/desimodel/0.3.1/data/focalplane/fiberpos.txt
outDir {outdir}/

PrintAscii true
PrintFits true
diagnose true

kind QSOLy-a QSOTracer LRG ELG FakeQSO FakeLRG SS SF
type QSO QSO LRG ELG QSO LRG SS SF
prio 3400 3400 3200 3000 3400 3200 0 0
priopost 3500 0 3200 0 0 0 0 0
[...]

PR #39, opened 2016-05-05
Checks on success of reading input MTL

apcooper:
Fixes in respect of #38.

Checks return code from cfitsio read routines and throws an exception on failure. Code will no longer run to completion with missing or unreadable MTL input for targets, skys or standards. This behavior might have been useful for debugging -- it could be reinstated with a parameter file or preprocessor switch if desired.


rncahn:
Looks ok to me.

PR #40, opened 2016-06-07
Development5

rncahn:
Primarily, this just re-incorporates the feature Louis Garrigue developed to produce graphics showing the location of each fiber and all available galaxies for plates at some controllable spacing, e.g. every 1000 plates.  Whether this is done or not is controlled by the features file.


forero:
@sbailey @rncahn I tested that this PR didn't break the fiberassign executable (i.e. [this sequence of commands still works](https://github.com/forero/SandboxDESI/blob/master/PR_test/20160603_make_mock_quick_survey.sh)). 

Are there any other series of standard tests?

sbailey:
I typically run `python test/test_fiberassign.py` but I suspect that your tests are more complete.  Let's go ahead and merge.

PR #41, opened 2016-06-14
Subpriority

rncahn:
This established subpriority as a member of target.  Eventually subpriority will be given for each target by MTL but for now it is artificially created at the time read_MTL is executed.  The subpriority is used only by find_best in the event that two targets have the same priority and the same number of remaining observations.  In this case, the higher priority wins.

Subpriority can also be used by those who want to established a detailed ordering of their targets, say in the MWS.

### DIFF
--- a/src/feat.cpp
+++ b/src/feat.cpp
@@ -107,8 +107,8 @@ void Feat::readInputFile(const char file[]) {
                 init_types();
                 init_ids_types();
             }
-            if (tok[0]=="prio")  for (int i=0; i<Categories; i++){prio.push_back(s2i(tok[i+1]));}
-            if (tok[0]=="priopost") for (int i=0; i<Categories; i++) priopost.push_back(s2i(tok[i+1]));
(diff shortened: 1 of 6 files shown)


forero:
```Q.subpriority= rand();``` should only be used if that field does not exist in the input MTL.

rncahn:
Right.  This is just a temporary version to get the functionality.  I need
to have you or Stephen fix the fits part eventually and regenerate the
files of mocks to include fixed subpriorities.

I realize I forgot to include a srand to make sure we get reproducible
results.

forero:
Right. The FITS and the srand() part should be there before merging the PR.
The scripts to generate the mocks already include the  subpriority
https://github.com/desihub/desitarget/blob/master/py/desitarget/mock/darktime.py#L260

rncahn:
Jaime,
 Good!  So did you already run it and generate new mocks with subpriority?

Bob

forero:
I did. But fiberassign currently does not use subpriority, so the data is there in the mocks but not used.

PR #46, opened 2016-07-27
Obsconditions

### LINKED ISSUE #45: Use observational conditions to differentiate targets on a tile by tile basis
Every tile has tag associated to it indicating the `dark`, `bright` or `gray` conditions for observations. This information is not used in fiberassign. We need to match the input tiles with the MTL file to enable that distinction based on observing conditions.

forero:
This fixes #45. 

The code now reads the column `obsconditions` from both the tile file (in FITS format) and the mtl to check whether the targets matches the program.

The critical line is here: https://github.com/desihub/fiberassign/blob/5fbc9c0722bdd2d86a776f14fe942113a78fb836/src/global.cpp#L55

Is it the right place to make the comparison between tile and mtl? 

I made simple tests with `quicksurvey` using dark time mocks as an input and checking that indeed only the `DARK` tiles have any targets allocated.


sbailey:
Looks good to me.  Bob isn't at the lab today but maybe watching email; let's give him until at least after lunch to comment.

PR #50, opened 2016-10-05
Outputs targetid (not index) for potential targets

apcooper:
Outputs the targetid value read from MTL rather than the internal index, for potential targetids. This now matches what happens with the targetids for the targets actually assigned to fibers.

(the script in fiberassign/bin to map the FA tile files back to the mocks is broken without this fix).

### DIFF
--- a/src/global.cpp
+++ b/src/global.cpp
@@ -1075,12 +1075,14 @@ void fa_write (int j, str outdir, const MTL & M, const Plates & P, const PP & pp
                     mws_target[i] = M[g].mws_target;
 		    strncpy(brickname[i], M[g].brickname, bricklen+1);
                 }
-                
+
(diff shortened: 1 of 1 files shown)


rncahn:
Running the new branch, I still find that a two-epoch run of fiberassign
fails because the files giving the assignments of each fiber for a
particular tile do not have the correct TARGETID, but instead have values
just up to the number of galaxies observed.

apcooper:
Hi Bob,

Right, this PR is a smaller fix that just makes the output of the potential targetids the same as that of the targetids assigned to fibres. It sounds like the problem you have is more fundamental and something to do with the interaction between changes to the assignment of targetids and multi-epoch runs.

I can look into this if you could please post (perhaps in a separate github issue) the actual output at the point of failure and the features file you're using.

Issue #58, opened 2017-01-03
sky/stds should steal from lowest not highest priority targets


sbailey:
Reported by Bob:

If there are not enough spare fibers to fill in standard stars and sky locations, fiberassign will steal back fibers from science targets to achieve a minimum density.  It should start with the lowest priority science targets.  It used to do this, but then we switched whether numerically higher priorities are more important or less important and this piece of code didn't get updated, thus sky/std are currently stealing from most important targets instead of least important.  Oops.

Higher numerical priority = more important.  Sky/Stds should steal from lowest numerical priority = lowest important science targets (and even then, only if needed).

PR #60, opened 2017-01-10
fixed priority ordering

rncahn:
This request fixes a bug that was created when we switched from making low numbers have high priority to the current state where high numbers have high priority.


sbailey:
fixes #58; merging now.

PR #61, opened 2017-01-10
Sort by subpriority

rncahn:
This request makes sure that whenever when consider selecting or replacing galaxies, we consider the galaxies in order of subpriority.  Previously, in some instances we used the first galaxy in a list that satisfied some criteria.  That list might not be independent of all possible means of reaching that point.  In particular, if we save the kd-tree results for use in later epochs we won't get the same result as if we recompute the kd-tree for every epoch if we don't take care to order each collection of considered galaxies in a fixed fashion.  The subpriority enables us to do this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,6 @@ bin/fiberassign_surveysim
 bin/split_*
 slurm*
 *.pyc
+script/*
+test/*
+*diff*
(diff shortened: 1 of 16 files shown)


sbailey:
The concept sounds good to me, but I haven't reviewed the code in detail.  Can you take a look, @forero ?  Thanks.

sbailey:
For the record: The latest cleanup PR generated some conflicts with master; Bob and I resolved these via the new github resolve conflicts webpage interface.  Bob will now grab this branch, reconfirm that it compiles and runs, and then we'll merge this branch.

PR #66, opened 2017-03-23
Fix fiber positioner

rncahn:
Previously, the x,y coordinates of the positioners and their assignments to spectrometers was given in lists with the order determined by the fiberpos file from which they were read.  That file also contains a column fiber, which wasn't used.  This version makes sure that the lists are order by the values in "fiber".


sbailey:
Documenting an analog conversation with Bob: he tested this by using two different fiberpos files with a different order of the entries and confirmed that the fiberassign output was identical.  Merging now.

PR #67, opened 2017-03-31
Remake fiber positioner

rncahn:
This version redesigns the fiber-positioner aspect of the code, making a class fpos for fiber-positioners and replacing the old PP class with FP as a vector of fpos.  Previously the coordinates of the positioners and the spectrometer assignments were simply vectors in the order the fiber-positioners were read-in.  This new structure will facilitate developments like accounting for broken fibers and stuck positioners.


sbailey:
I'll merge this now so that #69 can be debugged with this reorganization rather than generating merge conflicts.  A post-merge request though: remove `doc/Guide_to_FiberAssignment.tex.old`.  git history already has the old version (tex), as does DocDB (pdf).  We don't need to separately version it by hand under a different name.

Issue #69, opened 2017-03-31
fiberassign 0.4.0 dies if a target is both a stdstar and a science target


sbailey:
Using example files in on edison in `/scratch2/scratchdirs/sjbailey/desi/debug/fiberassign`, fiberassign/0.3.2 works fine but fiberassign/0.4.0 dies with
```
fiberassign fiberassign-params.txt
...
# map position in target list to immutable targetid at 0.0974 s
terminate called after throwing an instance of 'std::logic_error'
  what():  Duplicate targetid 4252430163699209429 in MTL
Aborted
```
The target appears in both standards-dark.fits and mtl.fits (once per file) since it is both a standard star and a requested MWS science target.  This is a perfectly legal case and should be allowed by fiberassign.

sbailey:
Note: this problem also applies to master and the remake_fiber_positioner branch in PR #67

sbailey:
For the record: in this sub-patch of sky, 100% of the dark time standard stars also appear as MWS targets, i.e. this is not a rare occurrence.

sbailey:
fixed in PR #72

PR #71, opened 2017-04-03
sort fibers by true number rather than position in list

rncahn:
Now that the fibers have become a class, we want to sort the collection of fibers by the true number of the fiber, not the position in the list of the fibers.


sbailey:
This defines a `sort_by_fiber_number` function but appears to not actually use it.  The actual sorting is done separately in `structs.cpp` function `read_fiber_positions`.  It also has some leftover debugging printouts to cleanup unless you think it is important to keep them for the log.

rncahn:
Yes.  Will do.  Also I am trying to solve your problem by moving the check
on duplicates to apply only to mtl, not SS.

rncahn:
How do I correct the pull request to make your changes?

sbailey:
> How do I correct the pull request to make your changes?

Update the code in your branch and push again to github.  Since the PR is associated with the sort_by_fiber_number branch, any updates on github of that branch automatically apply to the PR.

rncahn:
done

once you are happy with this, we can try my fix for the duplication issue.
Should I just push a new branch without asking for a PR so you can try out
the fix?

sbailey:
Merging now.

It's always ok to push new branches to github to share work in progress.  Please start a new branch off of master for fixing the TARGETID duplication issue #69 .

rncahn:
you want it off master rather than off the branch I sent for PR?

PR #72, opened 2017-04-03
move check for duplicate targets to apply only to mtl

rncahn:
Previously, problems arose if some Milky Way Survey (MWS) galaxies were also used as standard stars. To overcome the problem, we check for duplication of targets only in the mtl file, not using SS and SF, too.

### DIFF
--- a/src/fiberassign.cpp
+++ b/src/fiberassign.cpp
@@ -48,14 +48,10 @@ int main(int argc, char **argv) {
     M=Targ;
     printf(" Target size %d \n",M.size());
     //need to be able to match immutable target id to position in list
+    //check for duplicates on mtl only to allow duplication with SS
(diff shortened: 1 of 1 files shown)


sbailey:
fixes #69 

I tested on the example listed there to confirm it works in the sense that
1. it didn't crash
2. it assigned a similar number of targets to fibers (just over 80%, dominated by ELGs)
3. it did continue to assign standard stars

However, it isn't clear to me why this fix works.  You now add stdstars and sky to the targets list `M` after inverting it instead of before.  How is the inverted target list `invert_target` used?  Why isn't it problematic that standards and sky aren't in that list anymore?  Or were they extraneous before since they get assigned as special cases anyway?

rncahn:
I think the inversion isn't actually used.

PR #75, opened 2017-07-15
fix uninitialized F.Npetal

sbailey:
This PR fixes a bug where `read_fiber_positions(F)` was using `F.Npetal` to resize a vector prior to `F.Npetal = 10` being initilized.

I also converted some tabs to spaces while trying to read the code for debugging this.


sbailey:
fiberassign/master updated on cori with this version.  Will tag and install after rest of end-to-end is verified.

PR #84, opened 2017-10-16
Changes to fiber ordering in simple_assign

JianhuaHe:
Uses priority and subpriority to determine fiber assignment order rather than predefined fibre index.
This is a fix to the problem found and discussed at
https://desi.lbl.gov/trac/wiki/BrightGalaxyWG/TeleconNotes2/FibreAssignment

### DIFF
--- a/src/global.cpp
+++ b/src/global.cpp
@@ -280,6 +280,9 @@ inline int improve_fiber(int jused_begin, int jused, int k, MTL& M, Plates& P, c
 }
 // Assignment functions ------------------------------------------------------------------------------------------
 // Assign fibers naively
+bool inverse_pairCompare(const std::pair<double, int>& firstElem, const std::pair<double, int>& secondElem) {
+    return firstElem.first > secondElem.first;//used to sort galaxies by subpriority
+}
 
 void simple_assign(MTL &M, Plates& P, const FP& pp, const Feat& F, Assignment& A) {
     Time t;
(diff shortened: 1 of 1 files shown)


sbailey:
Thanks for contributing this.

When a high priority target is covered by two fibers, then the fiber number breaks the tie and the lower numbered fiber is assigned first.  i.e. the fiber number still has some impact, but it is less than before.  Correct?

I confirmed that running this works, and results in slightly higher QSO placements (high priority) and slightly fewer ELG placements (lower priority).  The original scheme of assigning in order of fiber number still considered the highest priority target for each fiber so I'm not sure how much of this change in fractions is due to guaranteeing that the highest priority targets are placed first.

JianhuaHe:
Yes, in the case that a high priority target in overlaps of two fibres then the fiber number is used to break the tie.  The fiber number still has some effects but it is in a deterministic way.   

The gain of this scheme is that it always places a fibre on a high priority target  if it collides with a low priority target. The scheme can guarantee that no high priority object is lost due to fibre collisions. A low priority target therefore has less or even no effect on high priority targets. The scheme gives a more clear rule of treatment of fibre collisions.

sbailey:
Thanks for the explanation.  Indeed, collisions are the case where the former code could have a lower-priority target on a lower-numbered fiber inadvertently block a higher-priority target on a higher-numbered target.  Looks good!

PR #86, opened 2017-11-02
update license; add readme

### LINKED ISSUE #85: Why does fiberassign have a BSD-2-clause license?
The majority of other DESI packages use a BSD-3-clause license.

weaverba137:
This PR fixes #85, by updating the LICENSE file to the DESI standard.  It also restores the missing README file.


sbailey:
Having a README is good, but the one you restored is completely outdated (which is probably why it was removed).  i.e. please don't merge until we get someone to write a correct README.

weaverba137:
OK, fair enough.  Who would be the best person to assign this to to update the documentation?

sbailey:
I've updated the README.rst to be minimalist but at least correct for how to compile the code and pointing to an example configuration file and further reading.

PR #101, opened 2018-02-25
Inputs information from BROKEN or STUCK fibers

forero:
The information about BROKEN and STUCK fibers can be read into fiberassign. I am leaving this PR to discuss what to do with this information.

### DIFF
--- a/bin/qa-fiberassign
+++ b/bin/qa-fiberassign
@@ -13,7 +13,7 @@ import numpy as np
 from astropy.table import Table, vstack
 from astropy.io import fits
 
-from desitarget import desi_mask
+from desitarget.targetmask import desi_mask
 import desimodel.io
 
 parser = argparse.ArgumentParser()
@@ -88,6 +88,9 @@ for filename in args.tilefiles:
         for err in errors:
             print('  {}'.format(err))
 
+    n_stuck = np.count_nonzero(fa['TARGETID']==-2)
+    n_broken = np.count_nonzero(fa['TARGETID']==-3)
+    print('{} stuck fibers\n{} broken fibers'.format(n_stuck, n_broken))
 
 print('{}/{} tiles had errors'.format(nfail, len(args.tilefiles)))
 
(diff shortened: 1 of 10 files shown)


sbailey:
* BROKEN: set DESI_TARGET bit NOTARGET (still needs to be added to desitarget, and used for unassigned fibers too), and put x,y = center of positioner (TDB, we might choose to put it 1mm offset of center, but the main point is to get it out of the way of colliding with anything)
* STUCK: set DESI_TARGET bit NOTARGET and leave it at the x,y where it is stuck, and make sure that no other assignments will run into that.

This PR looks like it isn't yet parsing the START_DATE / END_DATE columns to know if it is currently broken or stuck.  The file could have a history of problems that have since been fixed and it is now ok to assign the fiber even if it wasn't ok a year ago.

forero:
Right. Dates are not parsed. I didn't know how to use them. Mostly because it's not clear to me what is the meaning of "currently" inside fiberassign. Should it simply be the date when the code is run? Should it be a predefined input date? Should it be the planned observing date for each tile?

sbailey:
For normal operations usage, the date would be when the code is run.  For use in simulations, we would need an option to override that date to some other arbitrary date.  This could be a pain to do in C++ so we could consider a python wrapper to prepare the inputs and then call C++ to do the algorithmically intensive calculations.

forero:
OK. I will try to implement this.

forero:
@rncahn where in the code is the best place to check whether a fiber is broken/stuck and avoid making any assignment? Elements in FibPos have now a stuck/broken flag setup at the beginning that allow for an explicit check `if(FibPos[i].broken)` .

forero:
@sbailey @rncahn 
I think this is ready for review. 
The most important part is here
https://github.com/desihub/fiberassign/blob/4500491a9459be4c19453566388c833a24013c15/src/global.cpp#L144
That's the only place where the logic of a broken/stuck fiber impacts the assignment. All the other changes are bookkeeping.

forero:
I am merging now. The biggest change, from the interface point of view, is that the input file now requires a the fiberstatus filename; such a sample file is include under `test/`.

Issue #109, opened 2018-03-23
unassigned fibers RA, dec, x, y


sbailey:
Unassigned, stuck, and broken fibers should still set meaningful RA, DEC, X, and Y quantities rather than NaN.  This will require implementing the conversion from (x,y + telra,teldec) -> (RA,DEC)

desitarget will soon define a NO_TARGET bit which should be set for DESI_TARGET for these fibers.

A new FIBERMASK column should be included in the output, setting bits for broken and stuck fibers (instead of resetting DESI_TARGET bits as it currently does).  See desihub/desispec#485.

forero:
solved in #112

PR #112, opened 2018-04-24
Fibermask and  RA+dec for unassigned, stuck, and broken fibers

forero:
This PR solves #109

### DIFF
--- a/src/global.cpp
+++ b/src/global.cpp
@@ -727,7 +727,7 @@ void fa_write (int j, str outdir, const MTL & M, const Plates & P, const FP & pp
     // string arrays, since the CFITSIO API requires non-const pointers
     // to them (i.e. arrays of literals won't work).
     
-    size_t ncols = 13;
+    size_t ncols = 14;
     
     char ** ttype;
     char ** tform;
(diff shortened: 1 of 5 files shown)


src/global.cpp
```diff
@@ -884,40 +890,33 @@ void fa_write (int j, str outdir, const MTL & M, const Plates & P, const FP & pp
...
-                        target_id[i]=-1;
-                    }
+                    target_id[i]=-1;
                 }
 
                 if (g < 0) {
                     if(pp[fib].stuck){
-                        ra[i] = qNan;
-                        dec[i] = qNan;
+                        fibermask[i] = 2;
```
sbailey: Please move the fibermask magic numbers 0/2/4 into constants defined in a header file to make it easier to update them later if needed.

src/global.cpp
```diff
@@ -926,11 +925,24 @@ void fa_write (int j, str outdir, const MTL & M, const Plates & P, const FP & pp
...
+                    fibermask[i] = 0;
                     ra[i] = M[g].ra;
                     dec[i] = M[g].dec;
                     dpair proj = projection(g,j,M,P);
                     x_focal[i] = proj.f;
                     y_focal[i] = proj.s;
+                    
+                    //testing that xy2radec is working with good precision.
+                    xy2radec(&testra, &testdec, tilera, tiledec, x_focal[i], y_focal[i]);
+                    if(fabs(testra-ra[i])>(1.0/3600.0)){//arcsecond precision
```
sbailey: Please check at a precision of 0.01 arcsec instead of 1 arcsec for both ra and dec.  In the RA direction, please include the cos(dec_radians) term so that you are testing a real separation on the sky and not just delta(ra).

src/global.cpp
forero: Done.

src/global.cpp
forero: That implementation was producing offsets on the order of 0.5 arcseconds. I had to rewrite the implementation for radec2xy in order to have consistent transformations radec2xy<->xy2radec at machine precision.

forero:
@sbailey This seems to be ready. Please review again.

Issue #64, opened 2017-03-13
standard star DESI_TARGET mask in features file


sbailey:
Add an option to the features (parameters) file for what DESI_TARGET mask bits should be considered standard stars.  Two use cases:
* Dark and bright time have different but heavily overlapping sets of standard stars.  This mask would
allow us to use the same input catalog for both while selecting which subset should be used for a given fiber assignment run
* We'd like to simulate WD standards along with the other standards when generating mock inputs, but the spectro pipeline isn't ready yet to accept them.  This mask would allow us to include them in the standards file while temporarily turning them off until later when the spectro pipeline is ready.

Assigning to @forero

PR #114, opened 2018-05-15
standard star DESI_TARGET mask as input option 

forero:
This solves #64 by introducing the command line option `--starbit` to select the bit in the `desi_target` mask that signals a standard star. The default option is `33`.


forero:
@sbailey I've implemented the `starmask`. The match with `OBSCONDITIONS` was already in place. I think this is ready to merge.

forero:
I am going to merge this now. Basic tests confirm that `starmask` works as expected.

Issue #115, opened 2018-05-25
python wrapper for fiberassign


sbailey:
Make a python wrapper for fiberassign to simplify the FITS I/O logic and coordinating science fiber assignments, sky monitor fiber assignments, adding GFA targets, and adding row-matched target selection targets.  This could be as "simple" as spawning the C++ fiberassign code and then parsing its outputs; it doesn't have to be a full ctypes wrapper etc.

This code could also make it easier to do things like have default values for the `--footprint` and `--positioners` options (get them from `$DESIMODEL`).  It could also address #77 by pre-checking that input files exist before running the C++ code.

sbailey:
Also related to #99 to run fiberassign on arbitrary pointings.  This wrapper could take optional `--telra TELRA --teldec TELDEC --tileid TILEID` options that would be used to create a temporary tile definition file before running fiber assignment for that tile.

Issue #99, opened 2018-02-07
fiberassign on arbitrary pointings


forero:
We need a wrapper that works like "single_pointing_fiberassign --targets targets.fits --telra 15.2 --teldec 13.3 -o blat.fits" that would

* run MTL to convert targets.fits -> mtl.fits
* generate custom tile.fits, features file, etc. and run fiberassign 
* run GFA targets
* merge those together into a multi-HDU FITS file

sbailey:
Most of this was done by PR #115, but let's leave this issue open until the GFA targets have been included in that wrapper too (as an option, not a requirement).

sbailey:
PR #122 added GFA targets; closing this ticket.

PR #116, opened 2018-06-06
Python wrapper

forero:
First implementation of a python wrapper around `fiberassign`. Solves #115.

### DIFF
--- a/bin/run-fiberassign
+++ b/bin/run-fiberassign
@@ -1,0 +1,69 @@
+#!/usr/bin/env python
+
+"""
+Script handling fiberassign inputs
+"""
+
+import sys, os
+import argparse
+import desimodel.io
+import numpy as np
+import datetime
+from desitarget.targetmask import desi_mask
+
+
+parser = argparse.ArgumentParser()
+parser.add_argument("--mtl", type=str,  help="input targets (FITS file)", required=True)
+parser.add_argument("--sky", type=str,  help="input sky positions (FITS file)", required=True)
+parser.add_argument("--stdstar", type=str,  help="input std stars (FITS file)", required=True)
+parser.add_argument("--fibstatusfile", type=str,  help="list of positioners and its status (ECSV file)", required=True)
+parser.add_argument("--footprint", type=str,  help="list of tiles defining the footprint (FITS file)")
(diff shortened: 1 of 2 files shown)


bin/run-fiberassign
```diff
@@ -0,0 +1,69 @@
...
+parser.add_argument("--stdstar", type=str,  help="input std stars (FITS file)", required=True)
+parser.add_argument("--fibstatusfile", type=str,  help="list of positioners and its status (ECSV file)", required=True)
+parser.add_argument("--footprint", type=str,  help="list of tiles defining the footprint (FITS file)")
+parser.add_argument("--positioners", type=str,  help="list of positioners on the focal plane (text file)")
+
+parser.add_argument("--surveytiles", type=str,  help="set of tiles to run fiberassign on (text file)")
+
+parser.add_argument("--telra", type=float, help="Right Ascension of arbitrary pointing - overrides --surveytiles")
+parser.add_argument("--teldec", type=float, help="Declination of arbitrary pointing - overrides --surveytiles")
+parser.add_argument("--telID", type=int, help="Integer ID of arbitrary pointing - overrides --surveytiles")
```
sbailey: It looks like these options aren't used yet; let's comment them out until they are implemented.

`--telID` should be `--tileid`

src/Makefile
```diff
@@ -30,7 +30,7 @@ misc.o \
 structs.o \
 version.o
 
-EXECUTABLES = fiberassign
+EXECUTABLES = fiberassign_exec
```
sbailey: For consideration:  There are enough science collaboration fiberassign users out there that this could create confusion by renaming `fiberassign` to `fiberassign_exec` while having the python wrapper be called `run-fiberassign`.  If we are going to rename anything, I'd be inclined to make the python script be `fiberassign` and rename the C++ executable.  i.e. leave `fiberassign` as the name of the thing that the end-user would run.

forero: I keep `fiberassign` for the python wrapper and leave `fiberassign_exec` for the C++ executable.

bin/run-fiberassign
forero: This is fixed now.

forero:
@sbailey I've updated the python wrapper to run on arbitrary pointings (thus solving #99). Please review it again.

forero:
Thanks!

Issue #95, opened 2017-11-29
Sky Camera positions


forero:
There will be 20 positioners on the periphery of the 5000 fibers. Their purpose is measuring sky background and must be then assigned to blank sky positions. `fiberassign` must also find and report these sky location for every tile.

sbailey:
The locations of these positioners are in `$DESIMODEL/data/focalplane/fiberpos-all.fits` where `DEVICE_TYPE=='ETC'` (ETC=Exposure Time Calculator since that is what these positioners are used for).  Currently `desimodel.io.load_fiberpos()` only returns the 5000 normal positioner locations, but it could be extended with an option to read this file instead and return the locations of the requested positioner type.  Definitions coming from [DESI-0530](https://desi.lbl.gov/DocDB/cgi-bin/private/ShowDocument?docid=530):
* POS = normal positioners, which could be assigned to science, sky, or stdstars
* ETC = Exposure Time Calculator sky fiber positioners
* FIF = Fiber Illuminated Fiducials
* GIF = GFA Illuminated Fiducials
* OPT = location of a hold but potentially nothing in it
* NON = defined location, but no hole or device there (!)

forero:
Solved by #119

PR #119, opened 2018-06-12
Compute and report sky location (exposure time calculator) for every tile.

forero:
Solves #95. Everything is done within the python wrapper by running `fiberassign_exec` twice.
The final result is a tile file with three tables: `FIBERASSIGN` (5000 fibers), `POTENTIAL` (5000 fibers) and `SKYETC` (20 fibers).

### DIFF
--- a/bin/fiberassign
+++ b/bin/fiberassign
@@ -66,21 +66,44 @@ if not os.path.isdir(args.outdir):
     os.makedirs(args.outdir)
 
 # Check if output files already exist
-existing_files = False
-for line in open(args.surveytiles):
-    line = line.strip()
-    if line.startswith('#') or len(line)<2:
-        continue
-    tileid = int(line)
-    tilefile = os.path.join(args.outdir, 'tile_{:05d}.fits'.format(tileid))
-    if os.path.exists(tilefile):
-        existing_files = True
-        print('ERROR: {} already exists'.format(tilefile))
+def check_existing_files(checkdir):
+    existing_files = False
+    for line in open(args.surveytiles):
+        line = line.strip()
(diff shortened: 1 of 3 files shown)


sbailey:
The following test command works for master but produces no output for this sky branch.  What's missing?
```
basedir=/project/projectdirs/desi/datachallenge/reference_runs/18.3
fiberassign --mtl $basedir/targets/mtl.fits \
    --sky $basedir/targets/sky.fits \
    --stdstar $basedir/targets/standards-dark.fits \
    --fibstatusfile $basedir/fiberassign/fiberstatus.ecsv \
    --surveytiles $basedir/fiberassign/dark-tiles.txt \
    --outdir $SCRATCH/temp/
```

And another reason to use something like `tempfile.mkdtemp`: the user might be running fiberassign in a directory where they don't have write permission and the current branch fails to be able to write its temporary files there.

forero:
@sbailey I have included the `tempfile.mkdtemp` calls.

The problem you found comes from the fact that the `sky.fits` file you use as an input doesn't have a `NUMOBS` column, so `NUMOBS` is set to zero when it's read by `fiberassign`. When I read that file as the target file to be assigned for the `sky` fibers, then nothing gets assigned and the `tile_*.fits` file for those `sky` fibers are not written down. The absence of those files halts the computation of the `tile_*.fits` with the extended `SKYETC` table. I will fix that by setting `NUMOBS` to one if that column is not included in the input file.

forero:
@sbailey Things seem to be working now. Please check again.

sbailey:
Changes look good.  Thanks.  I verified that it works from a directory in which I don't have write permission and that the previously failing example now works.  Merging.

Issue #125, opened 2018-06-22
fiberassign is only processing one out of six dark tiles


moustakas:
I'm running fiber assignment with a non-standard tile file and only one out of six input tiles are being processed by fiber assignment.  I'm certain I'm doing something wrong (probably with my input tile file) but I'm not sure what to change.  

To reproduce the issue you'll need to unpack this tarball in any empty directory:
http://www.sos.siena.edu/~jmoustakas/tmp/test-fiberassign.tar.gz
and then run
```
fiberassign  --mtl mtl.fits --stdstar standards-dark.fits --sky sky.fits \
  --surveytiles dark-tiles.txt --footprint tiles.fits \
  --fibstatusfile fiberstatus.ecsv --outdir .
```

There are six input tiles, tile numbers `10000-100005` and everything runs successfully.  However, the only resulting output file is `'tile_10000.fits`. 

@sbailey @forero @rncahn your thoughts would be welcome.

Here's the full log:
```
# Read target, SS, SF files at 6.7e-05 s
star mask 60129542144
Finding file: standards-dark.fits
Found MTL input file: standards-dark.fits
Reading MTL input file standards-dark.fits
NUMOBS_MORE not found ... setting to 1
PRIORITY not found ... setting to 1
Keeping 2404 targets within ra/dec ranges
star mask 0
Finding file: sky.fits
Found MTL input file: sky.fits
Reading MTL input file sky.fits
NUMOBS_MORE not found ... setting to 1
PRIORITY not found ... setting to 1
Keeping 94208 targets within ra/dec ranges
star mask 0
Finding file: mtl.fits
Found MTL input file: mtl.fits
Reading MTL input file mtl.fits
Keeping 19217 targets within ra/dec ranges
# ...read targets  took : 0.08 s
 Target size 19217 
 Standard Star size 21621 
 Sky Fiber size 115829 
# map position in target list to immutable targetid at 0.0958 s
# assign priority classes at 0.0959 s
  class  0  number  5116
  class  1  number  3399
  class  2  number  13106
# ...priority list took : 0.0017 s
# Start positioners at 0.0976 s
before reading positioners 
read the positioner file
 sorted by fiber number
 i 0 FibPos[i].fib_num 0 
 i 1 FibPos[i].fib_num 1 
 i 2 FibPos[i].fib_num 2 
 i 3 FibPos[i].fib_num 3 
 i 4 FibPos[i].fib_num 4 
 i 5 FibPos[i].fib_num 5 
[...]

PR #126, opened 2018-06-22
do not assume tileid is 5 digits long

moustakas:
The solution is not particularly elegant, but it addresses #125.  Instead of assuming every tileid is 5 digits I assume that every file is a `.fits` file and trim the suffix from the filename to get the tileid.

### DIFF
--- a/bin/fiberassign
+++ b/bin/fiberassign
@@ -155,7 +155,7 @@ sky_tile_id = {}
 for sky_file in skys:
     f = sky_file.split('/')[-1]
     fileid = f.split("_")[-1]
-    fileid = fileid[0:5]
+    fileid = fileid[:-5]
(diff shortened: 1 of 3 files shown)


moustakas:
I also took the opportunity in this PR to update `bin/qa-fiberassign` with the new FITS extension names (`FIBERASSIGN_FIBERASSIGNMENTS` --> `FIBERASSIGN` and `POTENTIAL_ASSIGNMENTS` --> `POTENTIAL`) being written by `bin/fiberassign`.

sbailey:
Still fragile, but better than that was in there.  OK to merge for now, but let's revisit this in the next round of pending file naming updates, and use functions like `os.path.split` and `os.path.splitext` to do the parsing.

Issue #118, opened 2018-06-08
sign flip in (RA,dec) -> (x,y)


sbailey:
fiberassign has a sign flip in the transformation from (RA,dec) -> (x,y).

The transforms should be:
* Larger RA -> smaller (more negative) X (fiberassign has the opposite)
* Larger DEC -> larger (more positive) Y (fiberassign is correct here)

See [DESI-0481](https://desi.lbl.gov/DocDB/cgi-bin/private/ShowDocument?docid=481) and [DESI-0742](https://desi.lbl.gov/DocDB/cgi-bin/private/ShowDocument?docid=742) for background material on DESI coordinates.

Short (though perhaps not illuminating) explanation: when facing south and looking up from the primary mirror towards the focal plane and the sky, RA increases to the left (east) and dec increases up (north).  In this orientation the focal plane is "upside-down" with X increasing to the left (east) and Y increasing down (south).  BUT: the light bounces off a mirror on its way from the sky to the focal plane, thus flipping signs:  eastern targets with larger RA toward the left of the sky end up on the right hand side of the focal plane, i.e. smaller (more negative) X.  And more northern targets (larger DEC) end up at the lower (larger Y) portion of the focal plane.  It helps to draw this on a piece of paper and hold it above your head while looking up at it...

![radec2xy](https://user-images.githubusercontent.com/218471/41175926-8ac5dd82-6b13-11e8-8ec2-f5d17d8a6f08.png)

forero:
Fixed by #127

PR #127, opened 2018-06-25
fixes bug in radec<->xy conversion

forero:
Solves #118


forero:
Here is the output for a basic test. Things look reasonable.

tile

Issue #96, opened 2017-11-29
propagating input target information


forero:
There is additional target information that must be propagated through `fiberassign`. Where is the best place to make this propagation? Inside `fiberassign`? In `mtl`? 
@sbailey can you give us some examples of the kind of information/columns that we should be propagating? Is it every column that can be found in the `targets`/`mtl` file?

forero:
@sbailey  Do we know what information should be propagated? 

Related to this we have this comment from Kyle in the last internal review 

> We need to finalize the [fiberassign] outputs so that Klaus has something to test in mock observing.  We need to finalize the interface by which the fiber assignment outputs are transferred to the ICS system as part of the next field selector routine.

What is the minimal set of columns that Klaus needs?

djschlegel:
If there are to be any changes, they should be reflected in the data model for the tile files:
  http://desidatamodel.readthedocs.io/en/latest/DESI_TARGET/fiberassign/tile.html
It looks like the current outputs do correspond to the current data model.

sbailey:
Some details are still being worked out, but here is at least a starting set of changes to make to fiberassign output (and thus to MTL in some cases):

Columns to add:
  * [ ] SUBPRIORITY (float64), from targets -> MTL -> fiberassign
  * [ ] PETAL_LOC (int8, 0-9 = LOCATION // 1000)
  * [ ] DEVICE_LOC (int16, 0-526 = LOCATION % 1000)

Placeholder columns to add; these will come from targets -> MTL -> fiberassign but aren't in the target catalogs yet:

  * [ ] PMRA, PMDEC (float32) default 0.0 (proper motions from Gaia)
  * [ ] PMRA_IVAR, PMDEC_IVAR (float32) default 1.0 (inverse variances of PM*)
  * [ ] PARALLAX (float32) default 0.0
  * [ ] PARALLAX_IVAR (float32) default 1.0
  * [ ] Header keyword (not column) REFEPOCH = 2015.5 // Gaia Proper Motion Reference Epoch

Updates to existing columns:
  * [ ] Rename RA,DEC -> TARGET_RA, TARGET_DEC (for consistency with ICS and clarity of target vs. fiber vs. telescope...)
  * [ ] Replace BRICKNAME (S8) with BRICKID (int32)
  * [ ] Change X,Y coordinate system to Q(deg), S(mm)
    * The desimodel fiberpos.fits file has Q,S columns though the fiberpos.txt file does not; See DESI-0530 for coordinate system pictures
[...]

PR #141, opened 2018-09-02
propagates input information

forero:
This PR implements the solution #96. It doesn't  propagate/compute all the required information in the original issue. Some questions about the best way to implement it remain open. I continue the discussion in #96.


sbailey:
This PR looks fine, though I see commits very recently so I'm going to wait on merging until you confirm that you are done.  Please point to some example output at NERSC with the commands used to generate it using this branch.  Thanks.

forero:
@sbailey Right. I am still making commits. I will ping you once this PR is ready for review.

forero:
@sbailey Please go ahead and review this PR. I focused on the HDU1 columns. 
Here  I have inputs with the latest format from Adam: `/global/cscratch1/sd/forero/small_dr7.1`

This command should work with the inputs there:
`fiberassign --mtl cut_mtl_dr7.1-PR372.fits --sky cut_skies_dr7.1-0.22.0.fits --stdstar cut_std_dr7.1-PR372.fits --fibstatusfile fiberstatus.ecsv --footprint ./cut_desi_tiles.fits --outdir output_fiberassign/ --gfafile cut_gfas_dr7.1.fits`
 
The implementation in this PR is definitely not backwards compatible.  The script expects the latest `desitarget` format.

geordie666:
I wouldn't be *too* concerned about backwards compatibility. If I run on an older version of the sweeps, it will populate the columns in the new data model. So, if we need DR6/DR5 etc., I can rerun them.

sbailey:
Thanks.  I added some more columns and sped it up by pre-filtering what targets could be on what tiles so that the MTL + fiberassign join now takes <0.1 sec/tile instead of ~5 sec/tile (which would have taken ~14 hours to write all 10k tiles).  Merging now.

forero:
Thanks!

PR #146, opened 2018-09-25
Fix collision calculation

### LINKED ISSUE #145: Collision calculation has a bug
As part of my work re-factoring / re-writing fiberassign, I discovered a bug in the collision calculation.  The central body and ferrule holders are modeled as circles and segments, and overlap of these shapes are used to detect collisions.  For mult-point segments, each subsegment must be checked for a collision.  However, the current code returns immediately after checking only the first piece of a multi-point segment:

https://github.com/desihub/fiberassign/blob/master/src/collision.cpp#L186

For verification of the problem, using the small test data set, tile 1148, fiber 1597 is assigned to a target which produces the configuration in the attached image.  My new code identified this as a collision, but the master version did not.  I'll open a PR in a moment to fix this in master, even though the code will be replaced shortly.  The consequences are that any previous collision calculations involving the multi-point central body segment and the ferrule holder head, where the head crosses segments other than the first one, would be identified as not colliding.
collide

tskisner:
Small fix for intersections of circles with a multi-point segment.  Closes #145.

### DIFF
--- a/src/collision.cpp
+++ b/src/collision.cpp
@@ -183,13 +183,23 @@ bool intersect (const element & e1, const element & e2) {
     }
     if (e1.is_seg && !e2.is_seg) {
         for (size_t i = 0; i < e1.segs.size() - 1; i++) {
-            return intersect_seg_circ(e1.segs[i], e1.segs[i + 1], e2.O, e2.rad);
+            bool b = intersect_seg_circ(e1.segs[i], e1.segs[i + 1], e2.O,
+                                        e2.rad);
+            if (b) {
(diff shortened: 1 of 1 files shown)


rncahn:
Excellent work, Ted.  As I have said previously, it is important to check
the entire collision code against something written independently.

Bob

forero:
Thanks @tskisner . Sorry for replying too late. I was also checking against other runs to gauge the effect. This changes 1-5% of the target-fiber pairs in the final assignments on DR7.

PR #148, opened 2018-10-11
added plot to qa-fiberassign

akrolewski:
Added plot capability to qa-fiberassign using code used to produce Fig. 3 in https://arxiv.org/pdf/1809.07355.pdf.
Also restructured arguments so that qa-fiberassign takes a file with a list of tilefiles as input rather than a list of tilefiles directly (in case you want to run many tiles at once).

### DIFF
--- a/bin/qa-fiberassign
+++ b/bin/qa-fiberassign
@@ -15,17 +15,46 @@ from astropy.io import fits
 
 from desitarget.targetmask import desi_mask
 import desimodel.io
+from desimodel.focalplane.geometry import xy2radec
+
+import matplotlib
+matplotlib.use('Agg') 
+
+import matplotlib.pyplot as plt
+import time
+#plt.ion()
+#plt.show()
 
 parser = argparse.ArgumentParser()
 parser.add_argument("--targets", type=str,  help="input data")
 parser.add_argument("--verbose", action="store_true",  help="print non-error info per tile")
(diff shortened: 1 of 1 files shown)


sbailey:
Thanks for adding the code to make that plot.

Please update the arguments so that if you give it `--tilefile` it will use that, but otherwise it still lets you do `qa-fiberassign tile*.fits` too.

Please also update it so that if you don't provide `--plotdir`, then it won't spend the time making and saving N>>1 plots.

Does this produce meaningful output on bright time tiles that may not have any ELG, LRG, or QSO targets?

michaelJwilson:
Interest in finishing this off?  

image

image

Can be produced by 
https://github.com/michaelJwilson/fiberassign/blob/master/py/fiberassign/smith.py

weaverba137:
@sbailey, is this branch/PR worth keeping?  I think it would be impossible to merge in its current state.

weaverba137:
@araichoor, same question to you: is this branch/PR worth keeping?

araichoor:
I don t know; that happened before I arrived.
personally, I never used that script, so I don t really know if it s worth it to put effort to bring it in consistency with today s code.
on my side, I won t have time to do that.

maybe @akrolewski has a point of view? (though that s four years old...)

araichoor:
I ll close this PR.
I guess one can still revive the branch+PR if need be.

PR #150, opened 2018-10-26
Add option to print version to stdout

apcooper:
I wanted to get at the version string stored in the executable from the command line, so I added a `--version` option to the C code and the python wrapper.  To make this work I changed the wrapper to handle required arguments manually rather than via `required=True` in `argparse`.

```
apcooper@cori19:~/desi/code/fiberassign> ./bin/fiberassign --version
0.9.0.dev2241
```

### DIFF
--- a/bin/fiberassign
+++ b/bin/fiberassign
@@ -19,8 +19,8 @@ import shutil
 import subprocess
 
 parser = argparse.ArgumentParser()
-parser.add_argument("--mtl", type=str,  help="input targets (FITS file)", required=True)
-parser.add_argument("--sky", type=str,  help="input sky positions (FITS file)", required=True)
+parser.add_argument("--mtl", type=str,  help="input targets (FITS file)")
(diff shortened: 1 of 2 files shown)


bin/fiberassign
```diff
@@ -42,9 +42,17 @@ parser.add_argument("--nstarpetal", type=int, help="number of standard stars per
 parser.add_argument("--nskypetal", type=int, help="number of sky fibers per petal (default=40)", default=40)
 
 parser.add_argument("--nocleanup", dest='cleanup', default=True, action='store_false')
+parser.add_argument("--version",help='Print C code version and exit',action='store_true')
```
weaverba137: You should use `action='version'`, then you should not need to remove the required arguments.  See https://docs.python.org/3/library/argparse.html#action.  If necessary, get the version from the C++ code beforehand to pass set in `parser.add_argument()`.

apcooper: Nice, thanks. I've made that change and restored the `required=True` to the other options.

PR #156, opened 2018-12-13
fix POTENTIAL table

### LINKED ISSUE #155: POTENTIAL list is wrong
Context: in the fiberassign output, the FIBERASSIGN HDU has the list of what TARGETIDs are assigned to what FIBERs, and the POTENTIAL HDU has the list of all TARGETIDs that are covered by each FIBER.

Bug: In current master, the assigned TARGETID is almost never in the list of POTENTIAL targets, but it does appears for other FIBERs that are nowhere near the assigned fiber.  I suspect that this is due to a sorting bug (e.g. perhaps I introduced it in #147 when I sorted the output by FIBERID as needed downstream).

If it is simple to fix, it would be helpful to fix in master even if it is correct in the tsk_refactor branch.  We should also have unit tests or qa-fiberassign checks that would have caught this.

Adding this to 18.12 last minute in the hopes that we can get correct output for the next reference run.

sbailey:
This PR fixes #155 where the assigned TARGETID for each fiber didn't appear in the potential assignments list for each fiber.  It was a side affect of PR #147 that ensured that the FIBERASSIGN HDU output was sorted by TARGETID.

I also added a check in `qa-fiberassign` that would have caught this.

@tskisner a new fiberassign run of the 18.11 target list is in /global/cscratch1/sd/sjbailey/desi/fiberassign/18.11-redo .  Please check whether those potential assignments more closely match what you have in the tsk_refactor branch.

### DIFF
--- a/bin/fiberassign
+++ b/bin/fiberassign
@@ -367,14 +367,22 @@ for sky_id in sky_tile_id.keys():
         sky_data = fitsio.read(sky_tile_id[sky_id])
         fiber_data = fitsio.read(target_tile_id[sky_id], ext=1)
         potential_data = fitsio.read(target_tile_id[sky_id], ext=2)
-        
+
         target_data = make_target_hdu_data(fiber_data, mtl_data[mtl_tile_indices[int(sky_id)]])
         tile_data  = footprint[footprint['TILEID']==np.int(sky_id)]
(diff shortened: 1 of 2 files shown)


tskisner:
Hello @sbailey , yes the available targets HDU now looks good in master branch.

PR #166, opened 2019-01-24
pre-flight checks and --overwrite option

sbailey:
This PR checks for pre-existing output files to avoid the situation of running for 30 minutes and then crashing because some output file already exists.  Either the files must not exist, or the user must provide an `--overwrite` option.  Otherwise the code will print an error and exit before doing a bunch of calculations.

It also updates the I/O to write temporary files and only after they are successfully written rename them to the final output name.  This prevents the code from leaving corrupted files behind if interrupted in the middle of I/O (e.g. from a job timing out).

Note: this PR is into the tsk_refactor branch, not into master.

### DIFF
--- a/bin/fba_run
+++ b/bin/fba_run
@@ -4,13 +4,15 @@ Main fiber assignment entry point
 """
 import os
 
+import sys
+
 import argparse
 
 from datetime import datetime
 
 import desimodel.io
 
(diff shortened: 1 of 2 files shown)


tskisner:
Those changes all look good, thanks.

PR #167, opened 2019-01-25
allow user to override sciencemask, stdmask, etc.

sbailey:
This PR allows the user to override sciencemask, stdmask, safemask, and skymask via `fba_run` command line arguments.  These can either be int or string, e.g.
```
fba_run ... --sciencemask 3
fba_run ... --sciencemask LRG,ELG
```

It also introduces an `excludemask`, which will exclude any targets that have bits matching that mask set.  The default is `BRIGHT_OBJECT` and `IN_BRIGHT_OBJECT`.  Targets with those bits set may be in the target files but should never be assigned a fiber.

Question: This PR changes the default safemask to `BAD_SKY` instead of 0.  I know that the concept of safe targets was deprecated (at my request) by just treating BAD_SKY targets as ultra low priority normal targets, but the code was left in place in case I changed my mind (which I probably have but I don't recall why...).  Will setting safemask != 0 mess that up?

Along the way this update also allows `fiberassign.targets.desi_target_type(desi_target)` to accept scalar or vector inputs instead of just scalars.

Note: this PR is in to the tsk_refactor branch, not the master branch.  It is filling in some features needed to provide a backwards compatible `fiberassign` script that wraps `fba_run` and `fba_merge` using the same syntax as current master `fiberassign`.

### DIFF
--- a/bin/fba_run
+++ b/bin/fba_run
@@ -10,8 +10,6 @@ import argparse
 
 from datetime import datetime
 
-import desimodel.io
-
 from desitarget.targetmask import desi_mask
 
 from fiberassign.utils import GlobalTimers, Logger
@@ -32,6 +30,7 @@ from fiberassign.targets import (str_to_target_type, TARGET_TYPE_SCIENCE,
                                  )
 
 from fiberassign.assign import (Assignment, write_assignment_fits, result_path)
+
 
 def main():
     log = Logger.get()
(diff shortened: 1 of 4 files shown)


tskisner:
I'm going to push a style change- the code uses double quotes everywhere, and it is good to have consistency:

https://www.python.org/dev/peps/pep-0008/#id25

And especially for the triple quotes...  Other than that it looks good, thank you!

PR #168, opened 2019-01-25
update --targets option; add bin/fiberassign

sbailey:
This PR does 2 things:

1. Updates `--targets` option to accept multiple arguments instead of specifying it multiple times.  This is motivated by an upcoming change to the input target files where they will be split by healpix (and likely we'll want to do the same for MTL).  This change enables `fba_run --targets mtl*.fits ...` instead of `fba_run --targets mtl1.fits --targets mtl2.fits ... --targets mtl35.fits ...`.

2. Adds a new script `bin/fiberassign` that is option-compatible with `bin/fiberassign` in current master for the standard case of running on multiple tiles.  This just wraps `fba_run` and `fba_merge` but does so in a way that people who have previously been using the master branch can continue to use the same commands, without having to switch over to the 2-step `fba_run` + `fba_merge` with slightly different option names.

`bin/fiberassign` does *not* implement the arbitrary pointing options to original fiberassign (specifying an RA, dec instead of a list of tiles).  I suggest that we add that separately as a different script.

### DIFF
--- a/bin/fba_merge_results
+++ b/bin/fba_merge_results
@@ -4,6 +4,8 @@ Merge fiberassign results with input target files.
 """
 import os
 
+import sys
+
 import argparse
 
 from fiberassign.assign import merge_results, result_tiles
@@ -12,7 +14,7 @@ from fiberassign.assign import merge_results, result_tiles
 def main():
     parser = argparse.ArgumentParser()
 
-    parser.add_argument("--targets", type=str, required=True, action="append",
+    parser.add_argument("--targets", type=str, required=True, nargs="+",
                         help="Input file with targets of any type.  This "
                         "argument can be specified multiple times (for "
                         "example if standards / skies / science targets are "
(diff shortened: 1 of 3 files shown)


sbailey:
At minimum it's a transitional step to not break the rest of our end-to-end integration, and I wanted to do that in a way that was minimally disruptive to existing fba_run and fba_merge_results scripts (i.e. calling them via subprocess rather than gutting them into libraries).

Let's do that script -> library gutting in a separate update, focusing on what we would want to be available to a library to enable wrapping it with MPI to do 1000 mocks in parallel, rather than just avoiding scripts that call scripts with subprocess.

When `fba_run` has been augmented to output the full merged results in a single step, we could then consider making the non-backwards compatible change to renaming that to `fiberassign`.  But for now there is value in both having backwards compatibility with `fiberassign` while also having the lighter-weight faster minimalist `fba_run`.

tskisner:
Ok, that sounds fine.  Merge away :-)

Issue #165, opened 2019-01-14
Refactor: understand lower GRAY ELG assignment efficiency


sbailey:
Comparing tsk_refactor branch with master using a set of 299 tiles covering DR7.1 range 130 < RA < 150 and 0 < dec < 20, I find that tsk_refactor assigns ~5% fewer ELG targets  to GRAY tiles, making up the difference with stdstars and SKY fibers.  The loss of ELGs is not made up by more ELGs assigned during DARK time.  There is overall a loss of ~1% fewer science targets overall.

Track down specific cases starting with ELGs on GRAY tiles to understand this loss of assignment efficiency.

Here is what I ran for master:
```
basedir=/global/cscratch1/sd/sjbailey/desi/fiberassign/refactor
fiberassign \
    --mtl $basedir/mtl.fits 
    --stdstar $basedir/standards.fits \
    --sky $basedir/skies-dr7.1-0.27.0-lite.fits \
    --footprint $basedir/tiles-lite.fits \
    --outdir $SCRATCH/desi/fiberassign/refactor/master
```

and for tsk_refactor branch:
```
basedir=/global/cscratch1/sd/sjbailey/desi/fiberassign/refactor
fba_run --overwrite \
    --targets $basedir/mtl.fits $basedir/skies-dr7.1-0.27.0-lite.fits \
    --footprint $basedir/tiles-lite.fits \
    --dir $SCRATCH/desi/fiberassign/refactor/output

fba_merge_results \
    --targets $basedir/mtl.fits $basedir/skies-dr7.1-0.27.0-lite.fits \
    --dir $basedir/output \
    --out $SCRATCH/desi/fiberassign/refactor/output
```

I cached a copy of the outputs in `/global/cscratch1/sd/sjbailey/desi/fiberassign/refactor/compare/`.

forero:
@sbailey how is it that the `fba_run` guesses the standard stars file?

forero:
I ran these two commands:
`fiberassign --mtl mtl.fits --sky skies-dr7.1-0.27.0-lite.fits --stdstar standards.fits --outdir output_new/` (refactor version)

`oldfiberassign --mtl mtl.fits --sky skies-dr7.1-0.27.0-lite.fits --stdstar standards.fits --outdir output_old/` (legacy version)

and compared the lenght of the potential tables. These two should match because those are the available targets, there isn't any assignment involved there. However the two differ in almost every tile, some times by a large amount. Here is a histogram using 100 tiles:

[...]

PR #174, opened 2019-02-11
Apply per-petal standards limits when assigning science targets.

tskisner:
This PR is against the tsk_refactor branch and should NOT be merged as-is.  This is a demonstration of changes needed to account for per-petal standard limits when assigning science targets.

Essentially a new argument is added to the assign_unused and redistribute_science methods.  This argument is the maximum standards per petal limit to use when assigning things that are both standards and science targets.  This is a follow up the discussion in #165, where I mentioned that greater than 10 standards were being assigned per petal in some cases during the assignment of science targets (i.e. prior to any intentional assignment of standards).

It is not clear to me that this new behavior is desired, but here is a comparison.  First the existing code in tsk_refactor running on the dr7.1 "lite" dataset, looking at petal one of the last tile (45660).  First plot is the initial assignment of science targets.  Second plot is after the redistribution step.

[fiberassign_45660_science.pdf](https://github.com/desihub/fiberassign/files/2849893/fiberassign_45660_science.pdf)
[fiberassign_45660_science-redist.pdf](https://github.com/desihub/fiberassign/files/2849894/fiberassign_45660_science-redist.pdf)

Next we have the same data from this branch:

[fiberassign_45660_science.pdf](https://github.com/desihub/fiberassign/files/2849896/fiberassign_45660_science.pdf)
[fiberassign_45660_science-redist.pdf](https://github.com/desihub/fiberassign/files/2849897/fiberassign_45660_science-redist.pdf)

Recall that this is before any intentional assignment of standards.  The green targets are "dual" science and standards.  Note that in the code from this branch, only a maximum of 10 per petal of these dual targets are assigned during the assignment of science targets.

If these changes are desired, I will un-comment the remaining steps in fba_run, and also change the function signature of the Assignment class methods to take kwargs (since passing in "-1" is clunky).  Feedback welcome.

### DIFF
--- a/bin/fba_run
+++ b/bin/fba_run
@@ -242,33 +242,34 @@ def main():
     asgn = Assignment(tgs, tgsavail, favail)
 
     # First-pass assignment of science targets
-    asgn.assign_unused(TARGET_TYPE_SCIENCE)
+    asgn.assign_unused(TARGET_TYPE_SCIENCE, -1, "POS", -1, -1,
+                       args.standards_per_petal)
 
     # Redistribute science targets across available petals
-    asgn.redistribute_science()
+    asgn.redistribute_science(-1, -1, args.standards_per_petal)
 
-    # Assign standards, 10 per petal
-    asgn.assign_unused(TARGET_TYPE_STANDARD, args.standards_per_petal)
-    asgn.assign_force(TARGET_TYPE_STANDARD, args.standards_per_petal)
-
-    # Assign sky to unused fibers, up to 40 per petal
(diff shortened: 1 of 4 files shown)


tskisner:
After further offline discussion with @sbailey, I think there is consensus that making science assignment "aware" of per-petal standards limits is not desired.  Only bright tiles have enough of these "dual" targets to possibly get more than 10 per tile through the normal science assignment process (this was checked with a quick interactive test).  And in the case of bright tiles it is OK if the per-petal threshold is exceeded when placing these as science targets.  Closing this PR, but will leave the branch open for reference.

Issue #175, opened 2019-02-14
Treat BAD_SKY as a unique target type


tskisner:
This ticket attempts to capture discussion on recent telecons an offline conversations.  Currently fiberassign (both the legacy and new code) treats BAD_SKY targets as low-priority science targets.  Instead, I propose to treat these targets as separate from science, standards, and sky.  There is actually already a fourth target type built in to fiberassign called "safe" targets.  This target type was intended to simply be a safe place to put an unassigned fiber as a last resort.

Specifically, the proposed behavior is:

1.  When loading targets from a file, any target with BAD_SKY set in DESI_TARGET would by default be assigned the internal fiberassign type TARGET_TYPE_SAFE instead of TARGET_TYPE_SCIENCE.

2.  In the top-level main script, this target type would not be used until the very end, and then a single call to ```assign_unused(TARGET_TYPE_SAFE)``` would be made.  This line is actually already in the code and just needs to be uncommented.

The main consequence of this change is that there would likely be more "regular" sky targets assigned and fewer bad_sky targets.

After this is done, we should also close #152 .

sbailey:
+1

Fleshing out this comment a little more since I was the one who originally requested that BAD_SKY be treated just as ultra low priority science targets, but then in discussions with @tskisner and @kdawson1000 I keep re-discovering that this is a bad idea and then forgetting why:

> The main consequence of this change is that there would likely be more "regular" sky targets assigned and fewer bad_sky targets.

Currently BAD_SKY are just ultra-low priority science targets, which means that they get used as filler targets *before* using SKY as filler targets.  We'd rather use SCIENCE, STD, and SKY first, and then BAD_SKY as the last type of filler target.

i.e. our order of assignment should be:

1. science targets with numobs_more>0
2. stdstars on unused fibers up to some density limit (default 10/petal)
3. sky on unused fibers up to some density limit (default 40/petal)
4. if needed, bump low priority science to achieve minimum stdstar density
5. if needed, bump low priority science to achieve minimum sky density
6. assign science to any unused fibers (I think this is a no-op due since they should have been placed in step (1) anyway
7. assign std to any unused fibers (possibly exceeding original minimum stdstar density)
8. assign sky to any unused fibers (possibly exceeding original minimum sky density)
9. assign BAD_SKY to any unused fibers

At this point there should not be any unassigned fibers left, or otherwise we need to go back to target selection and get more SKY/BAD_SKY targets at the unassigned locations.

[...]

Issue #181, opened 2019-02-16
Fiberassign to recognize SV mask


Srheft:
@tskisner @sbailey:

It appears that the current Fiberassign code only recongnizes the main survey masks:
[https://github.com/desihub/fiberassign/blob/3beff2b8d484233fd705b33021e7f6972f11dcc2/py/fiberassign/assign.py#L29](url) 

If there is already a way to exchange the masks via "--options", please let me know and if the comment is valid (test materials are [here](http://www.astro.utah.edu/~u6022465/test/)), please accomodate the means to run fiberassign for SV targets (includes having `from desitarget.sv1.sv1_targetmask import desi_mask` ).

In my experience, [running fiberassign](http://www.astro.utah.edu/~u6022465/test/FA.log) with this MTL file ([MTL_DEEP2_1.fits](http://www.astro.utah.edu/~u6022465/test/)) from the SV catalog returns an ALL SKY assignment while the field is quite crowded with science targets of equaly high priority:

![image](https://user-images.githubusercontent.com/32471902/52898206-3b3aa580-3198-11e9-9639-1e81bd944ef8.png)

If you have a working example that proves otherwise, please share it here as I did not find any issue with the input files I used. Thanks.

sbailey:
fiberassign does not currently support this (but it should).  I think there are two pieces to this:
* what column name should be used for the targeting mask bits (default DESI_TARGET)?
* what should the default masks be for SCIENCE, STDSTAR, SKY, SAFE, and exclude locations, since the exactly bits will be different for different programs?

In the code we could add a --program option (default DESI, but also SV1, CMX, ...) and pass that as arguments to `fiberassign.targets.get_sciencemask()`, `.get_stdmask()` etc.

Opinions welcome if this really is "program" or if a different word should be used.

@Srheft could you please fill out / check these details:

Defaults for program DESI:
  * column name: `DESI_TARGET`
  * mask name: `desitarget.targetmask.desi_mask`
  * SCIENCE mask: `ELG|LRG|QSO|BGS_ANY|MWS_ANY|SECONDARY_ANY`
  * STD mask: `STD_FAINT|STD_BRIGHT|STD_WD`
  * SKY mask: `SKY`
  * SAFE mask: `BAD_SKY`
  * EXCLUDE mask: `BRIGHT_OBJECT|IN_BRIGHT_OBJECT`

Defaults for program SV1:
  * column name: `SV1_TARGET`
[...]

PR #188, opened 2019-03-05
Move high-level code into the package

tskisner:
This work moves the internals of the entry point scripts into the package.  The functions for parsing arguments and running the code are now in the fiberassign.run sub-module and are added to the API reference in the documentation.  Other changes:

- The new option to fba_run (`--mask_column`, default "DESI_TARGET") can be used to override the column in the FITS files containing the target bits.

- A new option to fba_run (`--by_tile`) does the outer loop over tiles and fully assigns each tile in sequence before moving on to the next.  @sbailey , can you verify that fiberassign.run.run_survey_bytile() essentially does the looping in the way your modified version of fba_run was working?

This work should address #181, since now it is easier to run the assignment interactively for commissioning and also to override the FITS column used for mask operations.


py/fiberassign/run.py
```diff
@@ -0,0 +1,997 @@
...
+                           "POS", tile_id, tile_id)
+        asgn.assign_force(TARGET_TYPE_STANDARD, args.standards_per_petal,
+                          tile_id, tile_id)
+
+        # Assign sky to unused fibers, up to 40 per petal
+        asgn.assign_unused(TARGET_TYPE_SKY, args.sky_per_petal, "POS",
+                           tile_id, tile_id)
+        asgn.assign_force(TARGET_TYPE_SKY, args.sky_per_petal,
+                          tile_id, tile_id)
+
```
sbailey: After doing these, we should have a final `assign_unused` for both standards and sky to fill in any remaining unused fibers since the previous calls max out at `args.standards_per_petal` and `args.sky_per_petal`.  i.e.
```python
#- I think this first one is a no-op, but would be good to confirm
### asgn.assign_unused(TARGET_TYPE_SCIENCE, -1, "POS", tile_id, tile_id)
asgn.assign_unused(TARGET_TYPE_STANDARD, -1, "POS", tile_id, tile_id)
asgn.assign_unused(TARGET_TYPE_SKY, -1, "POS", tile_id, tile_id)
```

With this addition, this is equivalent to what I tested before for my per-tile serial assignments tests.

py/fiberassign/run.py
```diff
@@ -0,0 +1,997 @@
...
+    Returns:
+        None
+
+    """
+    qadata = None
+    with open(args.qafile, "r") as f:
+        qadata = json.load(f)
+    plot_qa(qadata, args.outroot, labels=args.labels)
+    return
```
sbailey: We're at almost a 1000 lines for this file and it is mixing the command line parsing and run functions for multiple different scripts separated by hundreds of lines of code for the pieces that go together.  I'd prefer to organize these like desispec et al, with desispec/scripts/ and one module file per script file.

tskisner:
Ok, will make those changes and the single line of code to go from 5 -> 6 digits for the tile ID in the output files.

Srheft:
@tskisner : having trouble with correct use of --mask_column

- Using the branch **driver**,  my example tile file, MTL and SKY files are posted [here](http://www.astro.utah.edu/~u6022465/cmx/one_brightSTAR_in10PETALS/rmag_9th/).

- The MTL file has a column called **SV1_DESI_TARGET**  with target bits for SV that I aim to replace with the default **DESI_TARGET**  via the new option --mask_column.

- My runing command was: `fba_run --targets ./MTL_DEEP2_1.fits ./33-40--3-3-skies-dr7.1-0.27.0.fits --footprint ./tiles_18pRosette_DEEP2-1.fits --mask_column SV1_DESI_TARGET --by_tile --sky_per_petal 80 --standards_per_petal 20 --dir ./FAout_deep1 --write_all_targets --overwrite ` 

- targets.py complains about the field name: 
`fiberassign/targets.py", line 187, in append_target_table
    tgdata[typecol], sciencemask=sciencemask, stdmask=stdmask,
ValueError: no field of name SV1_DESI_TARGET
`

sbailey:
Structurally looks good; thanks.  Please resolve Sarah's test problems and the unittest failures (they appear to be real, not just failures of Travis itself) before merging.

tskisner:
The known problems are fixed, investigating the test dataset above now.

tskisner:
The input data files are missing in the link above, so I cannot test.  I can coordinate testing with @Srheft offline.  I would like to merge this branch since I am ready to open a PR that fixes #175, and which is based on code in this PR.  I'll open a separate issue if needed for fixing any problems with the `--mask_column` option.

PR #189, opened 2019-03-07
Restore "safe" target type

### LINKED ISSUE #175: Treat BAD_SKY as a unique target type

### LINKED ISSUE #187: Fibers remain unassigned in BAD_SKY regions
@sbailey, @tskisner:

Points shown in black are BAD_SKY (68719476736) locations. Was the motivation for including them in the sky catalog to have them as the last resort for fibers to get assigned to " made up yet sensible coordinates"?  If yes, then they are not getting picked up by any fibers. 
 

image

tskisner:
This restores the previous handling of BAD_SKY as a separate target type (a.k.a. safe).  This target type is assigned at the very end.  Running the dr7.1-lite dataset actually only had one safe target assigned (plot below).  The rest of the unassigned fibers were simply off the edge of the footprint.  This closes #175 and also should close #187.
dr7 1-lite_045623

### DIFF
--- a/py/fiberassign/assign.py
+++ b/py/fiberassign/assign.py
@@ -32,14 +32,9 @@ from ._version import __version__
 
 from .utils import Logger, Timer, default_mp_proc
 
-from .tiles import load_tiles
+from .targets import (TARGET_TYPE_SKY, desi_target_type)
 
-from .targets import (TARGET_TYPE_SCIENCE, TARGET_TYPE_SKY,
-                      TARGET_TYPE_STANDARD, TARGET_TYPE_SAFE,
-                      TargetTree, TargetsAvailable, FibersAvailable,
-                      load_target_file, desi_target_type)
-
-from .hardware import (load_hardware, FIBER_STATE_STUCK, FIBER_STATE_BROKEN)
+from .hardware import (FIBER_STATE_STUCK, FIBER_STATE_BROKEN)
 
 from ._internal import Assignment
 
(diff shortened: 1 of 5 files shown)


sbailey:
This looks good.  @Srheft please test on your dataset to verify that BAD_SKY locations are being assigned with this branch.

Srheft:
Yes, this branch assigns BAD_SKY. Two of the fibers are left unassigned [the full sky catalog was passed so it can not be the case of cutting the sky file too tight foredge fibers and even that issue has been fixed]. 
image

I think Ted meant to say that I have to use this branch not the master when he showd the assigned BAD_SKY [in #187](https://github.com/desihub/fiberassign/issues/187).

Issue #190, opened 2019-03-07
sky monitor assignment when no SKY are available


sbailey:
From #187:

The sky monitor fibers normally would be assigned to SKY targets.  There will be cases where there are no good SKY targets, in which case they should fall back to using a BAD_SKY location (but not a science target since ).  If we ever have cases without either SKY or BAD_SKY, track those down.

FIBERSTATUS bookkeeping is already included in FASSIGN for sky monitor positioners too, e.g. if one of them breaks, but we could also add a bit for BAD_SKY for convenience.

Srheft:
@sbailey:
Fibers 5010 and 5011 in tile 300005 in [this](http://www.astro.utah.edu/~u6022465/cmx/all_STDtest_SV0bright/) assignment are left unassigned although there seems to be BAD_SKY positions available for them. 

![image](https://user-images.githubusercontent.com/32471902/53999750-e4bee800-4101-11e9-9443-aef8f14d81f3.png)


@tskisner @kdawson1000:
On the last comment in #187:

> We are going to need a solution to the way that the sky monitor fibers are documented

qa.json files already provide a useful easy-access assignment summary but there may be a way to make it useful for ETC folks ([example](http://www.astro.utah.edu/~u6022465/cmx/all_STDtest_SV0bright/qa.json) from the current master for tile 300005). 

In Tile 300005 all the POS fibers are assigned so the key "unassigned" showes empty and we seem to be "good to go" if we don't look at the plots [in cases of 1000s of tiles, plotting or looking at all of them may not even happen]. 

option 1: adding the **unassigned ETC** fibers to list of "unassigned" fibers.
option 2: dividing "unassigned" key into "unassigned POS" and "unassigned ETC" 

Is this a sensible proposal? and if yes, does any other option come to mind?

tskisner:
The current code only attempts to assign sky targets to the ETC fibers:

https://github.com/desihub/fiberassign/blob/master/py/fiberassign/scripts/assign.py#L326

That is why the ETC fibers are not assigned (there are no sky targets there, only bad sky targets which are of TARGET_TYPE_SAFE).

This was my original question- do we want to allow the ETC fibers to be assigned to "bad sky"?  If so, it is a one line change.

I have no preference for how we want to track ETC fibers in the QA.

Srheft:
Correct. Just wanted to document an example for cases where there is simply no sky available and so we juat have to go for BAD_SKY.

dkirkby:
[...]

PR #191, opened 2019-03-11
If no sky targets are available for ETC fibers, assign safe targets.

tskisner:
For both normal positioners and ETC fibers, if assigned to a safe target, set the fourth bit (value=8) of FIBERSTATUS.  This change addresses #190, and downstream codes can check FIBERSTATUS to see which fibers are assigned to BAD_SKY or other "safe" (but undesirable) targets.


sbailey:
OK, for this PR, but I'll open a new ticket about how our state/bits are getting messy with hardcoded 1,2,4,8...

Issue #182, opened 2019-02-19
Improvements to event logging / reconstruction


tskisner:
There has been much discussion about tracking fiberassign "events" (why a particular fiber or target was not assigned).  For "debug" use cases exploring issues like target density, file formats, etc, there are already the DESI_DEBUG_{TARGET,FIBER,TILE} environment variables that allow digging deeper into these questions.

For production use, reporting every detail would be hugely expensive.  However, there is something we could do better.  We already store a dictionary of available target IDs for every tile/fiber.  As a small epsilon on the memory use, we could store one byte of flag bits for each of these available targets and indicate why that target was not assigned to a particular fiber.

This would require internal code changes but would not impact the python interfaces.  We would write this extra flag field in the available targets HDU.  The bits I can think of at the moment would be:

- 0 (no bits set, target is assigned to this fiber)
- 1 (target wrong type during assignment)
- 2 (no obs remaining)
- 3 (lower priority than chosen target)
- 4 (collision)
- 5 (per-petal threshold exceeded)
- 6 (bumped)

These bits will be flipped multiple times- for example when assigning science targets to unassigned fibers, lower priority targets would get bit 3 raised for that fiber.  An available sky target would get bit 1 raised.  Later if that fiber was re-assigned to a sky target (for example), the sky target would lower bit 1 and the bumped science target would raise bit 6.

Note:  There **will** be a performance penalty for all this based on memory access alone, but I won't know how bad that is until giving it a try.

tskisner:
In the process of creating a recent tutorial (https://portal.nersc.gov/project/desi/users/kisner/fiberassign/FiberAssignAlgorithms_Part2.html) I had to jump through some MTL hoops in order to get the number of targets that were not only "available" (physically reachable), to a positioner, but which also had `NUMOBS_MORE > 0` and hence could be "considered" for assignment.

Having gone down that road, I am more convinced that:

1.  We actually really do need this information tracked in the output.

2.  This information should "live" inside the `TargetsAvailable` class.  For every tile and positioner there is currently a list of available (physically reachable) target IDs.  There should also be a bitfield for each of these available targets indicating why that target did not get assigned to the specific positioner.

Not only would this be a relative easy code change, it would also allow us to easily write that extra byte as a column of the `FAVAIL` /  `POTENTIAL_TARGETS` HDU.  This would allow us to answer all of these questions about how many targets were really possible to assign.

PR #193, opened 2019-03-18
Support different default masks for each program.

tskisner:
Do not merge yet.  This needs to be tested with SV1 and CMX target files which have correct OBSCONDITIONS type (int64) and which match the OBSCONDITIONS of the footprint file used.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,7 +91,7 @@ env:
         # Binary packages to install with pip.
         - PIP_BINARY="pyyaml scipy matplotlib numpy astropy sphinx"
         # Packages to build with pip
-        - PIP_SOURCE="fitsio healpy"
+        - PIP_SOURCE="fitsio healpy=1.11"
         # DESI packages to install.
         - DESIUTIL_VERSION=master
         - DESIMODEL_VERSION=master
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,7 +89,7 @@ python:
 env:
     global:
         # Binary packages to install with pip.
-        - PIP_BINARY="pyyaml scipy matplotlib numpy astropy sphinx"
+        - PIP_BINARY="pyyaml scipy matplotlib numpy astropy sphinx healpy"
         # Packages to build with pip
         - PIP_SOURCE="fitsio"
         # DESI packages to install.
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,7 +91,7 @@ env:
         # Binary packages to install with pip.
         - PIP_BINARY="pyyaml scipy matplotlib numpy astropy sphinx"
         # Packages to build with pip
-        - PIP_SOURCE="fitsio healpy=1.11"
+        - PIP_SOURCE="fitsio healpy==1.11"
         # DESI packages to install.
         - DESIUTIL_VERSION=master
         - DESIMODEL_VERSION=master
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,9 +89,9 @@ python:
 env:
     global:
         # Binary packages to install with pip.
-        - PIP_BINARY="pyyaml scipy matplotlib numpy astropy sphinx healpy"
+        - PIP_BINARY="pyyaml scipy matplotlib numpy astropy sphinx"
         # Packages to build with pip
-        - PIP_SOURCE="fitsio"
+        - PIP_SOURCE="fitsio healpy"
         # DESI packages to install.
         - DESIUTIL_VERSION=master
         - DESIMODEL_VERSION=master
--- a/bin/fba_fulltest
+++ b/bin/fba_fulltest
@@ -1,0 +1,16 @@
+#!/usr/bin/env python
+"""
+Full test of fiberassign commandline tools on input files.
+"""
+
+from fiberassign.scripts.fulltest import (parse_fulltest, run_fulltest)
+
+
+def main():
+    args = parse_fulltest()
+    run_fulltest(args)
+    return
+
+
+if __name__ == "__main__":
+    main()
--- a/doc/_static/fba_fulltest.inc
+++ b/doc/_static/fba_fulltest.inc
@@ -1,0 +1,54 @@
+.. code-block:: console
+
+    usage: fba_fulltest [-h] --targets TARGETS [TARGETS ...] [--sky SKY [SKY ...]]
+                        [--gfafile GFAFILE] [--footprint FOOTPRINT]
+                        [--tiles TILES] [--out OUT] [--plotpetals PLOTPETALS]
+                        [--positioners POSITIONERS] [--status STATUS]
+                        [--rundate RUNDATE]
+                        [--standards_per_petal STANDARDS_PER_PETAL]
+                        [--sky_per_petal SKY_PER_PETAL]
+
+    optional arguments:
+      -h, --help            show this help message and exit
+      --targets TARGETS [TARGETS ...]
+                            Input file with targets of any type. This argument can
+                            be specified multiple times (for example if standards
+                            / skies / science targets are in different files). By
+                            default, the '--mask_column' (default
+                            DESI_TARGET)column and bitfield values defined in
+                            desitarget are used to determine the type of each
+                            target. Each filename may be optionally followed by
+                            comma and then one of the strings 'science',
+                            'standard', 'sky' or 'safe' to force all targets in
+                            that file to be treated as a fixed target type.
+      --sky SKY [SKY ...]   Input file with sky or 'bad sky' targets. This option
+                            exists in order to treat main-survey sky target files
+                            as valid for other survey types. If you are running a
+                            main survey assignment, you can just pass the sky file
+                            to the --targets list.
+      --gfafile GFAFILE     Optional GFA targets FITS file
+      --footprint FOOTPRINT
+                            Optional FITS file defining the footprint. If not
+                            specified, the default footprint from desimodel is
+                            used.
+      --tiles TILES         Optional text file containing a subset of the tile IDs
+                            to use in the footprint, one ID per line. Default uses
+                            all tiles in the footprint.
+      --out OUT             Top-level output directory.
+      --plotpetals PLOTPETALS
+                            Comma-separated list of petals to plot (default is all
+                            petals)
+      --positioners POSITIONERS
+                            Optional FITS file describing the fiber positioner
+                            locations. Default uses the file from desimodel.
+      --status STATUS       Optional fiber status file in astropy ECSV format.
+                            Default treats all fibers as good.
+      --rundate RUNDATE     Optional date to simulate for this run of fiber
+                            assignment, used with the fiber status file to
+                            determine which fibers currently have problems.
+                            Default uses the current date. Format is YYYY-MM-
+                            DDTHH:mm:ss in UTC time.
+      --standards_per_petal STANDARDS_PER_PETAL
+                            Required number of standards per petal
+      --sky_per_petal SKY_PER_PETAL
+                            Required number of sky targets per petal
--- a/doc/_static/fba_merge_results.inc
+++ b/doc/_static/fba_merge_results.inc
@@ -11,6 +11,12 @@
                             Input file with targets of any type. This argument can
                             be specified multiple times (for example if standards
                             / skies / science targets are in different files).
+      --sky SKY [SKY ...]   Input file with sky or 'bad sky' targets. This option
+                        exists in order to treat main-survey sky target files
+                        as valid for other survey types. If you are running a
+                        main survey assignment, you can just pass the sky file
+                        to the --targets list.
+
       --dir DIR             Directory containing fiberassign results.
       --prefix PREFIX       Prefix of each file (before the <tile>.fits).
       --split               Results are in tile prefix directories.
--- a/doc/_static/fba_run.inc
+++ b/doc/_static/fba_run.inc
@@ -22,6 +22,12 @@
                             followed by comma separated values that will be passed
                             to the arguments of the load_target_file() function.
                             See docs for details.
+      --sky SKY [SKY ...]   Input file with sky or 'bad sky' targets. This option
+                        exists in order to treat main-survey sky target files
+                        as valid for other survey types. If you are running a
+                        main survey assignment, you can just pass the sky file
+                        to the --targets list.
+
       --gfafile GFAFILE     Optional GFA targets FITS file
       --footprint FOOTPRINT
                             Optional FITS file defining the footprint. If not
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -118,6 +118,19 @@ fiber, or target specified will have all possible info logged.
     performance.  Do not use in large production runs.
 
 
+Data File Testing
+---------------------
+
+It is often useful to see if a given set of input target files are compatible
+with all of the commandline tools in fiberassign.  This usage pattern was so
+common that a high-level script was created:
+
+.. include:: _static/fba_fulltest.inc
+
+This runs the specified files through the assignment and merging, as well as
+running QA and plotting on both the raw outputs and the merged outputs.
+
+
 Legacy Compatibility Wrappers
 ---------------------------------------
 
--- a/py/fiberassign/assign.py
+++ b/py/fiberassign/assign.py
@@ -32,7 +32,8 @@ from ._version import __version__
 
 from .utils import Logger, Timer, default_mp_proc
 
-from .targets import (TARGET_TYPE_SKY, TARGET_TYPE_SAFE, desi_target_type)
+from .targets import (TARGET_TYPE_SKY, TARGET_TYPE_SAFE, desi_target_type,
+                      default_target_masks)
 
 from .hardware import (FIBER_STATE_STUCK, FIBER_STATE_BROKEN)
 
@@ -677,10 +678,15 @@ def read_assignment_fits_tile(params):
 
         targets_dtype = np.dtype(full_targets_columns)
         targets_data = np.zeros(ntarget, dtype=targets_dtype)
+        fsurvey, fcol, fsciencemask, fstdmask, fskymask, fsafemask, \
+            fexcludemask = default_target_masks(targets_data)
+
         for col in full_names:
             if col == "FBATYPE":
                 targets_data[col][:nrawtarget] = [
-                    desi_target_type(x) for x in fbtargets["DESI_TARGET"][:]]
+                    desi_target_type(x, fsciencemask, fstdmask, fskymask,
+                                     fsafemask, fexcludemask)
+                    for x in fbtargets[fcol][:]]
             elif col == "TARGET_RA":
                 targets_data[col][:nrawtarget] = fbtargets["RA"][:]
             elif col == "TARGET_DEC":
--- a/py/fiberassign/assign.py
+++ b/py/fiberassign/assign.py
@@ -66,9 +66,8 @@ results_assign_columns = OrderedDict([
     ("DEVICE_TYPE", "a3"),
     ("TARGET_RA", "f8"),
     ("TARGET_DEC", "f8"),
-    ("DESI_TARGET", "i8"),
-    ("BGS_TARGET", "i8"),
-    ("MWS_TARGET", "i8"),
+    ("FA_TARGET", "i8"),
+    ("FA_TYPE", "u1"),
     ("DESIGN_X", "f4"),
     ("DESIGN_Y", "f4"),
     ("DESIGN_Q", "f4"),
@@ -79,10 +78,8 @@ results_targets_columns = OrderedDict([
     ("TARGETID", "i8"),
     ("TARGET_RA", "f8"),
     ("TARGET_DEC", "f8"),
-    ("DESI_TARGET", "i8"),
-    ("BGS_TARGET", "i8"),
-    ("MWS_TARGET", "i8"),
-    ("FBATYPE", "u1"),
+    ("FA_TARGET", "i8"),
+    ("FA_TYPE", "u1"),
     ("PRIORITY", "i4"),
     ("SUBPRIORITY", "f8"),
     ("OBSCONDITIONS", "i4"),
@@ -214,9 +211,7 @@ def write_assignment_fits_tile(asgn, fulltarget, overwrite, params):
 
         tg_ra = np.empty(ntarget, dtype=np.float64)
         tg_dec = np.empty(ntarget, dtype=np.float64)
-        tg_desitarget = np.zeros(ntarget, dtype=np.int64)
-        tg_bgstarget = np.zeros(ntarget, dtype=np.int64)
-        tg_mwstarget = np.zeros(ntarget, dtype=np.int64)
+        tg_bits = np.zeros(ntarget, dtype=np.int64)
         tg_x = np.empty(ntarget, dtype=np.float64)
         tg_y = np.empty(ntarget, dtype=np.float64)
         tg_type = np.empty(ntarget, dtype=np.uint8)
@@ -229,13 +224,11 @@ def write_assignment_fits_tile(asgn, fulltarget, overwrite, params):
             tg_indx[tg] = indx
             props = tgs.get(tg)
             tg_obsrem[indx] = 0
-            if props.obs_remain > 0:
-                tg_obsrem[indx] = props.obs_remain
+            if props.obsremain > 0:
+                tg_obsrem[indx] = props.obsremain
             tg_ra[indx] = props.ra
             tg_dec[indx] = props.dec
-            tg_desitarget[indx] = props.desi_target
-            tg_bgstarget[indx] = props.bgs_target
-            tg_mwstarget[indx] = props.mws_target
+            tg_bits[indx] = props.bits
             tg_type[indx] = props.type
             tg_priority[indx] = props.priority
             tg_subpriority[indx] = props.subpriority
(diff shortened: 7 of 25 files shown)


tskisner:
Ok @Srheft , this branch works for me on a run combining an SV1 target file (after column type fixes) and using a sky target file from the main survey.  Here is what I ran:
```
#!/bin/bash

odir=./out
tfile=./tiles_18pRosette_DEEP2-1.fits

export DESI_LOGLEVEL=DEBUG

export OMP_NUM_THREADS=6

fba_run \
    --targets ./MTL_DEEP2_1.fits ./skies-dr7.1-PR459.fits \
    --footprint ${tfile} \
    --write_all_targets \
    --dir ${odir} \
    --overwrite

export OMP_NUM_THREADS=1

fba_run_qa \
    --footprint ${tfile} \
    --dir ${odir} \
    --prefix "fiberassign_"

mkdir -p ${odir}_plots
rm -f ${odir}_plots/*.pdf

fba_plot_results \
    --petals 5 \
    --footprint ${tfile} \
    --real_shapes \
    --dir ${odir} \
    --prefix "fiberassign_" \
    --out ${odir}_plots
```
Which produced this plot of petal 5 for tile 900005:
[fiberassign_900005.pdf](https://github.com/desihub/fiberassign/files/2981491/fiberassign_900005.pdf)

tskisner:
Another note- this branch requires the latest master version of desitarget...

Srheft:
@tskisner:

Thanks for this. Just tested the PR and I confirm the it recognizes SV masks and performs the assignment (see the plots[ here](http://www.astro.utah.edu/~u6022465/SV/18pRosette_SV_DEEP2_1/plots/) and the 100% assignment in [qa.json](http://www.astro.utah.edu/~u6022465/SV/18pRosette_SV_DEEP2_1/qa.json)).  The run was conditioned to `--sky_per_petal 80 --standards_per_petal 20`  and delivered!

Could you address  the source of the long list of WARNINGs I'm showing  in fba_run:

`WARNING: Target ID 30787645424534924 has unknown type (bitfield = 133120), skipping.
WARNING: Target ID 30787645424535083 has unknown type (bitfield = 133120), skipping.
WARNING: Target ID 30787645424536586 has unknown type (bitfield = 133120), skipping.
WARNING: Target ID 30787693742916996 has unknown type (bitfield = 133120), skipping.
WARNING: Target ID 30787693742917717 has unknown type (bitfield = 133120), skipping.
WARNING: Target ID 30787693742918603 has unknown type (bitfield = 133120), skipping.
WARNING: Target ID 30787639388933604 has unknown type (bitfield = 264192), skipping.
....
`

tskisner:
Those warnings are intentional.  There seem to be many targets in these various files with bit values that are inconsistent.  For example, in the file MTL_DEEP2_1.fits, there are targets with SV1_DESI_TARGET values of `133120` (a single bit at 2^11).  Looking in sv1_targetmask.yaml, this bit corresponds to `LRG_NORTH`, however the `LRG` bit (bit 0) has not been set.  I believe this is an error which should be fixed- if a target has one of the specialized bits set, then it should also have the general LRG / ELG / Whatever bit set.

Srheft:
I see. @geordie666, please see to it and respond here for documentation. Thanks.

tskisner:
Since I added this warning, I see there are many, many cases of this kind of thing.  For example, in the dr7.1 files there are targets with DESI_TARGET values of 512 (2^9 == QSO_COLOR), but without the QSO bit set.  Fiberassign currently relies on those "broad category" bits.  If this behavior is intentional, then we'll need to add every single one of these "specialized" bits to the mask used for checking if something is a science target.

Srheft:
@tskisner Is this PR intended to only fix the SV mask issue? I'm showing that the trouble with CMX_TARGET still standing.  Please refer to my last two email for addresses to the test files.

moustakas:
> Since I added this warning, I see there are many, many cases of this kind of thing. For example, in the dr7.1 files there are targets with DESI_TARGET values of 512 (2^9 == QSO_COLOR), but without the QSO bit set. Fiberassign currently relies on those "broad category" bits. If this behavior is intentional, then we'll need to add every single one of these "specialized" bits to the mask used for checking if something is a science target.

@tskisner I "discovered" the same feature late last year.  See the thread [here](https://github.com/desihub/desitarget/pull/426), particularly starting at [this comment](https://github.com/desihub/desitarget/pull/426#issuecomment-443800653).

tskisner:
After some small fixes to that file, I did manage to use it.  However, there are problems when reading the results back in for plotting or QA.

There is still a challenge to sort out for combining main survey files with non-main survey files.  Although the individual targets are correctly classified as science, standard, sky, safe when being read, on **output** the targets table will have both the DESI_TARGET column (from the skies) and the SV1 (or CMX) column from the other files.  When reading this table back in (for example, when plotting or doing QA), it means that one has to read both columns and choose the correct column value (and associated bitmask) to interpret what the target actually is.  Even worse is that fiberassign helpfully propagates the input DESI_TARGET values into the output directly.  I can disable that though.

I don't yet have a solution for this.  For plotting, I could ignore these and just use the internal FBATYPE, and fall back to parsing main survey DESI_TARGET values if that column does not exist (e.g. it was made by the legacy code).  For QA, I guess there is nothing to do but check for all column names and decide which mask to apply when looking for individual target types.

geordie666:
Just a note that not having the main bits set for `NORTH/SOUTH` informational bits is _entirely_ intentional. @moustakas already pointed to [one relevant thread](https://github.com/desihub/desitarget/pull/426) where @sbailey elucidates this point. We need the `NORTH/SOUTH` bits to track targets correctly where the imaging surveys overlap.

I'm surprised by @tskisner's comment that `QSO_COLOR` (2^9) is set by itself in some files, though. This is not a `NORTH/SOUTH` bit so shouldn't be informational. I'm hoping that this is actually 2^9 in the [main survey bitmask](https://github.com/desihub/desitarget/blob/master/py/desitarget/data/targetmask.yaml#L17MAIN survey bitmask), which would correspond to `ELG_NORTH`, an informational bit that won't necessarily have `ELG` set for a source that passed the "northern" targeting cuts but is in the southern footprint.

tskisner:
Ok, thanks for clarifying.  It sounds like these bit combinations are legitimate and simply mean that those targets should not be included in the assignment.  I'll remove the warning.

tskisner:
Summarizing recent offline and online discussions about what is needed to close this out:

1.  OBSCONDITIONS will be required for all target and tile files.  A missing column will raise an exception.

2.  The survey type ("main", "cmx", "sv1", etc) as identified by desitarget utilities will be required to be the same for all input target files for a given run of fiber assignment.  If using a totally custom survey, the same column name and bitmask definitions will be required for all input target files.

3.  Additional sanity checks will be added on input target files to check that the column types are correct.

4.  The raw fiberassign output files will have a header key identifying the survey type of the input target files.

5.  The target bitfield will be propagated into the per-target properties for each target available to the C++ code.  The raw fiberassign output files will have a column with these input target bits.  These bits along with the survey keyword will allow reconstruction of the target properties using desitarget utilities.

6.  The merged fiberassign outputs will additionally propagate all the target columns from the input (DESI_TARGET, BGS_TARGET, CMX_TARGET, etc).

7.  When reading output FITS files for purposes of visualization or QA, a Targets object can be constructed either from the raw format (target bitfield + survey keyword), or from the merged format (calling desitarget utilities on the merged table columns).

I'll implement these changes and push to this branch for review / testing.

sbailey:
Good for items 1-4.

> 5. The target bitfield will be propagated into the per-target properties for each target available to the C++ code. The raw fiberassign output files will have a column with these input target bits. These bits along with the survey keyword will allow reconstruction of the target properties using desitarget utilities.

Do we need to propagate the target bitfield into the C++ code?  The only case I know where that would be needed would be if we are doing a per-bit depth-vs-breadth choice for how to handle NUMOBS_MORE, but we decided to first try depth-first and quantify the impact on Lyman-alpha QSOs (which are high priority and sparse so are likely to get their requested number of obs regardless of depth vs. breadth).  i.e. let's not do that python-C++ surgery until we are really sure it is necessary.

> 6. The merged fiberassign outputs will additionally propagate all the target columns from the input (DESI_TARGET, BGS_TARGET, CMX_TARGET, etc).

> 7. When reading output FITS files for purposes of visualization or QA, a Targets object can be constructed either from the raw format (target bitfield + survey keyword), or from the merged format (calling desitarget utilities on the merged table columns).

Clarifying: For the main survey the raw format target bitfield is DESI_TARGET?  i.e we can distinguish ELG, LRG, QSO, SKY, various STD, BGS_ANY, and MWS_ANY, but we can't distinguish subtypes of BGS and MWS that use the BGS_TARGET and MWS_TARGET fields, correct?  I think that is ok.

tskisner:
The reason for passing the int64 target bitfield all the way through the C++ code is:

1.  For optional use in find_best() when considering which target to select.

2.  For output in the raw data files so that QA can distinguish sub-types of "science".  This is really useful instead of having to do the merging step before getting this info.

I'm not sure about the statement of "python-C++" surgery...  The current code already propagates this information but it is specific to the main survey.  I'm just going to pass in the corresponding target bitfield for the other surveys and dump it out to a generic column (e.g. FBATARGET) in the raw data (rather than dumping to a column named "DESI_TARGET", even for cmx and sv1).  This generic column has whatever the input values were for the particular survey type.  After merging, the original columns (DESI_TARGET, BGS_TARGET, MWS_TARGET, etc) would get added.

sbailey:
Currently the low-level fiberassign_*.fits format FASSIGN has DESI_TARGET, BGS_TARGET, and MWS_TARGET.  Are you replacing those with a single FBATARGET, which is either DESI_TARGET, SV1_DESI_TARGET, or CMX_TARGET depending upon the original inputs, and then not including BGS_TARGET and MWS_TARGET in the fiberassign_*.fits files?

> 1. For optional use in find_best() when considering which target to select.

I'm hoping we won't need that, and didn't want you to do extra work to support this until we are sure it is needed.

> 2. For output in the raw data files so that QA can distinguish sub-types of "science". This is really useful instead of having to do the merging step before getting this info.

I agree it is useful to have the subtypes in the low-level output for QA.  It seems it is a matter of taste about whether it is easier/desirable to propagate data from python -> C++ -> python just to include it in the output even if it was never used inside the C++, vs. having a "merge light" to add columns to the low-level output that were never needed by C++.

tskisner:
About the target columns- exactly.  Currently we actually propagate **all three** of DESI_TARGET, BGS_TARGET, and MWS_TARGET through to the raw output.  I'm proposing taking "whatever" column was used to determine the type on input (DESI_TARGET, CMX_TARGET, SV1_DESI_TARGET) and writing that out in the raw file as FBATARGET (along with a header keyword with the survey).  Then the original columns will be automatically copied back in during the merging step as just any other auxiliary column.

So, I'm actually proposing to propagate fewer data fields than is currently done, and make the output a generic name that is not specific to the main survey.

Also, the current code already has access to the target bits inside find_best(), even though that information is not currently used.

sbailey:
OK, thanks for clarifying the various target bit propagations.  +1 to proceed.

Srheft:
@tskisner @sbailey:


Speaking of the output content, two suggestions (no need to include the changes in this PR if it involves more than minimal effort/code change).

- If FBATARGET column eventually contained target bits of CMX/SV/MAIN survey, I suggest to have it as FA_TARGET to resemble DESI_TARGET, CMX_TARGET, SV1_DESI_TARGET.

- Upon agreement during the workshop, we will launch one large html table of tiles that takes the user to individual pages for the assignment statistics table per tile and it'd be greatly useful if there was a **3rd column in FAVAIL** HDU that would characterize [via a bit or a string keyword] the reason why any of the available TARGETS in the second column of FAVAIL are not getting picked up [so that we would color-code them based on those reasons like it was done in eBOSS in plots like below]. The DESI_DEBUG_TARGET feature is nice for diving into detail but it's not practical to run in bulk just to get that info.

@kdawson1000 could fill in more as to the significance of having this info reported/displayed. I concede that it has been easier to derive this info in eBOSS setting and the final vote might be to forego this or find alternative info to display. We shall decide on stuff like this before we solidify the fiberassign output format/content. 



image

sbailey:
@Srheft second request (why something didn't get assigned) is issue #182.  Let's address that separately and not block this PR by trying to also add that feature (which isn't completely trivial).

Srheft:
Sure. I just wanted to throw it out ther; hence the
 > (no need to include the changes in this PR if it involves more than minimal effort/code change)]. 

Truthfully, even if I was aware of #182 , I wouldn't know if that's the right place or an entirely separate ticket must be issued :)

tskisner:
@Srheft , I have no preference on the name- we can use something with *_TARGET if that is better.  The tracking of "why an available target was not assigned to a fiber" is already in issue #182.  That will be in a separate PR.

tskisner:
Sorry- replied before I saw the response from @sbailey ...

tskisner:
I believe this PR is now ready.  In the raw outputs, the column names for fiberassign-specific type and generic target bits are "FA_TARGET" (as requested by @srheft) and "FA_TYPE" (to be consistent with FA_TARGET).  The merged data has all the original columns from the inputs.  There is also now an "FA_SURV" header keyword in the output.

I cleaned up more of the targets.py functionality, centralizing the loading of target tables and their identification- either through the use of the FA_SURV header key and the FA_TARGET column or through the use of desitarget functions to identify the survey from the column names.

In order to use CMX files with this code there are these requirements:

1.  The latest desitarget master branch.

2.  The "MTL" file (which in the example I have seen contained one bright target with tiles dithered around it) must have the column CMX_TARGET and the bits must be set to one of the cmx defined bits (STD_GAIA|SV0_STD_BRIGHT|STD_TEST|STD_CALSPEC).

3.  The sky file must be converted to "cmx" survey type.  This involves changing the name of the DESI_TARGET column to CMX_TARGET.  The obsconditions for this file was 65535, so that is compatible with all tiles that I have seen.

4.  The MTL file target obsconditions must be compatible with the tile obsconditions.  In these example files, both had the value of "4".  Note that obsconditions must be an integer and there are now checks for this.

I tested this on both the dr7.1-lite main survey files, and also the CMX files Sarah gave me with the modifications to those files listed above.  Here is a plot of tile 200005 from that dataset with the one target assigned and the rest set to sky or safe:
[fiberassign_200005.pdf](https://github.com/desihub/fiberassign/files/3041958/fiberassign_200005.pdf)

sbailey:
Thanks @tskisner.  @Srheft please confirm that this works for what you need it to do, or if not what is broken / missing / unclear.

Srheft:
Created MTL for a set of SV target list via `desitarget.mtl.make_mtl(targetdata)` and made sure the header key 'SURVEY' exists in the header of the MTL file and it says 'sv1'; yet fiberassign complains about a mismatch between survey types.  Running command and input files are includeed below. 

Running:

 fba_run --targets ./[MTL_04Apr2019.fits](http://www.astro.utah.edu/~u6022465/mtl_test/MTL_04Apr2019.fits) ./[33-40--3-3-skies-dr7.1-0.27.0.fits](http://www.astro.utah.edu/~u6022465/mtl_test/33-40--3-3-skies-dr7.1-0.27.0.fits) ./[33-40--3-3-std-sv1-targets-dr7.1-0.27.0.fits](http://www.astro.utah.edu/~u6022465/mtl_test/33-40--3-3-std-sv1-targets-dr7.1-0.27.0.fits) --sky_per_petal 80 --standards_per_petal 20 --footprint ./[tiles_18pRosette_DEEP2-1.fits](http://www.astro.utah.edu/~u6022465/mtl_test/tiles_18pRosette_DEEP2-1.fits) --write_all_targets --dir ./FAout_rosette/ --overwrite

I get:
`
INFO: Reading fiber positions from /uufs/astro.utah.edu/common/home/u6022465/desi/software/desimodel/data/focalplane/fiberpos-all.fits
INFO: Target file ./MTL_04Apr2019.fits has 262063 rows.  Reading in chunks of 1000000
INFO: Target table using survey 'sv1', column SV1_DESI_TARGET:
INFO:   sciencemask LRG|ELG|QSO|BGS_ANY|MWS_ANY|SECONDARY_ANY
INFO:   stdmask     STD_FAINT|STD_WD|STD_BRIGHT
INFO:   skymask     SKY
INFO:   safemask    BAD_SKY
INFO:   excludemask BRIGHT_OBJECT|IN_BRIGHT_OBJECT
INFO: Read target file ./MTL_04Apr2019.fits:  0.52 seconds (1 calls)
INFO: Target file ./33-40--3-3-skies-dr7.1-0.27.0.fits has 96768 rows.  Reading in chunks of 1000000
INFO: Target table using survey 'main', column DESI_TARGET:
INFO:   sciencemask LRG|ELG|QSO|BGS_ANY|MWS_ANY|SECONDARY_ANY
INFO:   stdmask     STD_FAINT|STD_WD|STD_BRIGHT
INFO:   skymask     SKY
INFO:   safemask    BAD_SKY
INFO:   excludemask BRIGHT_OBJECT|IN_BRIGHT_OBJECT
ERROR: Targets object has survey type "sv1", cannot append data from survey type "main"
  File "/uufs/astro.utah.edu/common/home/u6022465//desi/software/fiberassign/bin/fba_run", line 20, in <module>
    main()
  File "/uufs/astro.utah.edu/common/home/u6022465//desi/software/fiberassign/bin/fba_run", line 15, in main
    run_assign_full(args)
  File "/uufs/astro.utah.edu/common/home/u6022465/.local/lib/python3.5/site-packages/fiberassign-1.0.0.dev2366-py3.5-linux-x86_64.egg/fiberassign/scripts/assign.py", line 271, in run_assign_full
    hw, tiles, tgs = run_assign_init(args)
  File "/uufs/astro.utah.edu/common/home/u6022465/.local/lib/python3.5/site-packages/fiberassign-1.0.0.dev2366-py3.5-linux-x86_64.egg/fiberassign/scripts/assign.py", line 249, in run_assign_init
    excludemask=args.excludemask)
  File "/uufs/astro.utah.edu/common/home/u6022465/.local/lib/python3.5/site-packages/fiberassign-1.0.0.dev2366-py3.5-linux-x86_64.egg/fiberassign/targets.py", line 634, in load_target_file
    excludemask=excludemask)
  File "/uufs/astro.utah.edu/common/home/u6022465/.local/lib/python3.5/site-packages/fiberassign-1.0.0.dev2366-py3.5-linux-x86_64.egg/fiberassign/targets.py", line 564, in load_target_table
    stdmask, skymask, safemask, excludemask)
  File "/uufs/astro.utah.edu/common/home/u6022465/.local/lib/python3.5/site-packages/fiberassign-1.0.0.dev2366-py3.5-linux-x86_64.egg/fiberassign/targets.py", line 407, in append_target_table
    d_subprior, d_obscond, d_type)
RuntimeError: Targets object has survey type "sv1", cannot append data from survey type "main"
`

tskisner:
@srheft, it looks like the sky file has not been converted to an "SV1 sky file".  In the long term, we should modify whatever tool is used to generate the sky file so that it can generate a sky file for a particular survey.  Is that script in desitarget?  Perhaps @geordie666 knows more.  This tool should be changed to accept the survey name, and then use the corresponding column name **and** the corresponding desi_mask / sv1_mask / cmx_mask bits to indicate sky / bad_sky.

Currently the sky / bad_sky bit definitions are identical between all the surveys.  Because of this coincidence, a workaround is to copy the main survey sky file to a new one, and then just change the name of the column.  This is what I did for my tests.  There is probably some easier way to do this, but I read in the data, changed the dtype field name and wrote it out to a new file.  This requires 2 copies of the (7GB!) sky file in memory.  Here is the script I used:
```
#!/usr/bin/env python3

import os
import sys

import fitsio

if len(sys.argv) != 4:
    print("Usage: {} <in file> <out file> <new column name>".format(sys.argv[0]))
    sys.exit()

infile = sys.argv[1]
outfile = sys.argv[2]
newcol = sys.argv[3]

if os.path.isfile(outfile):
    os.remove(outfile)

fd = fitsio.FITS(infile, "r")
ofd = fitsio.FITS(outfile, "rw")

header = fd[1].read_header()
data = fd[1].read()
nrows = fd[1].get_nrows()

oldnames = data.dtype.names
newnames = []
for nm in oldnames:
    if nm == "DESI_TARGET":
        newnames.append(newcol)
    else:
        newnames.append(nm)

data.dtype.names = tuple(newnames)
        
ofd.write(None, header=header, extname="PRIMARY")
ofd.write(data, header=header, extname="SKIES")

ofd.close()
fd.close()
```
Run with (for example):
```
./convert_sky.py skies-dr7.1-PR459.fits skies-dr7.1-PR459_cmx.fits CMX_TARGET
```
However, this only works because the sky bits are defined the same between the surveys.  The better solution is to generate a survey-specific sky file which uses the arbitrary bits defined by the survey-specific bitmask.

geordie666:
@tkisner: There will never be a survey-specific sky file. The algorithm to generate sky locations is the same regardless of `main`, `cmx` or `sv1`. Is it possible just to hack in a line of code that says "if this is a `skies-` file, and a `targets-` file exists, change the `dtype.names` for the `_TARGET` columns to the same strings as for the `targets-` file?

Otherwise we'll be carrying around an extra (quite large) file for no particular reason.

I intend to always reserve the same bits for `SKY` and `BADSKY` for the three surveys, as is the case now.

tskisner:
Ok, if the sky bits are always the same between different surveys, then the "workaround" above will always be ok.  Part of the point of this branch is to remove logic from fiberassign that tries to "guess" what the user is intending and to actually force the input files to all use the same bitmask definitions from desitarget.  Will a sky target file ever have other bits defined besides SKY and BAD_SKY?  Should fiberassign be responsible for verifying that any main survey targets that are loaded have only bit definitions that are also consistent the program of other target files?  

This kind of logic is error prone and confusing to the user, and has been a persistent source of lost productivity.  I feel strongly that the identification of target bits should come entirely from the input files and definitions in desitarget.

So- we can take the code snippet above and use it to convert any main survey sky file into files compatible with the other survey types.  We should check that in to whatever git repo is appropriate.

sbailey:
Since the skies file already has a different format from the target files (e.g. having APFLUX_G/R/Z but not FLUX_G/R/Z), how about making it a different argument to fiber assign instead of treating it as just another target file, e.g. `fba_run --targets blat.fits foo.fits --skies skies.fits ...`.  This is similar to original fiberassign.  Then the code can know that the `--skies` input will always use `DESI_TARGET`, and the other target files can derive on-the-fly for whether to use `DESI_TARGET`, `SV1_DESI_TARGET`, or `CMX_TARGET` (while still requiring that they be consistent if more than one target file is provided).

tskisner:
Ok, I can make those changes.  It requires passing some kind of "force sky" override down into the function where the table data is identified by desitarget's main_cmx_or_sv() function.

Srheft:
Thanks @tskisner.  

- Converted the sky file `(./convert_sky.py skies-dr7.1-PR459.fits 33-40--3-3-skies-dr7.1-0.27.0.fits SV1_DESI_TARGET)`.

- Removed the standard star file from the command as it does not contain OBSCONDITIONS 

- Ran this: 
         fba_run --targets ./[MTL_04Apr2019.fits](http://www.astro.utah.edu/~u6022465/mtl_test/MTL_04Apr2019.fits) ./[33-40--3-3-skies-dr7.1-0.27.0.fits](http://www.astro.utah.edu/~u6022465/mtl_test/33-40--3-3-skies-dr7.1-0.27.0.fits)  --sky_per_petal 80 --standards_per_petal 20 --footprint ./[tiles_18pRosette_DEEP2-1.fits](http://www.astro.utah.edu/~u6022465/mtl_test/tiles_18pRosette_DEEP2-1.fits) --write_all_targets --dir ./FAout_rosette/ --overwrite

- [Full screen log](https://github.com/desihub/fiberassign/files/3050087/apr04.txt) is attached and I ended up seeing [this](http://www.astro.utah.edu/~u6022465/4epfl/FAout/fiberassign_900008.pdf).  Change the tile number in the browser to see all the 0-17 pdf plots.  Skies are not getting picked up [could you examine my inputs?].

 - Although from fiberssign perspective, one can pass as many files to --targets, We may need to emphasize that people don't pass any target file without the OBSCONDITIONS column. OR shall we ask desitarget people to have OBSCONDITIONS in their target specific catalogs as well? if there is going to be class specific (e.g., std.fits, qso.fits, ...) catalogs for dr8+.

Srheft:
@tskisner please check my inputs as both my cmx and sv example runs show no sky assignments [converetd sky catalogs are linked in both cmx and sv cases]. OBSCONDITIONS is set to 65535 in all input files. You had a successful cmx example so there must be a problem with my inputs that no sky gets assigned. 

[This](http://www.astro.utah.edu/~u6022465/cmx_example/plots/) is the assignment I get when I run this cmx example:
 
fba_run --targets ./[MTL_allSTDs.fits](http://www.astro.utah.edu/~u6022465/cmx_example/MTL_allSTDs.fits) ./[skies-dr7.1-PR459_cmx.fits](http://www.astro.utah.edu/~u6022465/cmx_example/skies-dr7.1-PR459_cmx.fits) --footprint ./[tiles_allSTDs.fits](http://www.astro.utah.edu/~u6022465/cmx_example/tiles_allSTDs.fits) --dir ./

tskisner:
I think this is now ready to merge, unless anyone objects.

Srheft:
Please let me try a few things. Won't take long.

Srheft:
@tskisner 
running the script above would give an error as if the --sky option is not added:

fba_run: error: ambiguous option: --sky could match --skymask, --sky_per_petal

Srheft:
Thanks for waiting. Had to remove and re-fetch the entire issue_181 branch to get the new changes in the build. 
I confirm that using the new --sky feature, skies are assigned in both CMX and SV  example runs. Plots are [here](http://www.astro.utah.edu/~u6022465/cmx_example/plots/) and [here](http://www.astro.utah.edu/~u6022465/mtl_test/plots/).

PR #202, opened 2019-05-29
allow fiberassign --stdstar duplicates with --mtl

### LINKED ISSUE #198: Inconsistent behaviour for duplicate IDs across Science targets & standards
survey-tiles file (single tile with ID 100000):

image

Footprint file (Kyle's BGS tiling):

image

Call:

```
fiberassign 

--mtl /global/cscratch1/sd/mjwilson/svdc-summer2018/targets/mtl.fits 
--stdstar /global/cscratch1/sd/mjwilson/svdc-summer2018/targets/standards-dark.fits 
--sky /global/cscratch1/sd/mjwilson/svdc-summer2018/targets/sky.fits  
--surveytiles ~mjwilson/desi/survey-validation/svdc-summer2018/bgs-gama-tiles-kdawson_100000.txt  
--footprint ~mjwilson/desi/survey-validation/svdc-summer2018/bgs-gama-tiles-kdawson.fits 
--fibstatusfile /global/homes/m/mjwilson/desi/survey-validation/svdc-summer2018/fiberstatus.ecsv 
--outdir /global/cscratch1/sd/mjwilson/svdc-summer2018/fiberassign
```

Error message: 

image

Failure to over-ride default footprint with --footprint?  These directories have desi permissions for reproducing the error. 

Background (file production) is here:
https://github.com/michaelJwilson/survey-validation/blob/master/svdc-summer2018/bgs-gama-pipeline.ipynb

but I'm running fiberassign in the shell to test.

sbailey:
This PR fixes #198 by allowing fiberassign to have duplicate targets in the `--stdstar` file and the `--mtl`.  It also exposures the fba_run options `--sciencemask`, `--stdmask`, etc. to fiberassign.

Background: Legacy `fiberassign` allowed duplicate std/mtl targets and it worked due to how it did internal bookkeeping.  The refactored `fba_run` does not allow duplicate targets, but that broke some of the ways we had been using it.

This PR: If all `--stdstar` targets are already in MTL, it just ignores the file.  If it has some unique targets, those are filtered out and written to a separate temporary file before passing on to `fba_run`.  This allows users to continue to use `fiberassign` in the same manner as before, without worrying about std/mtl duplicates.

### DIFF
--- a/bin/fiberassign
+++ b/bin/fiberassign
@@ -5,7 +5,7 @@ Wrapper script for fba_run and fba_merge to provide script API compatibility
 with the legacy branch of fiberassign.
 """
 
-import argparse
+import argparse, os, sys, tempfile
 
 import fiberassign
 
@@ -43,8 +43,33 @@ parser.add_argument("--surveytiles",
 parser.add_argument("--outdir",
                     help="output directory (default = ./)", default="./")
 
+parser.add_argument("--sciencemask", required=False,
+                    default=None,
+                    help="Default DESI_TARGET mask to use for science "
+                         "targets")
+
+# DEPRECATED; use --stdmask instead
 parser.add_argument("--starmask",
-                    help="integer mask defining standard stars")
+                    help="DEPRECATED: use --stdmask instead")
+
+parser.add_argument("--stdmask", required=False,
+                    default=None,
+                    help="Default DESI_TARGET mask to use for stdstar "
+                         "targets")
+
+parser.add_argument("--skymask", required=False,
+                    default=None,
+                    help="Default DESI_TARGET mask to use for sky targets")
+
+parser.add_argument("--safemask", required=False,
+                    default=None,
+                    help="Default DESI_TARGET mask to use for safe "
+                    "backup targets")
+
+parser.add_argument("--excludemask", required=False,
+                    default=None,
+                    help="Default DESI_TARGET mask to exclude from "
+                    "any assignments")
 
 parser.add_argument("--rundate", help="run date [YYYY-MM-DD]")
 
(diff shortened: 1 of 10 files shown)


sbailey:
Hmm; a bunch of Travis test failures, but tests pass at NERSC.  Newer fitsio is first suspicion since it made some non-backwards compatible changes.

sbailey:
fitsio 1.0.x switched to reading string table columns as unicode strings instead of raw bytes, requiring tests like
```python
    science_rows = np.where(fiber_data["DEVICE_TYPE"] == b"POS")[0]
```
to be changed to
```python
    science_rows = np.where(fiber_data["DEVICE_TYPE"].astype(str) == "POS")[0]
```
to work with both 0.9.x and 1.0.x

Now I'm getting intermittent unicode parsing crashes from fitsio, e.g.
```
======================================================================
ERROR: test_io (fiberassign.test.test_assign.TestAssign)
----------------------------------------------------------------------
UnicodeDecodeError: 'ascii' codec can't decode byte 0xcb in position 0: ordinal not in range(128)

The above exception was the direct cause of the following exception:

SystemError: <class 'UnicodeDecodeError'> returned a result with an error set

[repeated many times]

The above exception was the direct cause of the following exception:

  File "/home/travis/build/desihub/fiberassign/py/fiberassign/test/test_assign.py", line 154, in test_io
    read_assignment_fits_tile((tid, infile))
  File "/home/travis/build/desihub/fiberassign/py/fiberassign/assign.py", line 621, in read_assignment_fits_tile
    fbassign = fd["FIBERASSIGN"].read()
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/fitsio/hdu/table.py", line 604, in read
    data = self._read_all(**keys)
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/fitsio/hdu/table.py", line 649, in _read_all
    array = self._maybe_decode_fits_ascii_strings_to_unicode_py3(array)
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/fitsio/hdu/table.py", line 1341, in _maybe_decode_fits_ascii_strings_to_unicode_py3
    array = array.astype(new_dt, copy=False)
SystemError: <class 'UnicodeDecodeError'> returned a result with an error set
```
I'm now trying to catch it in the act and cache the problematic test file to see if it is a problem with fitsio or with something we're writing.  But it doesn't fail every time the tests run...

sbailey:
Unicode parsing errors are our fault.  [assign.py line 859](https://github.com/desihub/fiberassign/blob/b29e3c15117da57d1dca7fbb9a19f7f6936f5e8f/py/fiberassign/assign.py#L859) creates an empty table to fill with fiber assignments:
```
    tile_targets = np.empty(len(tile_tgids), dtype=tile_targets_dtype)
``` 
That includes a BRICKNAME string column that never gets initialized, and thus randomly sometimes contains non-ASCII characters which aren't allowed in FITS.

For this particular case, I think replacing it with `np.zeros` and also actually filling in the brickname would fix this problem.  However, there are multiple other uses of `np.empty` throughout fiberassign, so pausing for a moment to consult with @tskisner .  I'd prefer to take the minor performance hit and initialize these other cases with `np.full(length, bogus_value, dtype)`, and then include checks at the end of initialization that no bogus values remain.  Thoughts?

sbailey:
FYI I'm proceeding with my proposed replacement of `np.empty` -> `np.zeros` and `np.full`.

sbailey:
For the record: I handled the uninitialized variables case, but I didn't add the (slow) code to fill in BRICKNAME in this PR.

Issue #184, opened 2019-02-21
bump targets based on their priority+subpriority, not on the sky/std subpriority


sbailey:
From #140:

> 3. When bumping science targets for standards or sky, both legacy and current code use the priority of the standard or sky to set the order.  The fibers are then looped over for each priority class from low to high.  This means that the fiber looping order will be random within a tile and this means that targets in a given priority class will be bumped randomly based on location on the tile.  Like this pseudocode:

```
foreach priority class (e.g. 1500 ... 3200):
    foreach **sky or standard** in subpriority (i.e. random) order:
        foreach fiber that can reach this sky or standard:
            if fiber is assigned to target of specified priority class:
                bump science target
                assign sky / standard
                try to reassign bumped science target
```
> Or if you prefer the actual code:
https://github.com/desihub/fiberassign/blob/master/src/assign.cpp#L843

Although this is random, it couples the bumping choice to the subpriority of the sky/std rather than to just the priority+subpriority of the science targets.  This means that a high subpriority ELG might get bumped before a low subpriority ELG, based upon the subpriorities of the standards/skies covering them.

I think it would be better to have the order of bumping be based only on the science target priority+subpriority, e.g. for bumping science targets to make way for sky:
```
for each sorted priority_plus_subpriority:
    if this target is covered by a sky:
        bump science target
        assign sky
        try to reassign bumped science target (see #176)
        if we have reached desired sky density:
            break
```

Related to #183 which requires a similar update in logic for looping over targets based upon priority+subpriority (from high to low in that case) instead of fibers.

kdawson1000:
This code/pseudocode is not what I had originally thought when I raised the concern.  It looks like the first loop is indeed on the priority of the science target, which is definitely what we want.  If I understand this correctly, the lowest priority science targets are getting bumped first when there is a need for sky/std fiber assignments.  The order within a target class is randomized not by the SUBPRIORITY of the target classes but by SUBPRIORITY of the science targets.  

This is pretty close to the goal, I believe.  One may want to loop on PRIORITY first, so we can account for the need for multiple observations.  But that is likely a fine detail.

tskisner:
This issue is now fixed in #258.

PR #204, opened 2019-06-02
Use LOCATION instead of FIBER

tskisner:
Background
----------------------------

During fiberassign, we are actually assigning theta/phi arm positions to devices at particular locations on the focalplane.  Where those positioners are connected on spectrographs should not matter to fiberassign.  However, due to historical reasons, the FIBER value has been used throughout the code as a primary key for indexing things.  This PR changes that behavior to use the unique (and fixed to the focalplane) "device location" as the primary indexing variable.

Consequences
--------------------------

Great care has been taken to ensure that these many small internal changes have no impact on the output files.  I several places there are known loops over things in fiber ID order, and I have implemented the same loop in those cases (those places are the subject of open tickets already).  There are only a few user-facing differences:

- The debugging environment variable `DESI_DEBUG_FIBER` is changed to `DESI_DEBUG_LOCATION`.
- When making plots, the assignments are labeled with the device location, not the fiber ID.

Currently the FIBER and SPECTROGRAPH data from the input fiberpos files are simply propagated to the outputs.  In the future, we may remove that info from the hardware description (see https://github.com/desihub/desimodel/pull/105).  After this branch is merged, fiberassign will be ready to accomodate that change easily.

TO-DO
------------------

The documentation needs updated in a few places before merging.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -6,13 +6,16 @@ fiberassign change log
 1.0.3 (unreleased)
 ------------------
 
-* No changes yet.
+* Fix an issue with reproducibility of the ordering of available tile-fibers
+  for each target (PR `#203`_).
+
+.. _`#203`: https://github.com/desihub/fiberassign/pull/203
 
 1.0.2 (2019-05-30)
 ------------------
 
 * PR `#202`_:
-  
+
   * Gracefully allow fiberassign --stdstar to have duplicates with --mtl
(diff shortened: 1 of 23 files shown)


tskisner:
Ok, the performance of this branch is now comparable to master, with the exception of forced assignment of sky targets.  That slowdown is due to mapping back and forth from location to fiber purely to reproduce identical outputs to master.  This will go away when #184 is addressed.

PR #206, opened 2019-06-24
Remove collision check "shortcut"

tskisner:
This shortcut in the code (to avoid doing an explicit collision calculation) was left over from the original codebase.  As part of debugging the focalplane format, I discovered that there were a small number of cases where doing the full calculation gave different results than the "shortcut".  This small PR removes the shortcut.

Note that the development in the "hwstate" branch had already removed this kind of shortcut, since it is not compatible with realistic positioner geometries.

### DIFF
--- a/py/fiberassign/targets.py
+++ b/py/fiberassign/targets.py
@@ -55,7 +55,7 @@ def default_main_sciencemask():
     sciencemask |= desi_mask["QSO"].mask
     sciencemask |= desi_mask["BGS_ANY"].mask
     sciencemask |= desi_mask["MWS_ANY"].mask
-    sciencemask |= desi_mask["SECONDARY_ANY"].mask
+    sciencemask |= desi_mask["SCND_ANY"].mask
     return sciencemask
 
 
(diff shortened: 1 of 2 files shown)


tskisner:
Hmmm, this looks like maybe some upstream change to desitarget is causing a failure?  Investigating.

geordie666:
'SECONDARY_ANY' was renamed 'SCND_ANY' based on a request from @sbailey a few months ago. I think that's the issue.

py/fiberassign/targets.py
```diff
@@ -107,7 +107,7 @@ def default_sv1_sciencemask():
     sciencemask |= sv1_mask["QSO"].mask
     sciencemask |= sv1_mask["BGS_ANY"].mask
     sciencemask |= sv1_mask["MWS_ANY"].mask
-    sciencemask |= sv1_mask["SECONDARY_ANY"].mask
+    sciencemask |= sv1_mask["SCND_ANY"].mask
```
sbailey: Apologies for the non-backwards compatible change.  To support new-fiberassign on old-targetfiles, let's do
```python
if "SCND_ANY" in sv1_mask.names():
    sciencemask |= sv1_mask["SCND_ANY"].mask
else:
    sciencemask |= sv1_mask["SECONDARY_ANY"].mask
```
and ditto for `desi_mask`

tskisner: Ok, no problem.

Issue #183, opened 2019-02-21
update order of initial assignment to prevent collisions from low priority targets


sbailey:
From #140:

> The initial assignment of unused fibers has always looped over fibers in order based on the total priority of the targets available to the fibers.

Consider this case:

fiber | targets covered (A=highest priority to C=lowest priority)
---- | ----
1 | A
2 | A C
3 | B (but would collide with fiber 2 if fiber 2 is assigned to C)

Fiber 1 grabs A, fiber 2 is considered next because it covered A but that was already taken so it gets assigned to low priority C, which creates a collision such that 3 cannot reach B.  i.e. the presence of a low priority target in the list could result in a higher priority target not getting assigned.

Suggested alternative:

* Loop over targets in descending order of priority+subpriority.
* For each target, assign it to the lowest numbered unassigned fiber that covers it
  * this could be optimized a bit, e.g. prioritizing fibers with fewer options

I think this means that collisions preventing assignments would only happen with higher priority targets and never with lower priority targets.

i.e. the core logic shift is looping over targets and considering what fibers could be assigned, instead of looping over fibers and considering what targets could be assigned.

forero:
I wonder whether assigning the target to the closest fiber would be useful as well. That should decrease the positioner collision rate.

sbailey:
@forero years ago we discussed having each fiber pick its closest target, but decided against that because it might cause subtle small scale assignment efficiency variations (targets closest to fibers get picked at higher efficiency than those on the edges).  This case is a little different (picking positioners near targets instead of picking targets near positioners) but similar caution might apply.

forero:
@sbailey there already tons of systematic efficiency variations. Many can be picked up by pair-weighting probabilities. Some others are trickier, i.e. low priority targets are less complete in regions around high priority targets (I don't know how to correct for that). This suggestion adds a new one to the list, but a least it should also be picked up by pair-weighting probabilities and should lead to higher completeness for science targets.

sbailey:
Fair enough.  When we implement this, we could try both:

1. pick the lowest available fiber number that covers the target
2. pick the closest available fiber that covers the target

and see how much (2) improves the overall efficiency by reducing collisions.

tskisner:
[...]

Issue #179, opened 2019-02-15
Fiber order of science target redistribution


tskisner:
Currently when redistributing science targets (to load balance based number of assignments per petal), the targets are considered in fiber ID order.  This could imprint on the final results, so we should instead consider assigned targets for reassignment based on their priority.

tskisner:
This is fixed in #258.

Issue #176, opened 2019-02-14
Recursive bumping of science targets


tskisner:
This ticket is tracking conversations about the way that science targets are "bumped" in order to place standards and sky.  This is **not** proposing any change to the algorithm for deciding which original targets to bump, but rather what to do after we have bumped a target.

Currently, when a science target is "bumped" in order to place a standard or sky, an attempt is made to re-assign that science target to an empty fiber on a later tile.  If all available future tile/fiber placements are already assigned (even if assigned to lower priority targets), then the bumped target simply loses an observation.

Under this proposal, if a bumped science target can be placed on a future tile/fiber that is assigned to a lower priority target, we would move the original target and bump the lower priority target.  This would be repeated until the current target found an empty tile/fiber, or until there were no lower-priority targets to bump.

Note:  this change would have no impact in the case of running fiber assignment one tile at a time (which is another strategy under testing / consideration), it only impacts the case of collectively doing the assignment of many tiles at once.

Considering the typical number of available tile/fibers for a given target, **and** considering that we bump lowest-priority science targets first, the number of occurrences where the bumped science target finds an even-lower-priority target on a later tile to further bump is probably small.

The question at hand is whether implementing the above behavior is worth an attempt.

PR #207, opened 2019-06-24
Use a dynamic focalplane model

tskisner:
This work removes many hard-coded properties and makes use of the newly-added desimodel.io.load_focalplane() to retrieve all positioner properties and the exclusion polygons to use for each positioner.

Opening this PR to more easily track comments and changes.  Do not merge yet.

### DIFF
--- a/py/fiberassign/assign.py
+++ b/py/fiberassign/assign.py
@@ -35,7 +35,8 @@ from .utils import Logger, Timer, default_mp_proc
 from .targets import (TARGET_TYPE_SKY, TARGET_TYPE_SAFE, desi_target_type,
                       default_target_masks, default_survey_target_masks)
 
-from .hardware import (FIBER_STATE_STUCK, FIBER_STATE_BROKEN)
+from .hardware import (FIBER_STATE_UNASSIGNED, FIBER_STATE_STUCK,
+                       FIBER_STATE_BROKEN)
 
 from ._internal import Assignment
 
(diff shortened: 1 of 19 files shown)


tskisner:
Ok, this branch is now re-based against master (which was just tagged 1.0.4).

tskisner:
I believe this branch is ready to merge.  This code currently **requires** a recent version of desimodel (>= 0.9.2).  In order to use the lastest hardware model, you also need the current trunk of the desimodel data svn.  Desimodel svn now includes two hardware models (at least one more will be added in the coming weeks).  The run time passed to `fba_run` (default is current time) determines which hardware model is selected.  To use the old model, select a date prior to 2019-09-16.  Here is a comparison of the old and new focalplanes showing the positioners at the zero point of their theta range and at the zero angle of their phi range plus PI.  Also plotted are the default exclusion polygons:

[test_focalplane_compare.pdf](https://github.com/desihub/fiberassign/files/3633288/test_focalplane_compare.pdf)

In addition to the unit tests, I have run this on about 300 tiles of DR7.1 comparing the output of master with this branch using the new hardware model.  In that comparison there are noticeably fewer science targets assigned per petal with this new hardware model.  This is likely due to:

1.  The realistic exclusion polygons are larger than the historic ones used originally in fiberassign.
2.  Realistic positioners may not be able to reach the very center of the patrol disc due to different arm lengths.

The new realism here is just something we have to deal with.  It remains to be seen if the assignment efficiency can be improved as part of the work on #176, #179, #183, and #184.

sbailey:
@Srheft has independently run this (thanks!) and verified that it doesn't fundamentally break things, though there are slightly fewer science assignments due to the increased exclusion zones causing more collisions to avoid (oh well, that is reality).  Merging now.

sbailey:
@tskisner please update doc/changes.rst in master with a summary of the changes in this PR.  Thanks.

PR #208, opened 2019-06-24
Sort by FIBER; set sciencemask et al by name or number even for SV

sbailey:
This PR adds a few features used by the Survey Validation Data Challenge:
* sort output by FIBER (even though downstream code shouldn't care about the fiber order, it currently does care)
* allow setting sciencemask etc. by name in addition to number, including for SV target catalogs
* There were also some updates to how it parses `fiberassign` options into `fba_run` and `fba_merge_results` options, which might make that slightly more maintainable by making it opt-in instead of opt-out.

### DIFF
--- a/bin/fiberassign
+++ b/bin/fiberassign
@@ -153,21 +153,15 @@ run_assign_full(assign_args)
 # ------
 # - fba_merge_results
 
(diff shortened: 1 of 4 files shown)


tskisner:
This looks good, thanks!

PR #219, opened 2019-10-01
Implement field rotation

tskisner:
DO NOT MERGE THIS PR.  I am still debugging an issue that may be due to an error in the upstream desimodel focalplane theta / phi ranges.

### DIFF
--- a/doc/nb/simple_plots.ipynb
+++ b/doc/nb/simple_plots.ipynb
@@ -108,9 +108,9 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "DESI 17.9",
+   "display_name": "Python 3",
    "language": "python",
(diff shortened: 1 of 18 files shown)


tskisner:
This work is now ready.  The unit test for the field rotation compares the petal assignments before and after a field rotation of 36 degrees (i.e. one petal location).  This test does not pass due to the previously implemented "tie breaking" of assignment order based on fiber ID rather than device location.  See section of code here:

https://github.com/desihub/fiberassign/blob/master/src/assign.cpp#L332-L339

This was implemented to agree with a previous version of the code, but we can throw that away at this point.  I verified that if I switch back to using location for tie-breaking then the rotated assignment matches.  I am not changing the algorithm in this PR, since it makes more sense to do it as part of #179, #183, and #184.

PR #225, opened 2019-10-09
added cmx_science bits for first light targets

Srheft:
@tskisner 
This is just to add first light targets to cmx_science mask.


tskisner:
Thanks @Srheft , this is fine.  The unit tests fail simply because the DESI svn server is unavailable to download the desimodel data.  We could switch travis tests to use the mirror, but I don't think that is needed for a couple day outage.  Merging now.

PR #226, opened 2019-10-10
use desimodel.focalplane.fieldrot

sbailey:
Uses the new `desimodel.focalplane.fieldrot.field_rotation_angle(mjd)` to get the field rotation to use for each tile if it isn't otherwise specified.  Requires desimodel master (to be tagged as 0.10.1).

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -7,8 +7,12 @@ fiberassign change log
 ------------------
 
 * fiberassign support for CMX targets + MAIN skies (PR `#224`_).
+* Added cmx_science bits for first light targets (PR `#225`_).
+* Use per-tile field rotations from desimodel.focalplane.fieldrot (PR `#226`_).
 
 .. _`#224`: https://github.com/desihub/fiberassign/pull/224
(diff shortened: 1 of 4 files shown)


sbailey:
Note: this doesn't (yet?) adjust the GFA_TARGETS selection based upon the field rotation.  The largest field rotation for DESI is ~0.4 degrees, while the GFAs subtend ~4.5 degrees.  We're planning on using a scale factor of 2.0 for which targets are included in the list to allow for initial pointing slop, so that should also be sufficient to absorb any non-zero field rotations too.  Updating the GFA target selection to support field rotation is probably best done on the desimodel side in `desimodel.focalplane.gfa`.

sbailey:
I also added `--fieldrot` and `--obsdate` command line options to `fba_run` (but not the outer `fiberassign` wrapper) and expanded the tests to include various forms of how the date can be specified.  I verified that running with different `--fieldrot` propagates to the output files FIELDROT keyword  and different DESIGN_X/Y values and that they have the right sign of rotation (I think...).

PR #230, opened 2019-10-14
qa-related-phase2

Srheft:
Please don't merge this until further notice; will add one more commit for fiber location match up that despite the updates gives unmatched error.

### DIFF
--- a/bin/qa-fiberassign
+++ b/bin/qa-fiberassign
@@ -159,8 +159,8 @@ for filename in args.tilefiles:
         errors.append('fiber:location map incorrect')
 
     ii = fa['TARGETID'] >= 0
-    dx = (fa['DESIGN_X'][ii] - fiberpos['X'][ii])
-    dy = (fa['DESIGN_Y'][ii] - fiberpos['Y'][ii])
+    dx = (fa['FIBERASSIGN_X'][ii] - fiberpos['X'][ii])
+    dy = (fa['FIBERASSIGN_X'][ii] - fiberpos['Y'][ii])
     r = np.sqrt(dx**2 + dy**2)
     if np.max(r) > 6:
         errors.append('Fibers assigned to more than 6mm from positioner center')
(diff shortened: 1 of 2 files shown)


sbailey:
These changes are all fine with me; feel free to merge this whenever you are ready.

Srheft:
@sbailey 
Just a note for documentation that. prior to this PR, qa-fiberassign would raise three errors per tile at all times, two of them turned out to be unjusted:

1-    **800 fibers with CMX_TARGET=0**   ---> this is not an error, SKY fibers do not have dedicated bits in different programs; 
2-   **fiber:location map incorrect** ---> turned out to be a fiber ID mismatch between ETC fibers with -1 and POS fibers with ID >5000 
3-   Fibers assigned to more than 6mm from positioner center  ---> this has stayed as an errror and needs further investigation. 

I applied the key change between the last  focalplane format and the current as 'X' no longer exists; X-> OFFSET_X 
and I am not sure the physical meaning of the offset_x is the same as X in the context of defining the radius here:
   
`dx = (fa['FIBERASSIGN_X'][ii] - fiberpos['OFFSET_X'][ii])`
`dy = (fa['FIBERASSIGN_Y'][ii] - fiberpos['OFFSET_Y'][ii])`
 ` r = np.sqrt(dx**2 + dy**2)`

This radous is consistently >6mm for all the tiles that I ever processd and just by the low of probability, it doesn't look right to always be true; if its correct then maybe the patrol radius has to be changed to that higher number if positioners keep reaching to targets outside their patro radius.  I can address this in a ticket  [and provide more stats as to what is the calculated 'r' value for the tiles that I processed with qa-fiberassign] if you see fit. 

Please merge if you apprive of my last commit.

sbailey:
@Srheft yes `OFFSET_X/Y` has the same meaning as `X/Y`; it's the name that the hardware team used in their upstream files.  I'll merge this PR now and Let's chase the radius >6mm separately.  I see that you are using `desimodel.io.load_focalplane` which should be correct.  It may have something to do with whether the comparisons are done with or without FIELDROT.

PR #233, opened 2019-10-23
Implement GFA and petal boundary exclusion zones

tskisner:
This branch is still undergoing testing.  It uses the polygons defined in the old collision ConfigObj format for the GFA and petal boundary keepout zones.  Initial tests with extra logging enabled and plots from the unit tests indicate things are working as implemented.

One question:  I had assumed that only the PHI arms could possibly extend into the GFA areas and off the petal boundaries.  Surely the THETA arm would not get close enough to the edge?

However, if I plot these polygons it looks like some device locations are on the petal polygon:
[fiberassign_001165.pdf](https://github.com/desihub/fiberassign/files/3764786/fiberassign_001165.pdf)

### DIFF
--- a/py/fiberassign/hardware.py
+++ b/py/fiberassign/hardware.py
@@ -97,7 +97,7 @@ def load_hardware(focalplane=None, rundate=None):
 
     for nm, shp in exclude.items():
         excl[nm] = dict()
-        for obj in ["theta", "phi", "gfa", "petal"]:
+        for obj in shp.keys():
             cr = list()
             for crc in shp[obj]["circles"]:
                 cr.append(Circle(crc[0], crc[1]))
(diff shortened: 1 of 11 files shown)


sbailey:
What is the source of your petal edge exclusion polygon?  It appears inconsistent with the pictures in [DESI-3453](https://desi.lbl.gov/DocDB/cgi-bin/private/RetrieveFile?docid=3453;filename=petal_map_flatxy.pdf;version=3) where the exclusion polygon does not interfere with the ETC positioners.

tskisner:
This came from the same file in the fp_settings svn repo where the theta and phi arm keepout zones are defined.  Perhaps we have just reached the end of the usefulness of those files and should switch to the DB info, but I had hoped to implement the general keepout mechanism first in fiberassign and then simply update the upstream focalplane file in desimodel as a separate step.

tskisner:
Ok, after more exploring, the numbers in `fp_settings/collision_settings/_collision_settings_DEFAULT.conf` agree with **version 1** of DESI-3453.  Since that time (01/09/2018), the DocDB document was updated but the files in svn were not.

Here is what I'm going to do:

1.  Copy this collision setting file into desimodel svn with updated exclusion polygons to match version 2 of DESI-3453.

2.  Use this file to generate a new focalplane file.  We could implement this under desimodel PR #127 or as a separate task.

3.  Proceed with this fiberassign work and get it merged

4.  Return to desimodel DB reading.

Thoughts?

tskisner:
I've implemented the new petal exclusion polygon to match the file `petal_map_flatxy.pdf` drawing in DESI-3453 and here is the same plot as above with this new shape:

[fiberassign_001165.pdf](https://github.com/desihub/fiberassign/files/3765584/fiberassign_001165.pdf)

I still have a concern- the central (THETA) exclusion polygon can extend into this petal exclusion zone (especially if we make that polygon larger in the future).  However, is this really physical?  The THETA arm should always be within the outer "clear rotation envelope" (E_0) defined in DESI-0899.  I see no point in spending time checking for collisions between the theta arm and the petal / GFA exclusion zones.  The code currently only checks the PHI arm for collisions on these.

sbailey:
Agreed with the update from DESI-3453, and agreed that we shouldn't need to check the theta arm unless we hear otherwise.

sbailey:
It looks like you commented out a bunch of tests, maybe for faster testing.  Let's uncomment those before merging.  Otherwise go ahead and merge when ready.

PR #240, opened 2019-12-04
Propagate run date / time and dependency versions to output headers

tskisner:
Change the header keys to FA_RUN and FA_PLAN for the run time and planned observing time.  Add dependency versions to output FITS header.


tskisner:
The travis failure was a sporadic network error.  I think these changes are non-controversial, but @sbailey should verify that he likes the keywords, etc.

sbailey:
Looks good, thanks.  Merging now.

PR #242, opened 2019-12-06
Doc example

rncahn:
This updates examples.rst to use 19.10 rather than 18.11.  Also, it provides a debug example analogous to the the first one in examples.rst, but instead built on the small example, so one can use it as input for the debug study, which looks at the process by which a particular fiber is finally assigned.


sbailey:
Thanks for the update.

PR #245, opened 2019-12-10
bits4cmxnsv

Srheft:
@tskisner this is just to add the recently added bits to the CMX and SV programs. Please take a look and merge whenever you see fit.  Thank you.


sbailey:
Looks like this got lost in the shuffle.  merging now.

PR #246, opened 2019-12-21
Travis updates

tskisner:
Update to latest xenial image.  Remove un-needed packages.  Explicitly specify every entry of the job matrix to just test the combinations we care about.  Add tests for gcc9 with python 3.6 and 3.7.  Pin our astropy and matplotlib versions to match desiconda.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,107 +1,123 @@
-# We set the language to python, since we are focused on a recent python
-# version and want to use the build matrix for testing different compiler
-# versions.
+# All jobs use the python language.  This brings in the travis python environments
+# which are available for several python versions.  The python version for each job is
+# specified in the build matrix below.
 language: python
 
-# We use APT/brew to get as many packages as possible.  Then we pip install the
-# rest of our dependencies.
+# Manually specify every entry in the job matrix.  For older compiler versions, we
+# only test with the oldest supported python version.  For the latest available
+# compiler version, we test with several python versions.
 
-matrix:
+jobs:
   include:
     # GCC 4.9 - this is the oldest version that has good C++11 support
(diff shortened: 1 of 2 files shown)


sbailey:
@weaverba137 please review this after @tskisner gets tests to pass again.  Thanks.

tskisner:
This PR is NOT ready for merge- I will squash all these debugging commits once it is working.

tskisner:
Ok @weaverba137 , I disabled the experimental builds for now (commented out) since I could not seem to get allowed_failures to match those entries and ignore them.

The main features of this travis file are that the different OS / compiler / python cases are listed manually.  This is somewhat tedious, but it does mean that we can test only older python versions with older compilers (for example).  Or we could manually add an entry for using conda on linux without also looping over every gcc version.

tskisner:
You raise a good point:  currently the test suite is being run on a separately-built version of the package, which is useful if the tests are not part of the installed package (for example if they were in a "tests" directory that was outside of the package directory.  In the case of desi packages, the test suite for each package (and its data) is actually installed along with the package.  To run the installed version of the test suite, I think we just need to do `python -m <blah>` where the module to load is whatever runs the test suite.  I'll try to experiment with that a bit locally.

tskisner:
Ok @weaverba137 , after installation, this now does:
```
python -c 'import fiberassign.test; fiberassign.test.runtests()'
```
So that we are sure to be running the installed tests.  If you approve, then let's merge this.

weaverba137:
That's a good solution.  Merge when ready.  To save some time on my side, I don't think we need to incorporate this into the 19.12 release, so I won't necessarily cut a new tag.

PR #247, opened 2020-01-08
Use curved focal surface internally for assignment

tskisner:
[WIP]:  Do not merge yet.

The geometric inputs to fiberassign are a mix of CS5 and curved focal surface (positioner-local) coordinates.  This work:

1. Converts everything into curved focal surface coordinates on read.

2.  Target RA/DEC is projected to curved focal surface coordinates when doing assignment (including collision checks)

3.  On output, assigned target RA / DEC is converted to CS5 for writing, since that is the agreed upon coordinate system for the output files.

This work also fixes the **ancient** radial platescale calculation used in fiberassign that is just simply wrong.  Instead we get the platescale data from desimodel and use python routines to interpolate that to high resolution.  This is then passed into compiled code where a simple linear interpolation is fast and provides accuracy to 1e-6 mm.

This PR should not be merged until https://github.com/desihub/desimodel/pull/130 is merged.


tskisner:
I believe this PR is ready @sbailey .

sbailey:
Apologies for the delay on this.  The sanity check that I wanted to do (but haven't found the time for) was fiberassign several individual tiles with both master and curvedfp branch, and then plot the focal plane location of every target that was assigned differently.  I believe those should only occur near the edges of the focal plane, because that is where the change in platescale matters most, and where the curved surface vs. CS5 x,y plane matter most.  Additionally, plot their distance to the nearest positioner: changes should only occur at the outer reaches of a positioner where the differences move something from possible to impossible or vice-a-versa.  If there are any exceptions to those two expectations, chase a few cases in detail to make sure they are understood (e.g. some subtlety about collisions, where a small movement of one positioner creates a collision that wasn't previously there).

Additionally, for the assignments that are the same, the output CS5 FIBERASSIGN_X,Y should only differ by a smoothly varying radial-only term due to the change in platescale modeling.

tskisner:
Radial Coordinate Changes
===================================

Just to re-emphasize, the largest change in this branch compared to master is the use of a realistic platescale from desimodel.  The current master branch uses a hard-coded and incorrect platescale that was implemented in the legacy codebase years ago.  A second-order change is doing the positioner assignment and collision checks in the curved focal surface rather than CS5.

Here is a plot comparing the combined effects on the radial difference between master and this branch.  This just samples some RA/DEC points and converts them into X/Y in the coordinates used for assignment (CS5 in master and curved in this branch):

compare_platescale_diff

Assignment Changes
=======================

As you can see, there are slight differences even at short radial distances.  The next question is how does this impact the assignment.  To get a clean testing dataset I used a realization of the purely synthetic data used by the unit tests.  I generated that and copied it to a new location and then ran both the current master and this branch on it.

Changes to Available Targets
----------------------------------

For each positioner, fiberassign initially computes the targets available to that positioner based simply on geometry.  As you can imagine, if the locations of targets change due to the coordinate conversions, then targets on the edge of a positioners reach may go from being available to not available (and vice versa).  Usually this just moves an available target from one positioner to another.  However, there are cases where a target becomes unavailable to **any** fiber, since there are dead zones in between positioners depending on arm lengths and the buffer region around each patrol radius.  Here is a plot of the targets that are considered available (to **any** positioner) in one case and not the other:

compare_targets_extra

Before you get too concerned, this synthetic data set has a high target density, so it is not surprising that a small shift changes which targets are available across the entire tile.  However, I plotted some of these cases, and you can see that the targets which changed state from available to not all lie on the edge of positioner reach or sometimes in the small area in between 3 positioners that is not reachable:

target_shifts_curvedfp
target_shifts_master


Changes to Assignment
---------------------------

Obviously if you change what targets are available, then you would expect changes to the assignment.  Here I plot just the positioners that changed assignments between master and this branch:

compare-assign

As you can see, many positioners changed due to an assigned target shifting slightly out of the patrol radius.  However, because the available targets are changing slightly across the tile (due to the coordinate changes above), there can potentially be other effects.

Conclusion
==================

I think all of these impacts and changes are reasonable considering the level of change in the RA/DEC --> X/Y conversion implied by using the modern platescale and using the curved focal surface rather than CS5.

It would be great to merge this so that we can move on with other algorithm improvements.

sbailey:
I'm still surprised that the density of differences is approximately the same at small radius compared to large radius, given that both the platescale changes and the curved focal surface fiber reach are bigger effects at large radius than at small.  I've poked around at the transforms some, and don't find anything wrong though.  Merging now so that this doesn't continue to hold up other work.

PR #248, opened 2020-01-14
Fixes creation of sv1_sciencemask

apcooper:
The code to create `sv1_sciencemask` in `targets.py` was looking up `scnd_mask` bits in `desi_mask`. 

Unless I'm missing something, this doesn't work. This PR just removes those lines. `SCND_ANY` is being `OR-`ed, so all secondary targets are counted as science anyway.

If some secondary targets shouldn't be counted as science, an alternative would be to use a custom bitmask in place of `SCND_ANY`, based on just the whitelisted secondary bits.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -6,7 +6,9 @@ fiberassign change log
 1.3.1 (unreleased)
 ------------------
 
-* No changes yet.
+* Fix issue with creation of SV1 science target mask (PR `#248`_).
+
+.. _`#248`: https://github.com/desihub/fiberassign/pull/248
 
 1.3.0 (2019-12-20)
 ------------------
--- a/py/fiberassign/targets.py
+++ b/py/fiberassign/targets.py
@@ -21,6 +21,7 @@ from desitarget.targetmask import desi_mask
 from desitarget.cmx.cmx_targetmask import cmx_mask
 
 from desitarget.sv1.sv1_targetmask import desi_mask as sv1_mask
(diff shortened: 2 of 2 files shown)


sbailey:
@Srheft and @geordie666 could you take a look at this and comment?  This is undoing the bits that @Srheft added in PR #245 .

Srheft:
@sbailey : seems like @apcooper made this PR after his latest assignment test of MWS tiles that would use the secondary target bits which did not come up in assignment of stellar tiles in CMX. so assigning new bitmask were not properly tested. if this change has made the fiberassign to properly assign the MWS secondary targets, then this is the way to go.

apcooper:
@Srheft: thanks for the clarification, yes, that's exactly how this came about -- I was testing the MWS tiles with the latest fiberassign and the change that I've un-done with this PR was causing problems.

Srheft:
OK; I'll be doing the global SV assignments for all programs in the near future and if any issue came up given the changes in this PR, then we can make the appropriate changes again.

geordie666:
I also agree that this PR is the way to go. It's always possible that we'll break instances downstream. Or we'll discover use-cases where we want non-science-secondaries. But, in any case, I think we should merge this and wait to see if those cases arise in the future.

sbailey:
Apologies for letting this dangle; in the meantime an equivalent fix (I think) was applied to PR #250, so I'm closing this without merging.

If the original problem persists, please open a new ticket describing the problem.  Thanks.

PR #250, opened 2020-02-11
Fix for #249: distinguish betwen scnd and desi mask.

michaelJwilson:
Fix for #249:   distinguish betwen scnd and desi mask.  Runs to completion.  I don't think I broke the logic.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -6,7 +6,9 @@ fiberassign change log
 1.3.1 (unreleased)
 ------------------
 
-* No changes yet.
+* Change targets to correctly look up desi and secondary mask (PR `#250`_).
+
+.. _`#250`: https://github.com/desihub/fiberassign/pull/235
 
 1.3.0 (2019-12-20)
 ------------------
--- a/py/fiberassign/targets.py
+++ b/py/fiberassign/targets.py
@@ -21,6 +21,7 @@ from desitarget.targetmask import desi_mask
 from desitarget.cmx.cmx_targetmask import cmx_mask
 
 from desitarget.sv1.sv1_targetmask import desi_mask as sv1_mask
+from desitarget.sv1.sv1_targetmask import scnd_mask as sv1_scnd_mask
 
 from desitarget.targets import main_cmx_or_sv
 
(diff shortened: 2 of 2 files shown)


forero:
@Srheft Does this change look good to you? Is it compatible with your current workflow for commissioning tiles?

michaelJwilson:
Added PR number to changes.rst

sbailey:
This PR is related to the dangling PR #248.  @geordie666 could you compare them and coordinate the solution?

@michaelJwilson `py/fiberassign/_internal.cpython-36m-x86_64-linux-gnu.so` was accidentally added.  Please remove it from your branch.

@geordie666 could you revive the BFG repo cleaner and expunge that file from the history?  You are our cleanup expert for that. :)  Thanks.  https://help.github.com/en/github/authenticating-to-github/removing-sensitive-data-from-a-repository

michaelJwilson:
Removed _internal.so.

Srheft:
@forero I don't use any SV type mask and data at the moment; I'll report the potential issues as soon as they come up.

geordie666:
@sbailey: I'm not close enough to fiberassign to know whether this PR will be the correct choice for the real data for SV. This PR seems correct, though, and would supersede PR #248. So, I'd say we should merge this (if @forero finds it works with the mocks), fix any issues as they arise later, and delete PR #248 without merging it.

I'll wheel out the BFG later today and remove @michaelJwilson's `.so` file from history.

geordie666:
I believe I have now removed @michaelJwilson's ".so" file from the general git history. Anyone attached to this thread should remove their `fiberassign` repository completely and clone a new version:
```
\rm -rf desitarget
git clone https://github.com/desihub/desitarget
```
To prevent reintroducing this file.

It would be great if the first person who does that could also confirm for me that the file `py/fiberassign/_internal.cpython-36m-x86_64-linux-gnu.so` is no longer present in their newly cloned repository.

sbailey:
I confirm that py/fiberassign/_internal.cpython-36m-x86_64-linux-gnu.so appears to be successfully removed from the git history.

I will proceed with merging this PR and closing #248 without merging.

PR #253, opened 2020-02-27
Extended QA

tskisner:
This PR adds some additional functions for looking at the details of target types that have been assigned.  This work is not exposed to any commandline scripts, but is useful for custom work.

Fixing the fitsio travis tests is a work in progress.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,6 @@ jobs:
             - g++-4.9
       env:
         - COMPILERS="export CC=$(which gcc-4.9) && export CXX=$(which g++-4.9)"
-        - PIP_BINARY="pyyaml scipy matplotlib==2.1.2 numpy astropy==2 sphinx"
     # GCC 5 - default version for ubuntu 16.04 LTS
     - os: linux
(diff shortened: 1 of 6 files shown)


tskisner:
Some other notes about this PR:

- The NUMOBS_MORE value for each target is no longer written to the output files.  This was incorrect behavior, since the **input** NUMOBS_MORE value from the target catalogs is read by fiberassign and internally decremented on each assignment.  After running fiberassign, the internal NUMOBS_MORE value reflects the final status after assigning all tiles.  Writing this same value to all output tiles does not reflect the changing NUMOBS_MORE across the tiles.  The new NUMOBS_MORE value for each target should be calculated by the code that is creating an updated MTL file for the next run of fiberassign.

- When loading output assignment files, we now prefer the merged `FIBERASSIGN` HDU over the raw `FASSIGN` HDU if they both exist.  The reason for this is that the merged HDU has additional information like the starting requested number of observations.  This extra information is useful for the new target QA.

- Fix a typo in the little-used by-tile assignment function.

PR #254, opened 2020-02-29
 making lite merge default and full merge an option

Srheft:
@tskisner, @sbailey :

## Trivia:

As of few days ago, ICS accepts commissioning tiles that are much lighter in size than the origonal merged  fiberassign-0XXXXX.fits file (up to 90% lighter!). This was achieved in a postproduction step by redacting **three HDUs** out of 8 HDUs hat were traditionally written into the merged output; the redacted HDUs are: **TARGETS**, **FTARGETS**, **FASSIGN**. 

## This PR: 
Per @sbailey request, full merge is now a running option to be activated by adding **--full_merge** to fba_merge_results; e.g., 

`fba_merge_results --targets ./MTL_tile66000.fits --sky skies_tile66000.fits suppskies_tile66000.fits --dir ${odir} --full_merge`

not adding this --full_merge option would activate the default merging mode in which there are only 5 HDUs in the merged output:
`>>> a.info()`
`Filename: ./FAout/fiberassign-066000.fits`
`No.    Name      Ver    Type      Cards   Dimensions   Format`
`  0  PRIMARY       1 PrimaryHDU      35   ()      `
`  1  FIBERASSIGN    1 BinTableHDU    287   5000R x 125C   [K, I, J, J, J, J, D, D, E, E, E, E, E, E, K, B, 3A, E, E, I, J, D, J, J, I, J, 8A, J, 4A, E, E, 5E, E, E, E, E, E, E, E, E, E, E, E, E, E, E, E, E, E, E, E, I, I, I, E, E, E, E, E, E, E, E, E, E, E, E, E, E, E, E, E, E, I, I, I, E, E, E, E, E, E, E, E, E, E, E, E, E, E, E, E, E, E, E, E, B, B, I, K, 2A, E, E, E, E, E, E, E, E, L, E, L, E, E, 1A, K, J, J, K, E, E, E, E, K, K, K]   `
  `2  SKY_MONITOR    1 BinTableHDU     85   20R x 24C   [J, J, I, K, J, J, K, B, D, D, E, E, 8A, J, I, J, J, D, E, E, E, E, E, E]   `
  `3  GFA_TARGETS    1 BinTableHDU    113   1937R x 38C   [J, K, J, J, D, D, E, E, 4A, E, E, E, E, E, E, K, 2A, E, E, E, E, E, E, E, E, E, E, E, E, E, E, K, E, K, I, I, I, I]   `
  `4  POTENTIAL_ASSIGNMENTS    1 BinTableHDU     43   422863R x 3C   [K, J, J]   `
  `5  FAVAIL        1 BinTableHDU     43   422863R x 3C   [J, J, K]   `

Originally I had a branch where --lite_merge was an option and default was the full merge; we can go back to that, if more desired.

### DIFF
--- a/py/fiberassign/assign.py
+++ b/py/fiberassign/assign.py
@@ -431,6 +431,7 @@ def write_assignment_fits_tile(asgn, fulltarget, overwrite, params):
         # tm.report("  copy targets data tile {}".format(tile_id))
         # tm.clear()
         # tm.start()
+        
 
         fd.write(fdata, header=header, extname="FTARGETS")
         del fdata
@@ -710,6 +711,7 @@ def read_assignment_fits_tile(params):
         full_targets_columns = [(x, y) for x, y in
                                 results_targets_columns.items()]
         full_names = [x for x, y in results_targets_columns.items()]
+
         fbtargets = fd["TARGETS"].read()
         nrawtarget = fd["TARGETS"].get_nrows()
 
@@ -894,7 +896,8 @@ merged_potential_columns = OrderedDict([
 ])
 
 
-def merge_results_tile(out_dtype, copy_fba, params):
+def merge_results_tile(out_dtype, copy_fba, full_merge, params):
+     
     """Merge results for one tile.
 
     This uses target catalog data which has been pre-staged to shared memory.
@@ -906,6 +909,8 @@ def merge_results_tile(out_dtype, copy_fba, params):
             HDU.  This is the union of columns chosen from the input catalogs.
         copy_fba (bool):  If True, copy the original raw fiberassign HDUs onto
             the end of the output file.
+        full_merge (bool): if True, copy 4th, 6th and 7th HDUs onto the end of
+            the output file.
         params (tuple):  The tile ID and input / output files.  Set by
             multiprocessing call.
     Returns:
@@ -922,7 +927,7 @@ def merge_results_tile(out_dtype, copy_fba, params):
 
     inhead, fiber_data, targets_data, avail_data, gfa_targets = \
         read_assignment_fits_tile((tile_id, infile))
-
+   
     # tm.stop()
     # tm.report("  read input data {}".format(tile_id))
     # tm.clear()
@@ -1173,7 +1178,10 @@ def merge_results_tile(out_dtype, copy_fba, params):
         else:
             newnames.append(nm)
     tile_targets.dtype.names = newnames
-    fd.write(tile_targets, header=inhead, extname="TARGETS")
+
+    if full_merge:
+
+        fd.write(tile_targets, header=inhead, extname="TARGETS")
     del tile_targets
 
     # Write the "POTENTIAL_ASSIGNMENTS" HDU
@@ -1189,10 +1197,14 @@ def merge_results_tile(out_dtype, copy_fba, params):
     fd.write(potential, header=inhead, extname="POTENTIAL_ASSIGNMENTS")
 
     # Now copy the original HDUs
-    if copy_fba:
+    if copy_fba and full_merge:
         fd.write(fiber_data, header=inhead, extname="FASSIGN")
         fd.write(targets_data, header=inhead, extname="FTARGETS")
         fd.write(avail_data, header=inhead, extname="FAVAIL")
+    
+    if copy_fba and not full_merge:
+        fd.write(avail_data, header=inhead, extname="FAVAIL")
+
 
     # Close the file
     fd.close()
@@ -1214,7 +1226,8 @@ def merge_results_tile(out_dtype, copy_fba, params):
 def merge_results(targetfiles, skyfiles, tiles, result_dir=".",
                   result_prefix="fba-", result_split_dir=False,
                   out_dir=None, out_prefix="fiberassign-", out_split_dir=False,
-                  columns=None, copy_fba=True):
+                  columns=None, copy_fba=True, full_merge=False):
+    
     """Merge target files and assignment output.
 
     Full target data is stored in shared memory and then multiple processes
@@ -1236,7 +1249,7 @@ def merge_results(targetfiles, skyfiles, tiles, result_dir=".",
             target files (default is all).
         copy_fba (bool):  If True, propagate the original raw HDUs at the end
             of each merged output file.
-
+        full_merge (bool): If True, write TARGETS, FASSIGN and FTARGETS HDUs.
     Returns:
         None.
 
@@ -1349,7 +1362,7 @@ def merge_results(targetfiles, skyfiles, tiles, result_dir=".",
     # For each tile, find the target IDs used.  Construct the output recarray
     # and copy data into place.
 
-    merge_tile = partial(merge_results_tile, out_dtype, copy_fba)
+    merge_tile = partial(merge_results_tile, out_dtype, copy_fba, full_merge)
 
     if out_dir is None:
         out_dir = result_dir
--- a/py/fiberassign/scripts/merge.py
+++ b/py/fiberassign/scripts/merge.py
@@ -83,6 +83,12 @@ def parse_merge(optlist=None):
                         help="Do not copy the raw fiberassign HDUs to the "
                              "merged output.")
 
+    parser.add_argument("--full_merge", required=False, default=False,
+                        action="store_true",
+                        help="copy the TARGETS, FASSIGN and FTARGETS "
+                             "HDUs to the merged output.")
+
+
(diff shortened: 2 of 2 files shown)


tskisner:
I will try to look at this tomorrow.  Some of the reading code in fiberassign (for reading back in the output files) assumes that the data is either "raw" or "merged".  It detects this by the presence of the `FIBERASSIGN` HDU and if that exists it also assumes that the `TARGETS` HDU exists.  It sounds like "lite-merged" files will break this.

Srheft:
Thanks @tskisner ; My tests have been just to start from scratch and see if --full_merge or lack there of in the **fba_merge_results** command would result in desired number of HDUs in the outputs and it did. I should do the travis test locally to see where those reading in errors are coming into play.

tskisner:
Hello @Srheft and @sbailey , I just wanted to remind you of 2 options which impact the file size of the output:

1.  `fba_run --write_all_targets`:  if you enable this, then **ALL** available targets are written to the `FTARGETS` HDU.  Without this option, only 5000 targets are written to that HDU.

2.  `fba_merge_results --skip_raw`:  If enabled, then only the merged HDUs are in the output.

Example:  Imagine you had 76,543 available targets on a tile.  The 4 possible permutations of the above options would give you:

-  `fba_run ...` and `fba_merge_results --skip_raw ...`:

    FIBERASSIGN (5000 rows of ~120 columns)
    SKY_MONITOR
    GFA_TARGETS
    TARGETS (5000 rows of ~120 columns)
    POTENTIAL_ASSIGNMENTS

-  `fba_run ...` and `fba_merge_results ...`:

    FIBERASSIGN (5000 rows of ~120 columns)
    SKY_MONITOR
    GFA_TARGETS
    TARGETS (5000 rows of ~120 columns)
    POTENTIAL_ASSIGNMENTS
    FASSIGN (5000 rows of 14 columns)
    FTARGETS (5000 rows of 8 columns)
    FAVAIL

-  `fba_run --write_all_targets ...` and `fba_merge_results --skip_raw ...`:

    FIBERASSIGN (5000 rows of ~120 columns)
    SKY_MONITOR
    GFA_TARGETS
    TARGETS (76543 rows of ~120 columns)
    POTENTIAL_ASSIGNMENTS

-  `fba_run --write_all_targets ...` and `fba_merge_results ...`:

    FIBERASSIGN (5000 rows of ~120 columns)
    SKY_MONITOR
    GFA_TARGETS
    TARGETS (76543 rows of ~120 columns)
    POTENTIAL_ASSIGNMENTS
    FASSIGN (5000 rows of 14 columns)
    FTARGETS (76543 rows of 8 columns)
    FAVAIL

If you are concerned about file sizes, I think just using the first case above would get you most of the way.  Are you currently using the `--write_all_targets` option?  If so, then that will definitely create large files for high target density.  There are other things we could do that would be even more useful:

1.  Only merge the target catalog data to the `TARGETS` HDU rather than duplicating it to the `FIBERASSIGN` HDU.

2.  Only store the "raw" assignment files for online use and do the merging later / elsewhere.

Srheft:
Thanks @tskisner;   yeah I have been using `--write_all_targets` but not using that option still creates larger files than the 7-9MB that we get with 5HDU output. 

My impression is that this PR was meant to make the "5 HDU merge or skip_raw merge"  the default for `fba_merge_results` so we wouldn't need to activate it with `--skip_raw`[?] OR, we  could just have things as they are and use the `--skip_raw`; I'm OK with either of these ways to go about reducing the output size; Stephen?

tskisner:
Some other thoughts on this issue:

- If we are going to produce files without the full set of available targets and their properties, then we should provide a tool to reconstruct files later with all available targets.  This is not something that is is easy to do efficiently- it requires loading a large volume of target data into memory to extract the target properties for a particular file.  There are lots of hoops we jump through (using python shared memory, etc) to do this during the merging.

- The downstream analysis is going to need the available targets and at least a subset of their properties.

- What if we made the "raw" format align with only the information needed by online systems, and then made the "merged" format contain "everything" (perhaps with some de-duplication).

These points are mainly just to drive discussion, unfortunately I can't play with file formats for a couple weeks at least...

Srheft:
> If we are going to produce files without the full set of available targets and their properties, then we should provide a tool to reconstruct files later with all available targets. This is not something that is is easy to do efficiently- it requires loading a large volume of target data into memory to extract the target properties for a particular file. There are lots of hoops we jump through (using python shared memory, etc) to do this during the merging.

This is a very valid point; Ashley Ross touched on this last December in Columbus meeting as well. And at that point we were only talking about little data mining modules for fiberassign outputs; we had not thought about the volume of the loaded data to do this; that's a whole other point to consider;

I did advocate for saving the "heavy" fiberassign files for ALL the tiles we have observed so far just to help with future in-depth analysis and inquiries that are surely to come. The impression I got from higher ups was to "let's not worry about that right now".

And to prove their point, I'd say that so far, the science groups have been surprisingly fine with the HUDs that are available to them; No one really asked the fiberassign side to provide them with photometric info or any out of reach data from catalogs. This gave me pause as to maybe we are "good to go" with the 5HDUs that we currently available in the "lite merge".  

> What if we made the "raw" format align with only the information needed by online systems, and then made the "merged" format contain "everything" (perhaps with some de-duplication).

 We could certainly take a meaningful step toward the "de-duplication process". Naive question: How feasible it is to do something about making the the `fba_run` output to become ICS-useful?  had it had the SKY_MONITOR and GFA_TARGET HDUs, then we wouldn't even run merge_result step for kittpeak storage [we would just store the merged files on NERSC for further analysis];

ashleyjross:
From the target file, all that *really* needs to be in each HDU is the TARGETID. It would be convenient to have more than that, but we should be able to match back to whatever information given the IDs.

This principal already seems to be in place for the 'FAVAIL' HDU. 

As a related issue (apologies if this is too much of a digression), I think we definitely want to be able to reproduce the FAVAIL and FTARGETS HDUs (probably just using science targets is fine). I'm using these to produce random catalogs right now (and everything seems super efficient). My understanding that these should be constant if the mtl file is constant and the hardware state is constant.

forero:
Comparing the results from `fiberassign`in `master` and `lite_merge` on DR8 data I only get a reduction of 40% in size for the final `fiberassign-.fits` files.  

I find it inconvenient loosing  `TARGETS`, `FTARGETS`, `FASSIGN` and having to reconstruct information from the `TARGETID`. 

The reconstruction would require 1) reading large target files (vs. small files of 9Mb from fiberassign) and 2) figuring out what exact target file was used to generate that assignment (something that doesn't seem to be stored anywhere).

Srheft:
Thanks for great discussion everyone;

Size Reduction is surely not the same for every assignment file;
Particularly if the input MTL has all the columns of the target catalog and the assignment is run with --write_all_targets flag, a reduction size from a 90 MB to 9 MB is observed; so definitely >40%  and this is observed for hundreds of tiles in cmx survey. 

I do also emphasize that the great use for TARGETS and FTARGETS HDUs are not in dispute here. They can exist as default or get added by an activating flag [and that's the point of this PR on whether have it as default or have it as an option]. Removing the numerous duplicate columns in these HDUs definitely helps with the disk space efficiency. This PR was requested and meant to be mainly for Stephen to catch up with what David and I mean by 'lite' version of a tile in emails.

forero:
It also looks like the `fiberassign` script is missing an option to force the full merge option.

Srheft:
> It also looks like the fiberassign script is missing an option to force the full merge option.

Sorry not sure what you mean by this; as to whether you are suggesting an addition to this PR, or a gheneral point about the `fiberassign` script. One outcome of this whole thread of discussion could be that we don't merge this to begin with and so full merge remains the default and always enforced.

tskisner:
@forero you are correct.  I developed the separate `fba_run` and `fba_merge_results` command line tools because those 2 steps require very different parallelism.  @sbailey felt strongly that there should be "one script" to run, so the "fiberassign" commandline script became a wrapper around the two underlying tools.  However, not all the options are exposed to that wrapper.  I never use the `fiberassign` script personally, because that script forces you to either have threads (which slows down the merging) or use only multiprocessing (which slows down some steps of the assignment).

I agree though that we should propagate all of the underlying options to the `fiberassign` wrapper.  Or else remove the wrapper altogether.

Srheft:
@tskisner ah now I see what Jaime point meant to indiucate and the explanation helped. Thanks.

sbailey:
closing as deprecated by other fiberassign format changes PRs.

PR #255, opened 2020-03-05
Don't propagate 2D target columns

### LINKED ISSUE #237: Don't propagate 2D target columns into FIBERASSIGN HDU
ICS uses Pandas dataframes to parse fiberassign tables, but Pandas doesn't support 2D columns.  When merging the target catalog into the FIBERASSIGN table, exclude 2D columns.

Example problem at KPNO:
```python
import fitsio
t = fitsio.read('/data/tiles/ALL_tiles/v1/fiberassign-059958.fits', 'FIBERASSIGN')
for name in t.dtype.names:
    if len(t[name].shape) > 1:
        print(name, t[name].shape)

...
DCHISQ (5000, 5)
```

It is not yet known whether we have to exclude these from both the FIBERASSIGN and TARGETS HDU, or whether it is sufficient to exclude them from just the FIBERASSIGN HDU (hopefully).

forero:
Fixes #237

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -7,8 +7,10 @@ fiberassign change log
 ------------------
 
 * Change targets to correctly look up desi and secondary mask (PR `#250`_).
-
+* Avoid propagation of 2D target columns into FIBERASSIGN and TARGETS HDU (PR `#255`_)
+  
 .. _`#250`: https://github.com/desihub/fiberassign/pull/235
+.. _`#255`: https://github.com/desihub/fiberassign/pull/255
 
 1.3.0 (2019-12-20)
 ------------------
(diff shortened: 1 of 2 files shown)


forero:
@tskisner Thanks!

@Srheft what do think about these changes? can you confirm that this won't break anything for the SV work you are doing?

Srheft:
yeah I have been doing this change at the MTL level and haven't gotten a chance to test this branch this past busy week; I'd rather try this before give you the green light;  if it's rushed to merge, go ahead and I make a ticket if I discover a problem. Thanks

forero:
I will go for the second option. I'll merge now and will let you open a ticket if you discover a problem. Thanks!

PR #258, opened 2020-03-16
Change assignment algorithm to be based on target order, not fibers

tskisner:
This is a substantial algorithm change which addresses #179, #183, and #184:

  * Targets now have a concept of "total priority".  This is a
    combination of priority, subpriority, and remaining obs.
    Currently this is constructed the same for all targets using
    a "breadth first" weighting scheme.  However, this could be
    easily changed to a per-target behavior based on bits in
    the DESI_TARGET field for each target.

  * The Assignment::assign_unused() method loops over targets in
    order of total priority and assigns them to available fibers.
    This is different from the previous code which looped over
    fibers and assigned them to targets.

  * The Assignment::assign_force() method loops over assigned
    science targets in inverse total priority order and bumps
    them.  The sky or standards assigned (if multiple are
    available for the fiber) are tried in total priority order.

  * The Assignment::redistribute_science() method loops over
    assigned science targets in inverse total priority order
    and attempts to move them to later tiles which both have
    an available fiber and are more sparsely assigned.

  * All old vestiges of doing things in fiber order are gone.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -6,7 +6,9 @@ fiberassign change log
 1.3.2 (unreleased)
 ------------------
 
-* No changes yet
+* Change assignment algorithm to be based on target order (PR `#258`_).
+
+.. _`#258`: https://github.com/desihub/fiberassign/pull/258
 
 1.3.1 (2020-03-13)
 ------------------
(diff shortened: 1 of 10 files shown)


sbailey:
This is a nice conceptual simplification of the code.

Confirming / clarifying:
  * `Target::total_priority()` is the function that uniquely determines each target's total priority, and nothing else uses `priority` or `subpriority` anymore
  * obsremain is only used in `total_priority`, and in one other check to skip targets with `obsremain<=0`
  * when you say "loops over targets in order of total priority" this means from highest total priority to lowest total priority (not in numeric order of priority)
  * when you say "inverse total priority order" that means from lowest total priority to highest total priority

Right?

tskisner:
Correct on all your points.

tskisner:
Also, since the obs_remain value is only used inside the `total_priority()` function, we can customize how that folds in to this total priority on a target-by-target basis using bits in DESI_TARGET to control behavior.  So depth-first vs breadth-first can be different for the different target classes (for example).

tskisner:
NOTE:  I just rebased this branch against master in order to bring in the fix from #259.  If you were already testing this locally, then you need to delete your local branch and check it out again.  For example:
```
git checkout master
git fetch
git rebase origin/master
git branch -D target_order
git checkout -b target_order -t origin/target_order
```

PR #259, opened 2020-03-16
Fix radial interpolation to work with recent desimodel update

tskisner:
Change the bounds of the radial interpolation to be based on the original data rather than a hardcoded value.


sbailey:
For the record, this was motivated by the update to the as-built platescale, which cuts off at 419.988569980 instead of 420.0, which caused problems for some versions of scipy.  The change looks fine, merging now.

PR #264, opened 2020-04-25
Check positioner motion when constructing TargetsAvailable.

### LINKED ISSUE #263: When computing `TargetsAvailable` check for actual positioning ability
This is the result of offline discussion with @michaelJwilson .  The `TargetsAvailable` class has been used internally in fiberassign as a way of reducing the number of targets to check for each positioner when assigning a positioner.  Currently this just gets all targets in a circle around each location based on the sum of the arm lengths minus some buffer.  Only when the positioner is actually assigned do we check that the positioner can physically move to reach the target.

This is a problem for downstream analyses that interpret the "available" targets written to the output as being physically reachable (not just "worthy of checking").

The solution is to check that the positioner can reach each target during the construction of a `TargetsAvailable` class.

tskisner:
The TargetsAvailable class stores the initial list of targets
to consider for each positioner.  This is before any checks for
collisions or the remaining obs for each target.  Previously,
the check for whether these targets are physically reachable was
done at the same time as collision checks during assignment.
This work moves that "physically reachable" check to the earlier
step when computing TargetsAvailable.  So now the available targets
that are written to the output include this constraint.

Closes #263.

### DIFF
--- a/src/assign.cpp
+++ b/src/assign.cpp
@@ -1093,18 +1093,9 @@ bool fba::Assignment::ok_to_assign (fba::Hardware const * hw, int32_t tile,
         return false;
     }
 
-    // Can this positioner actually reach the target location?
-    fbg::dpair tpos = target_xy.at(target);
-    if (hw->position_xy_bad(loc, tpos)) {
-        if (extra_log) {
-            logmsg.str("");
-            logmsg << "ok_to_assign: tile " << tile << ", loc "
-                << loc << ", target " << target << " cannot reach X/Y = "
-                << tpos.first << ", " << tpos.second;
-            logger.debug_tfg(tile, loc, target, logmsg.str().c_str());
-        }
-        return false;
-    }
+    // NOTE:  When building the TargetsAvailable instance, we already check that the
+    // positioner can physically reach every available target.  No need to check that
+    // here.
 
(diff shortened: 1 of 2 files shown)


michaelJwilson:
It's clear what you're trying to do so it largely looks good to me.  Maybe Stephen can still verify. 
Are there any potential 'gotchas' in the definition of bad_xy here?  Is it set early with all required hardware constraints etc?  Timeline wise, it'd be great to get this sorted by e.g. tomorrow or Wed.  
Although I can work off the PR if necessary.

sbailey:
A clarification note since it wasn't clear to me: previously FAVAIL / POTENTIAL_ASSIGNMENTS HDUs included targets that were close enough to consider even if upon further inspection it turned out that they weren't physically reachable.  This PR updates that so that only physically reachable targets are included (which are the ones we care about for the HDU).

michaelJwilson:
That's correct. The decision of physically reachable has now been moved before the FAVAIL HDU is defined and therefore included in that decision.  Before, it was a decision based solely on the arm lengths.  So, e.g. for the reduced reach in SV data (that was implemented by theta, phi restrictions) objects that were in FAVAIL were very much not reachable.

PR #267, opened 2020-05-07
Fix positioner angle range checks.

### LINKED ISSUE #266: Incorrect effective positioner angles
This issue is to track a critical problem discovered as part of recent effort to understand fiberassign efficiencies.  Thanks to Anand and @Cyeche for help in illustrating the issue.  

It looks like the theta / phi angle ranges of each positioner are not being referenced or rotated correctly to the petal they are in.  This results in an effectively reduced range of the positioners as the petals go from location 3 (at zero rotation from the coordinate system) around to location 2.

This can be illustrated with a small test on one tile that samples a high uniform density of targets (e.g. 50000 / sq deg), computes the targets available to all positioners, and then plots the targets **not** available to **any** positioner.  See attached plots of the positioner locations and the "unreachable" targets.  Petal location 3 has the expected spots where targets might be unreachable, while petal location 2 has many more in a suspicious angular slice of the patrol region.

This *might* be a desimodel issue rather than fiberassign, but I will leave this issue here until it is clear the source of the problem.

coverage_50000
target_reachability_zoom

tskisner:
This work:

  - Handles the case of arbitrary positioner theta / phi
    offsets and min / max.

  - Adds a unit test which attempts to move every positioner
    to a circle of spots halfway from the center to the patrol
    radius.  It also verifies intended failure in the case of
    restricted phi range.

Attached are plots of "*unreachable targets*" computed by taking uniform random RA / DEC points with a density of 50K per sq. deg. and computing the targets available for each positioner.  Then these available targets were removed from the input sample and the remaining targets were plotted.  Starting with the old legacy focalplane model (with equal 3mm arm lengths) you can see good coverage across the patrol radius.  With the more realistic focalplane, you can see some unreachable targets near the center of each positioner and more around the edge of the patrol radius due to varying arm lengths.  The final plot shows the commissioning focalplane with restricted phi range.  Compare these plots to the ones in #266.

Note to @Srheft and @djschlegel , the current master branch did **NOT** apply the phi angle restriction correctly, which means that previous runs did not limit the phi range properly.  This likely explains why you were assigning targets that were then rejected by the online system.  This is now fixed, as you can see in the plots.

unreachable_2012-06-01
unreachable_2020-01-01
unreachable_2020-04-01

Closes #266.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -3,11 +3,19 @@
 fiberassign change log
 ======================
 
-1.4.0 (unreleased)
+1.4.1 (unreleased)
 ------------------
 
-* No changes yet.
+* Fix bug in the range checking of positioner theta / phi angles (PR `#267`_).
+* Move the checks for positioner reachability from the assignment code to the
+  TargetsAvailable class (PR `#264`_).
+* Use a specific rundate for unit tests, to ensure consistent focalplane
+  model (PR `#262`_).
 
+.. _`#267`: https://github.com/desihub/fiberassign/pull/267
+.. _`#264`: https://github.com/desihub/fiberassign/pull/264
+.. _`#262`: https://github.com/desihub/fiberassign/pull/262
+  
 1.4.0 (2020-03-19)
 ------------------
 
--- a/py/fiberassign/test/test_hardware.py
+++ b/py/fiberassign/test/test_hardware.py
@@ -4,7 +4,11 @@ Test fiberassign target operations.
 
 import unittest
 
+from datetime import datetime
+
 import numpy as np
+
+import desimodel.io as dmio
 
 from fiberassign.utils import Timer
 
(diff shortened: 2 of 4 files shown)


Srheft:
@tskisner  wow! I'm going to compare the new run of the same observed tiles with the the ones passed to ICS on the nights of 14th and 15th and see if we can spot the rejected cases in difference between these two runs. Thanks much!

michaelJwilson:
Given that some of the online ics feedback was lost, is it possible to rerun the assignments and deduce what fibers should have been reported?  Or is this not relevant / the case anymore.

forero:
I am running the code. Just give me on more hour to inspect the results.

tskisner:
@forero , of course, no problem.

tskisner:
Ok, I'll update the changelog and then merge.

PR #268, opened 2020-05-13
Remove redundant ORs with mws scnd bits

apcooper:
The creation of `sv1_sciencemask` in `targets.py` involves the following:

```python
if "SCND_ANY" in desi_mask.names():
        sciencemask |= desi_mask["SCND_ANY"].mask
        sciencemask |= sv1_scnd_mask['MWS_WD_SV']
        sciencemask |= sv1_scnd_mask['MWS_SPECIAL_DRACO_SV']
        :
        :  
        : etc.
 else:
        sciencemask |= desi_mask["SECONDARY_ANY"].mask
        sciencemask |= sv1_scnd_mask['MWS_WD_SV']
        sciencemask |= sv1_scnd_mask['MWS_SPECIAL_DRACO_SV']
        :
        : 
        : etc. 
```

Since all targets with `SCND_ANY` are `|=`'d on the first line, the subsequent `|=`'s with individual MWS secondary bits don't have any effect (any of those bits being set implies `SNCD_ANY` is set). This PR just removes those lines.

The following points are redundant after that change, but for completeness:
- Even if it was necessary to `|=` against the MWS bits, there's no need to do that inside the `if`, which seems only to be there to deal with the name of `SCND_ANY`;
- Maintaining a separate list of secondary bits inside FA is hard to maintain, because it requires lock-step updates with any changes to the mask in `desitarget` (this was the motivation for the PR).

There may be good reasons to remove the check on `SCND_ANY` and only whitelist (or perhaps blacklist) individual bits, if some of the SCND targets shouldn't be part of the 'science mask' (e.g. `VETO`?). I think that's a separate problem.

(If this looks familiar: the same change was suggested alongside an actual bugfix in a previous PR. Before that PR was reviewed, the bug was fixed separately, so the PR got deleted as a duplicate.)

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -11,10 +11,12 @@ fiberassign change log
   TargetsAvailable class (PR `#264`_).
 * Use a specific rundate for unit tests, to ensure consistent focalplane
   model (PR `#262`_).
+* Simplify handling of MWS secondary bits in creating sv1_sciencemask (PR `#268`_).
 
-.. _`#267`: https://github.com/desihub/fiberassign/pull/267
-.. _`#264`: https://github.com/desihub/fiberassign/pull/264
 .. _`#262`: https://github.com/desihub/fiberassign/pull/262
+.. _`#264`: https://github.com/desihub/fiberassign/pull/264
+.. _`#267`: https://github.com/desihub/fiberassign/pull/267
+.. _`#268`: https://github.com/desihub/fiberassign/pull/268
   
 1.4.0 (2020-03-19)
 ------------------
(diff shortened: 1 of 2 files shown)


sbailey:
This looks good to me, but pinging @geordie666 as a cross check: will secondary target mask bits *always* be used for science-like targets with normal PRIORITY, NUMOBS_MORE rules, such that desi_mask["SCND_ANY"] can be used to flag science-like targets (as opposed to SKY and STD which follow special rules)?  I think that is the case, but I want to make sure that we are in agreement about the future usage of SCND bits.

geordie666:
@sbailey: It's certainly my intention that the secondary target mask will _always_ be used for science-like targets. 

I suppose it's possible that somebody might ask for a special program of skies or standards, but if that special program wasn't "science-like" then we'd have to get creative.

sbailey:
@geordie666 thanks for the cross check.  Let's hold the line that secondary targets are for secondary science-like targets and not other special cases.  This PR looks good.  Merging now.

PR #274, opened 2020-10-05
Enable easier embedding of fiberassign in other codes

tskisner:
This is a work in progress.  Also here is a prototype pip weight script that might eventually move elsewhere.

### DIFF
--- a/py/fiberassign/assign.py
+++ b/py/fiberassign/assign.py
@@ -32,10 +32,10 @@ from desitarget.targetmask import desi_mask
 
 from ._version import __version__
 
-from .utils import Logger, Timer, default_mp_proc
+from .utils import Logger, Timer, default_mp_proc, GlobalTimers
 
-from .targets import (TARGET_TYPE_SKY, TARGET_TYPE_SUPPSKY,
-                      TARGET_TYPE_SAFE, desi_target_type,
+from .targets import (TARGET_TYPE_SCIENCE, TARGET_TYPE_SKY, TARGET_TYPE_SUPPSKY,
+                      TARGET_TYPE_STANDARD, TARGET_TYPE_SAFE, desi_target_type,
                       default_target_masks, default_survey_target_masks)
 
 from .hardware import (FIBER_STATE_UNASSIGNED, FIBER_STATE_STUCK,
@@ -1131,13 +1131,13 @@ def merge_results_tile(out_dtype, copy_fba, params):
         if len(outdata[c].shape) < 2: #Don't propagate 2D target columns into FIBERASSIGN HDU
             cols_to_keep.append(c)
     outdata = outdata[cols_to_keep]
-    
+
     cols_to_keep = list()
     for c in tile_targets.dtype.names:
         if len(tile_targets[c].shape) < 2: #Don't propagate 2D target columns into TARGETS HDU
             cols_to_keep.append(c)
     tile_targets = tile_targets[cols_to_keep]
-    
+
     # Create the file
     if os.path.isfile(outfile):
         os.remove(outfile)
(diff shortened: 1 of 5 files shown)


tskisner:
Mainly this PR just moves the sequence of assignment steps into a single function that can be called both by the commandline tools and also in external code.  Additionally it makes some of the C++ properties of tiles to be read/write.

bin/fba_pipweights
```diff
@@ -0,0 +1,197 @@
...
+
+import argparse
+
+from bitarray import bitarray
+
+import fitsio
+
+import fiberassign
+
+from fiberassign.utils import Logger, distribute_discrete
```
sbailey, line 20: `distribute_discrete` was added to utils and imported here, but doesn't appear to be used.  Is this a leftover of some other parallelism testing?

py/fiberassign/utils.py
```diff
@@ -59,3 +59,76 @@ def option_list(opts):
                 else:
                     optlist.append("{}".format(val))
     return optlist
+
+# This is effectively the "Painter's Partition Problem".
+
+def distribute_required_groups(A, max_per_group):
+    ngroup = 1
```
sbailey, line 66: Please include a docstring for `distribute_required_groups` and `distribute_partition`

sbailey (commented):
Refactor looks fine, with a couple requests for docstrings, and an inline question about some code that was added but doesn't seem to be used.

Regarding `fba_pipweights`, I'm a bit nervous about having analysis-style code getting into the core fiberassign repo which might start mixing up tags needed for evolving analysis needs vs. tags needed for core operations work.  Maybe ok to include this here as an example, while raising that concern as something to watch out for.  If needed that could be moved into an analysis repo later.

tskisner:
Thanks for the review, I agree we could move the fba_pipweights script somewhere else (and the discrete distribution helper function).  Any suggestions for an appropriate repo?

sbailey:
I'm not sure if the ideal analysis repo exists yet for `fba_pipweights`.  Possibilities for now:
  * desihub/tutorials (which has 5 fiberassign-related notebooks already)
  * a new fiberassign_examples repo for cookbook-style examples
  * leave it here, and migrate later if it becomes a problem

Perhaps @ashleyjross has an LSS catalog repo where `fba_pipweights` should live?

ashleyjross:
I think any of those three options are good for now. 

 

I don't think it really fits in any of the current LSS stuff, which is all sandbox level at this point anyway.

 

Thanks,

Ashley

ashleyjross:
What is that status here? Once it is officially available, there are people ready to use it.

 

(If you want to add it to a new folder under LSS, e.g. LSS/Sandbox/PIP that would also be perfectly fine for me.)

 

Thanks,

Ashley

 

From: Ashley Ross <ashley.jacob.ross@gmail.com>
Date: Friday, October 23, 2020 at 12:34 PM
To: desihub/fiberassign <reply@reply.github.com>, desihub/fiberassign <fiberassign@noreply.github.com>
Subject: Re: [desihub/fiberassign] Enable easier embedding of fiberassign in other codes (#274)

 

I think any of those three options are good for now. 

 

I don't think it really fits in any of the current LSS stuff, which is all sandbox level at this point anyway.

 

Thanks,

Ashley

tskisner:
Ok, this script is not even complete, so I think the most important thing is to get it into a place where people are not afraid to hack on it.  Sounds like the LSS repo would be good.  I'll open a PR to move it over there.

ashleyjross:
Sounds good; I think you can just add the new folder to master, no PR necessary there (for better or worse).

 

Thanks!

Ashley

tskisner:
Ok, I have removed the pipweight script and squashed the remaining commits. This PR now just includes the changes that make it easier to call fiberassign within other codes, and also the changes to support modifying Targets and Tiles in memory for dithering.  I'll move the pip weight work in progress to LSS.

tskisner:
Unit tests run locally...  Will merge after travis runs someday.

Issue #275, opened 2020-11-01
Designing Commissionning fibre-assign files


araichoor:
Principle:
Having one checked-in script (cmx-assign.py) which generates the fiberassign-{tileid}.fits files for the re-start of the commissioning, to ensure reproducibility/archiving.

Approach:
I have currently designed five commissioning flavors; it should be easy to add few other ones.

forero:
Currently the script lives in https://github.com/desihub/fiberassign/blob/cmx_assign/bin/cmx-assign.py

forero:
A first version of this script has been added in PR #277

PR #277, opened 2020-11-10
Cmx fiberassign

forero:
This PR adds the fba script for commissioning started by @anand_raichoor  

It needs desitarget master to work.

It should work both on NERSC and KPNO

It can be run as follows:

`fba_cmx --dr dr8 --dtver 0.43.0 --rundate 2020-03-06T00:00:00 --seed 77  --intileid 1513  --tileid 1513  --obscon bright  --flavor dithlost  --outdir ./001513_bright_dithlost`

that line should produce the following output

https://github.com/forero/cmx-desi/tree/main/cmx_fiberassign/001513_bright_dithlost


forero:
This PR addresses #275

bin/fba_cmx
```diff
@@ -0,0 +1,817 @@
...
+dotile= True
+dosky = True
+dostd = True
+dogfa = True
+dotarg= True
+dofa  = True
+doplot= True
+
+
+# AR reading arguments
```
tskisner, line 57: Do we want to put this code in a `main()` and call that at the bottom?

bin/fba_cmx
```diff
@@ -0,0 +1,817 @@
...
+    else:                      fdict['msks'] = 'SV0_WD,SV0_BGS,SV0_MWS_FAINT'
+elif (args.flavor=='focus'):
+    sys.exit('not implemented yet; exiting')
+if (fdict['seed']!=-1):
+    np.random.seed(fdict['seed'])
+
+
+
+# AR directories
+hostname = os.getenv('HOSTNAME')
```
tskisner, line 174: This seems fragile- are we only running this script at NERSC and KPNO?  And only on cori?  I guess if this is just a quick-and-dirty thing for the next couple months then that is fine, since the next machine will not be online yet.

bin/fba_cmx
```diff
@@ -0,0 +1,817 @@
...
+
+# AR extra-hdu for dithering
+extradatamodel = np.array([], dtype=[
+    ('UNDITHER_RA','>f8'),('UNDITHER_DEC','>f8'),('TARGETID', '>i8')
+    ])
+
+
+
+# AR obscon
+if   (args.obscon=='dark,bright'): obscon = 'DARK|GRAY|BRIGHT'
```
tskisner, line 240: This is somewhat unusual style.  I know DESI has very loose coding standards, but we might consider running this script through an autoformatter like `black` or `autopep8`.

bin/fba_cmx
```diff
@@ -0,0 +1,817 @@
...
+elif (args.obscon=='dark'):        obscon = 'DARK|GRAY'
+else:                              obscon = 'BRIGHT' # AR bright
+log.info('{:.1f}s\tsetting obscon = {}'.format(time()-start,obscon))
+
+
+
+# AR get matching index for two np arrays, those should be arrays with unique values, like id
+# AR https://stackoverflow.com/questions/32653441/find-indices-of-common-values-in-two-arrays
+# AR we get: A[maskA] = B[maskB]
+def unq_searchsorted(A,B):
```
tskisner, line 250: If we had a `main()`, this would go outside of it.  Or if it is generally useful we could move it to the utils package or into `desiutils`.

bin/fba_cmx
```diff
@@ -0,0 +1,817 @@
...
+        os.rename(tmpfn,outfn)
+        log.info('{:.1f}s\tmtl targets written to {} , moved to {}'.format(time()-start,tmpfn,outfn)) 
+    else:
+        log.info('{:.1f}s\tmtl targets NOT written to {} (0 targets to write)'.format(time()-start,tmpfn)) 
+    return True
+
+
+
+# AR tiles
+if (dotile==True):
```
tskisner, line 291: This is redundant:  dotile is already a bool, so `if(dotile)` is fine.  Same with other cases below.

tskisner (commented):
This review is somewhat pointless since the PR was already merged.  If my review is requested in the future, please just ping me if I do not get to it quickly enough (by direct email or `@mention`).  I think that overall we should consider running an autoformatter on code.  I usually run `black` locally when working on fiberassign python code, but there are other tools out there too.  Anything to make sure that the code has basic compliance to PEP8.

Issue #271, opened 2020-09-21
review, simplify, and finalize fiberassign output format


sbailey:
Related to dangling PR #254: review the fiberassign output formats, simplify/trim them, and finalize it on something we are happy to use for years to come.  A non-exhaustive list of items to check:
* The lightweight `fba-*.fits` format was designed to minimize I/O and simplify debugging and studies with mocks, but it lacks the full information needed for actual observations, which are post-facto merged into the full `fiberassign-*.fits` format.  The HDUs from `fba` are propagated forwards so that any QA scripts on the `fba` files also works on the `fiberassign` files, but this results in replicated information
  * FAVAIL vs. POTENTIAL_ASSIGNMENTS
  * FASSIGN vs. FIBERASSIGN
* How much targeting information should be kept for targets that were assigned?
  * and should this be kept in the FIBERASSIGN HDU or a separate TARGETS HDU?
* How much targeting information should be kept for targets that were reachable but not assigned?
  * previous default was everything, including for reachable SKY targets, but that involved an HDU with 80% default values due to most columns for Tractor-targets not applying to the much larger number of SKY targets...
* Do we need to maintain backwards compatibility with tiles that were already observed and have a non-ideal format?
* Find and review emails on the desi-data list (and possibly elsewhere) before re-inventing the wheel again

Finalize this before restarting observations in Fall 2020.

tskisner:
Just had an adhoc conversation with @forero and @geordie666 on a cancelled zoom call...  Testing at KPNO on a couple hundred tiles showed:

1.  Out of a 1.5 hour run time, one hour was spent on the merging step
2.  Merging all the input columns ran out of memory.

Here is a proposition:  can fiberassign just write out the minimal file needed for ICS and then we can do the merging later at NERSC as a convenience step?  We could add some extra header keys with checksums of the input target files, to ensure that post-facto merging is using the same target files.  I think the minimal file format would include:

- The main fiberassign HDU with only assigned targets and minimal columns needed by ICS
- The potential / available targets HDU with the current columns plus one additional bitfield to be used with #182 
- Any sky / gfa / other HDUs needed

The result would be that the merged files would still be needed for fancy plots or any QA that needed additional properties of available targets, but actually running the assignment in operations would be fast and light.

forero:
From [desi-data 5128]:

Suggested fiberassign file columns (minimally needed for ops, plus a few more for future developments, keeping more columns for the assigned targets than the potential targets):

[...]

PR #279, opened 2020-11-25
Less target columns in fiberassign files

forero:
Addresses #271 by reading a minimal set of columns from the target files to be propagated into the `fiberassign` files.


tskisner (approved):
Looks good, thanks for implementing this!

forero:
Thanks @tskisner !

PR #282, opened 2020-12-03
adding SUPP_SKY targets to OBJTYPE=SKY

araichoor:
This PR is to set OBJTYPE="SKY" for SUPP_SKY targets in the assign files (those currently have OBJTYPE="TGT").
The modification is only two lines in py/fiberassign/assign.py to check DESI_TARGET against desi_mask.SUPP_SKY, when determining the OBJTYPE value.


sbailey (approved):
Looks good, thanks.

PR #283, opened 2020-12-06
Fba cmx pm gfa

araichoor:
This PR adds a proper motion correction to GFA_TARGETS passing the Gaia astrometric excess noise (AEN) criterion.
It also modifies the GUIDE_FLAG/ETC_FLAG/FOCUS_FLAG value (-> 2**4) of GFA_TARGETS which do not pass the AEN criterion.
In this way, only GFA_TARGETS with properly updated positions should be used for guiding.

Two additional minor corrections (addition of ISDITH keyword + correct form of Gaussian for plotting in QA plots)


schlafly (approved):
Looks good to me.  Ready to merge.

PR #292, opened 2020-12-15
more accurate proper motion correction using tools from legacypipe

araichoor:
This PR implements a more accurate proper motion correction, with using tools from legacypipe.

### DIFF
--- a/bin/fba_sv1
+++ b/bin/fba_sv1
@@ -9,15 +9,15 @@ from astropy.io import fits
 from astropy.table import Table
 import fitsio
 from desitarget.io import read_targets_in_tiles, write_targets, write_mtl
-from desitarget.sv1.sv1_targetmask import desi_mask
-from desitarget.sv1.sv1_targetmask import bgs_mask
+from desitarget.sv1.sv1_targetmask import desi_mask, bgs_mask
+from desitarget.cmx.cmx_targetmask import cmx_mask
 from desitarget.targetmask import obsconditions
 from desitarget.targets import set_obsconditions
 from desimodel.footprint import is_point_in_desi
 from desimodel.focalplane.geometry import get_tile_radius_deg
 import desimodel.io as dmio
 from fiberassign.scripts.assign import parse_assign, run_assign_bytile, run_assign_full
-from fiberassign.scripts.merge import parse_merge, run_merge
+from fiberassign.assign import merge_results
 from fiberassign.utils import Logger
 import fiberassign
 from time import time
@@ -31,6 +31,22 @@ from astropy.time import Time
 from argparse import ArgumentParser
 from desiutil.redirect import stdouterr_redirected
 import matplotlib.image as mpimg
+
+
+# AR authorized args.faflavor
+faflavors_all = [
+    "cmxm33",
+    "sv1elg",
+    "sv1elgqso",
+    "sv1lrgqso",
+    "sv1bgsmws",
+    "elg",
+    "lrgqso",
+    "bgsmws",
+    "1percdark",
+    "1percbright",
+]
+
 
 # AR copied from make_mtl()
 mtldatamodel = np.array(
@@ -78,7 +94,7 @@ def get_isaen(g, aen):
     )
 
 
-# AR courtesy of DL : adapter from legacypipe.survey
+# AR courtesy of DL : adapted from legacypipe.survey
 # AR originally named "radec_at_mjd()", renamed to get_nowradec
 def get_nowradec(ra, dec, pmra, pmdec, parallax, ref_year, new_obstime):
     # AR new_obstime = Time.now()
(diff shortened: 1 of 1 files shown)


bin/fba_sv1
```diff
@@ -78,24 +78,78 @@ def get_isaen(g, aen):
     )
 
 
-# AR get the updated ra,dec with proper motion
-def get_nowradec(ra, dec, pmra, pmdec, parallax, ref_epoch, new_obstime):
+# AR courtesy of DL : from astrometry.net's util/starutil_numpy.py
+def datetomjd(d):
+    d0 = datetime(1858, 11, 17, 0, 0, 0)
+    dt = d - d0
+    return dt.days + (dt.seconds + dt.microseconds / 1e6) / 86400.0
```
schlafly, line 85: You might compare the results of this with astropy.time.Time(date).mjd.  I hadn't used apply_space_motion before, but I have used the astropy.time library a lot, and would expect this to work and be more obvious.

bin/fba_sv1
```diff
@@ -78,24 +78,78 @@ def get_isaen(g, aen):
...
+
+
+# AR courtesy of DL : adapted from legacypipe.survey
+def mjd_to_year(mjd):
+    # mjd_to_year(57205.875) -> 2015.5
+    daysperyear = 365.25  # Julian years, by definition
+    J2000 = datetime(2000, 1, 1, 12, 0, 0, 0, tzinfo=None)
+    mjd2k = datetomjd(J2000)
+    return (mjd - mjd2k) / daysperyear + 2000.0
```
schlafly, line 94: This is likewise Time(mjd, format='mjd').jyear.

bin/fba_sv1
```diff
@@ -78,24 +78,78 @@ def get_isaen(g, aen):
...
-        pm_dec=pmdec * units.mas / units.yr,
-        frame="icrs",
-        obstime=Time(ref_epoch, format="jyear"),
-        distance=Distance(parallax=parallax_clip * units.mas),
-    )
-    nowc = c.apply_space_motion(new_obstime=new_obstime)
-    return nowc.ra.value, nowc.dec.value
+    equinox = 53084.28  # mjd of the spring equinox in 2004
+    daysperyear = 365.25  # Julian years, by definition
+    axistilt = 23.44  # degrees
```
schlafly, line 116: I might be inclined to define djyear = Time(mjd, format='mjd').jyear - Time(equinox, format='mjd').jyear and use that in the definition of suntheta below, rather than defining daysperyear here.

bin/fba_sv1
```diff
@@ -78,24 +78,78 @@ def get_isaen(g, aen):
...
+    - pmra,pmdec are in mas/yr.
+      pmra is in angular speed (ie, has a cos(dec) factor)
+    - parallax is in mas.
+    Returns: RA,Dec
+    """
+    mjd = new_obstime.mjd
     # AR new_obstime = Time.now()
     # AR clipping Gaia PARALLAX to >1e-3 (setting distance at <1Mpc)
     parallax_clip = parallax.copy()
     parallax_clip[(~np.isfinite(parallax)) | (parallax < 1e-3)] = 1e-3
```
schlafly, line 113: We should discuss on list if this is still desired.  This was necessary when I thought that apply_space_motion was going to be the logical route, since a negative distance doesn't make a lot of sense.  But a negative parallax is fine in the context of this code.  We are going to be inserting a small noise bias if we clip the parallaxes in this way, so I would probably vote for deleting the clipping now.

schlafly (approved):
The only other suggestion I would make would be to get Dustin's results and check that they're identical.  IIRC they were very similar before, but since apply_space_motion wasn't applying parallax, there were very small differences everywhere, and since astropy didn't like some proper motions when they implied space motions that were >speed of light, it didn't apply corrections for some sources at all.  Both of those sources of disagreement should disappear.  Maybe some tiny residual related to different handling of time / leap seconds / ... if you follow my suggestions of using astropy.time code in place of Dustin's routines.

bin/fba_sv1
araichoor: correct; actually I "mjd" comes from "new_obstime.mjd"; so I can simply use "new_obstime.jyear", and forget about the mjd.

bin/fba_sv1
araichoor: correct; besides the datetomjd() is just used once in "mjd2k = datetomjd(J2000)".

bin/fba_sv1
araichoor: I did check for one new_obstime, and it provides the same value, so I ll go for it.

bin/fba_sv1
araichoor: I agree with you, I ve currently removed the parallax clipping

araichoor:
The pm-correction change has been commited in https://github.com/desihub/fiberassign/commit/c7962726fb9cb6beac160b2946944e2ae999aa23.
Closing the PR.

PR #307, opened 2021-02-04
Fba cmx update

forero:
Updates to produce new dither tiles

The following commands can be used to produce the results stored here https://data.desi.lbl.gov/desi/users/forero/tiles_dither_february/

`fba_cmx --dr dr9 --dtver 0.49.0 --rundate 2021-01-20T00:00:00 --seed 77  --tilera 188.0  --tiledec 30.0 --tileid 80743  --faflavor dithprec  --outdir ./080743`
`fba_cmx --dr dr9 --dtver 0.49.0 --rundate 2021-01-20T00:00:00 --seed 77  --tilera 111.64  --tiledec 30.0 --tileid 80756  --faflavor dithprec  --outdir ./080756`
`fba_cmx --dr dr9 --dtver 0.49.0 --rundate 2021-01-20T00:00:00 --seed 77  --tilera 111.64  --tiledec 30.0 --tileid 80769  --faflavor dithfocus --outdir ./080769`

### DIFF
--- a/bin/fba_cmx
+++ b/bin/fba_cmx
@@ -232,7 +232,7 @@ def main():
         "starfaint",
         "scidark",
         "scibright",
-        "focus",
+        "dithfocus",
     ]:
         log.error(
             "args.faflavor not in dithprec,dithlost,starfaint,scidark,scibright,focus; exiting"
@@ -307,9 +307,15 @@ def main():
     elif args.faflavor == "scibright":
         # fdict['msks'] = 'SV0_WD,MINI_SV_BGS_BRIGHT,SV0_MWS_FAINT'
         fdict["msks"] = "SV0_WD,SV0_BGS,SV0_MWS_FAINT"
-    elif args.faflavor == "focus":
-        log.error("faflavor==focus not implemented yet; exiting")
-        sys.exit()
+    elif args.faflavor == "dithfocus":
+        if tile_in_desi == 0:
+            fdict["msks"] = "STD_DITHER_GAIA"
+        else:
+            fdict["msks"] = "STD_DITHER"
+        fdict["seed"] = args.seed
+        fdict["ndither"] = 12
+        fdict["gfrac"] = 1.0
+        fdict["gwidth"] = 2.0    
     if fdict["seed"] != -1:
         np.random.seed(fdict["seed"])
 
@@ -325,7 +331,7 @@ def main():
         )
 
     mydirs = {}
-    if (args.faflavor in ["dithprec", "dithlost"]) & (tile_in_desi == 0):
+    if (args.faflavor in ["dithprec", "dithlost", "dithfocus"]) & (tile_in_desi == 0):
         mydirs["targ"] = os.path.join(
             path_to_targets, "gaiadr2", args.dtver, "targets/cmx/resolve/supp"
         )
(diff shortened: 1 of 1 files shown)


araichoor:
(commenting by order of appearance; I didn t run the code, comments are just from reading it; I hope they are relevant)

**"shutil.move"** : is there a reason to change "os.rename" into "shutil.move"?

https://github.com/desihub/fiberassign/blob/a977cb8a9d6712f6f56aad50f26aec12dc381a71/bin/fba_cmx#L76
I think that the "tmpoutdir" argument is useless here: it is used in write_mtl(); but write_mtl() already creates a dedicated folder to write its output (like mtldir + "dr9/0.50.0/mtl/sv1/no-obscon/"), and after calling write_mtl(), we move that output to "outfn".

**custom_read_targets_in_tiles():**
https://github.com/desihub/fiberassign/blob/a977cb8a9d6712f6f56aad50f26aec12dc381a71/bin/fba_cmx#L137-L162
I think it s not required to add that function here; the need of it in fba_sv1 is when querying targets both in the ls-dr9/ and the gaiadr2/ folders; I think we won t use that for fba_cmx if fba_cmx is used for dithering only.
though, it doesn t hurt to add it... if keeping it, then also replace read_targets_in_tiles() in dosky (see how that changes in fba_sv1), dostd, dotarg, to be consistent.

**"write_targets(tmpoutdir,..)" instead of "write_targets(args.outdir,...)"**
for the same reason as above, I think this change is un-necessary
https://github.com/desihub/fiberassign/blob/a977cb8a9d6712f6f56aad50f26aec12dc381a71/bin/fba_cmx#L623
https://github.com/desihub/fiberassign/blob/a977cb8a9d6712f6f56aad50f26aec12dc381a71/bin/fba_cmx#L647

**log message "start generating..."**
https://github.com/desihub/fiberassign/blob/a977cb8a9d6712f6f56aad50f26aec12dc381a71/bin/fba_cmx#L634
it would be nice to also add it for dosky, dostd, dotarg.

**renaming of fba*fits files:**
https://github.com/desihub/fiberassign/blob/a977cb8a9d6712f6f56aad50f26aec12dc381a71/bin/fba_cmx#L880-L888
https://github.com/desihub/fiberassign/blob/a977cb8a9d6712f6f56aad50f26aec12dc381a71/bin/fba_cmx#L1029-L1038
maybe those steps are now useless, with the new use of tmpoutdir in the run_assign_full, run_merge? don t know for sure without running the code.

**default rundate:**
I ll update it to full reach petal0, ie 2021-01-19T00:00:00
https://github.com/desihub/fiberassign/blob/a977cb8a9d6712f6f56aad50f26aec12dc381a71/bin/fba_cmx#L1485-L1492

**black syntax formatting:**
lastly, if not done, maybe run a "black fba_cmx" before re-submitting, so that the syntax is standardized.

forero:
Thanks for the review.

Here are my replies:
* The shutil is needed because the tmp file is created on a different file system than $SCRATCH. shutil works regardless of the filesystem.
* I had to add `"tmpoutdir` to avoid the code crashing when I had multiple runs at the same time. There is a racing conditions if multiple proceses try to write and move the same file. 
* I will remove `custom_read_targets`.
* I will add `log message "start generating..."` to the other steps
* I will remove the creation of tmp files in the fiberassign-merge process.
* I will  update the defaul rundate.

PR #309, opened 2021-02-11
remove -Wstrict-prototypes

marcelo-alvarez:
This branch corrects a distutils bug that adds spurious -Wstrict-prototypes flag when compiling c++ extensions, by removing the flag in setup.py.


sbailey:
On travis the -Wstrict-prototypes flag wasn't there, causing a test build failure when trying to remove it.  Added a check and retrying.

PR #313, opened 2021-03-12
Add github workflow to run unit tests on Linux / MacOS

tskisner:
Add a new simple workflow that tests python 3.6, 3.7, and 3.8 on ubuntu-latest and macos.


weaverba137:
Let me know when the tests are mostly passing.  I will set up coveralls at that point.

sbailey:
What's the status here?  I see that the macos tests take longer than the others, and one timed out.  This is the only repo where we've tried mac tests; I suggest we drop back to just testing on linux.  Any objections?  I also suggest that we *not* wait for coveralls as part of this PR and address that separately.

i.e. my suggestion is to:
  * drop macos tests
  * wait for tests to repass
  * merge

should we be waiting for anything else?

tskisner:
I am actively debugging the underlying error locally.  That is why I bundled the test artifact and downloaded it.  My concern is not OSX per se but if that points to some other issue.  This branch only "passes" because I have disabled the check for errors in order to inspect the outputs.  Feel free to contact me by phone or slack with questions.

tskisner:
This is now ready for review @araichoor .  It fixes up the unit tests and runs them with github actions when a PR is created against master.  After this is merged I'll open a separate PR for the desimodel support.

@weaverba137 , let's punt the coveralls integration to a separate PR after this one, so that it can proceed in parallel with other fiberassign work.

weaverba137:
Fine with me.

py/fiberassign/assign.py
```diff
@@ -823,57 +823,57 @@ def merge_results_tile_initialize(tgbufs, tgdtypes, tgshapes, skybufs,
...
-    #('SCND_TARGET', '>i8'), 
-    ('SUBPRIORITY', '>f8'), 
-    ('OBSCONDITIONS', '>i8'), 
-    ('PRIORITY_INIT', '>i8'), 
-    ('NUMOBS_INIT', '>i8')   
+    #('SCND_TARGET', '>i8'),
+    ('SUBPRIORITY', '>f8'),
+    ('OBSCONDITIONS', '>i8'),
+    ('PRIORITY_INIT', '>i8'),
+    ('NUMOBS_INIT', '>i4')
```
sbailey, line 876: Although NUMOBS_INIT as i4 is a more appropriate type than the original i8, this is another data model change for fiberassign output that creates bookkeeping hassle for downstream code that does things like stack these tables into a catalog.  We're already dealing with other type and column changes so this doesn't have to be a blocking factor, but I wanted to highlight this in case others anticipate problems.

py/fiberassign/test/test_qa.py
```diff
@@ -258,16 +258,30 @@ def test_science(self):
...
+        # fstring notation if we increase the python version requirement in setup.py.
+        # This mechanism for finding the qa-fiberassign script only works when running
+        # tests in place, and not when running installed tests with:
+        #
+        #  $> python -c 'import fiberassign.test; fiberassign.test.runtests()'
+        #
+        # The solution is to move qa-fiberassign into the fiberassign.scripts package
+        # so that it can be tested along with the other scripts (qa-fiberassign is
+        # a legacy script from before we refactored the package.)  Once that is done,
+        # Re-enable this test and call the entry point directly.
```
sbailey, line 272: f-string notation is fine; we now use that extensively in desispec and py3.6 is the minimum supported version now.

Auto-discovery of where `qa-fiberassign` is and actually running that is still desirable but doesn't have to be a blocking factor for this PR.  See [`desispec.test.test_binscripts`](https://github.com/desihub/desispec/blob/master/py/desispec/test/test_binscripts.py) for a multi-step discovery of where the scripts might be depending upon whether you are testing a bare git clone or an installed package.  And it has been occasionally useful to test that the actual script works, and not just the `desispec.scripts.blah` version of it works.

py/fiberassign/test/test_vis.py
```diff
@@ -151,11 +151,11 @@ def test_plotpos(self):
...
-        time = "2012-12-12T00:00:00"
-        suffix = "{}_simple".format(time)
-        self._load_and_plotpos(time, test_dir, suffix, simple=True)
-        suffix = "{}".format(time)
-        self._load_and_plotpos(time, test_dir, suffix, simple=False)
+        # time = "2012-12-12T00:00:00"
+        # suffix = "{}_simple".format(time)
+        # self._load_and_plotpos(time, test_dir, suffix, simple=True)
+        # suffix = "{}".format(time)
+        # self._load_and_plotpos(time, test_dir, suffix, simple=False)
```
sbailey, line 158: What part of this doesn't work any more?  Or is this a leftover debugging comment-out?

sbailey (commented):
Adding github actions is good, thanks.  Could you clarify what the other intended changes are?  I see a NUMOBS_INIT type change and a few tests commented out, but it is unclear to me what the fundamental underlying issue with the failing tests was and whether that is now addressed, or postponed.

py/fiberassign/assign.py
tskisner: Ok, will change to i8.  This column is used by some of the target QA.  When it was removed, in the changes to the file format, this broke that QA.  However, if we really don't want to propagate this column, then I can disable the unit tests that depend on it.  I guess I was the only one using that target QA :-)

py/fiberassign/test/test_qa.py
tskisner: Ok, I will bump the requirements to python >= 3.6.  Will look at the desispec code referenced and if it is quick to get in will do it before merge.

py/fiberassign/test/test_vis.py
tskisner: Forgotten comment-out, will fix.

py/fiberassign/assign.py
sbailey: Please don't re-add a column that had been removed unless there has been group discussion that re-adding it is really necessary.  I thought you were just changing type from i8 to i4 which could be disruptive but ok, but adding and subtracting columns is a bigger deal that shouldn't be done lightly now that we are actively observing.

tskisner: Ok, I guess my point was that removing the column broke things, but this went undetected since unit tests were not being run.  Restoring this column fixes the unit tests.  If restoring this column breaks downstream code, then I guess we should disable the target QA code that needs it.  I can do that with a separate small PR if you like.

sbailey: Removing NUMOBS_INIT was part of a purposeful trimming down of the fiberassign format that was discussed by the entire group and jointly decided upon.  It was an error that we weren't running unit tests and didn't realize it, leading us to not realize that it broke qa-fiberassign, but the fix now is to update qa-fiberassign to not need NUMOBS_INIT rather than adding that column back.  If re-removing it requires you to comment out the qa-fiberassign test now so that you can merge this and focus on the other more critical fiberassign+desimodel updates, ok.

tskisner: Just to be clear, the NUMOBS_INIT column was used by a function (qa_targets) which is separate from the `qa-fiberassign` script.  I will self-merge a PR shortly that re-removes the NUMOBS_INIT column and comments out the unit test of the qa_targets() function.

PR #315, opened 2021-03-17
Support new desimodel format which has extra state information.

tskisner:
- Change fiberassign state bits to match those used in desimodel
  for STUCK, BROKEN, and RESTRICTED states.

- Support new desimodel focalplane / state information which
  is created from a dump of the online DB.  This new format
  includes specifying the fixed theta / phi angles of positioners
  which cannot move.

- When a positioner is unmovable, we mark it as having no
  available targets.  When moving neighboring positioners,
  check for collisions with the fixed positioners using their
  constant theta / phi location.

Note:  this should not be merged until after resolving the lingering discussion on merged PR #313.  Also, there is one additional small change in progress on this branch to write out the X/Y values of unassigned positioners to their fixed theta/phi location if they are stuck / broken.


tskisner:
This is now ready for review.  In summary, there are no user-facing changes to common APIs or file formats.  The only changes that a user should notice are:

- Broken fibers and stuck positioners now have their output RA/DEC set to the projected location implied by their fixed theta / phi arm angles.  These positioners are still marked as "unassigned" with a negative target ID.  A future internal change could count these positioners as SUPP_SKY (for example).
- When using "new" desimodel data, the state of positioners and the exclusion zones for non-functioning positioners should agree with the online system and reduce or eliminate the number of rejected moves.
- The `FIBER_STATE_*` bits (written to the FIBERSTATUS column) now match the values in desimodel for stuck / broken / restricted.

forero:
I managed to run `fba_run` from this PR with the new outputs from `desi_sync_focalplane` (PR https://github.com/desihub/desimodel/pull/143)

tskisner:
Thanks to @forero and @araichoor for extensive testing. Merging this now.

PR #316, opened 2021-03-17
Remove NUMOBS_INIT column in merged TARGETS table

tskisner:
This re-removes the column and comments out the tests of the `qa_targets()` function.

### DIFF
--- a/py/fiberassign/assign.py
+++ b/py/fiberassign/assign.py
@@ -872,8 +872,7 @@ minimal_target_columns = OrderedDict([
     #('SCND_TARGET', '>i8'),
     ('SUBPRIORITY', '>f8'),
     ('OBSCONDITIONS', '>i8'),
-    ('PRIORITY_INIT', '>i8'),
-    ('NUMOBS_INIT', '>i8')
+    ('PRIORITY_INIT', '>i8')
 ])
 
 merged_fiberassign_swap = {
@@ -911,8 +910,7 @@ merged_fiberassign_req_columns = OrderedDict([
     #("NUMTARGET", "i2"),
     ("PRIORITY", "i4"),
     ("SUBPRIORITY", "f8"),
-    ("OBSCONDITIONS", "i4"),
-    ("NUMOBS_INIT", ">i8")
+    ("OBSCONDITIONS", "i4")
 ])
 
 # AR commenting out NUMTARGET ([desi-survey 1032])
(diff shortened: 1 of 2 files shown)


tskisner:
Self-merging this trivial and requested change.

sbailey:
@tskisner apologies, but I was confused by the discussion in PR #313 about NUMOBS_INIT.  I went back and checked existing files and NUMOBS_INIT *is* in the SV1 fiberassign outputs, it was *not* removed in the format trimming cleanup, and it should *not* be removed now.  i.e. we need to undo this PR (I think?), and I remain confused about why PR #313 needed to change anything about NUMOBS_INIT since the discussion there seemed to imply that it was missing and that you were adding it back, but here you seem to be removing it.  I think the confusion arises from the bin/fba_sv1 format getting out of sync with bin/fba_run + bin/fba_merge .  Maybe.

/global/cfs/cdirs/desi/target/fiberassign/tiles/trunk/080/fiberassign*.fits.fz are the file formats we are trying to preserve (still accessible via dtn01 while cori is out).

i.e. regardless of anything else I say, the goal is to *not* change the fiberassign output format other than which *DESI_TARGET columns are included depending upon SV1, SV2, main...

Mentioning @araichoor as the current defacto owner of the fiberassign format (though he's booked in the TS workshop this morning)

tskisner:
@sbailey and @araichoor , here is the history as I understand it from looking at commits touching `assign.py`:

- When loading target files for the merging step, the list of columns that are extracted from these target files is given here:  https://github.com/desihub/fiberassign/blob/master/py/fiberassign/assign.py#L824

- Anand made this commit:  which added a new list of columns, https://github.com/desihub/fiberassign/commit/d2f94e3f155f1cb208ccef2d400c08cc8121ffa9#diff-890b2f98e24e660b751aa182d690f3fc4ce97af08b1cd8f254f84ff4a48c89deR944 This is used to truncate the merged target table to a very small set of columns.  In other words, first the longer list of columns is extracted from the target files and merged, and then the result is further trimmed.  This commit is effectively where the NUMOBS_INIT column went away from the TARGET table since only that small set of columns now ends up in the output.

- I see now that *I* am actually the one who removed the NUMOBS_INIT column from the "larger" list of columns for the first time in this very PR.  However, that had no impact on the output file columns in the TARGET table, which is set by the smaller set of columns above (and which never included NUMOBS_INIT).

So, just to summarize, the columns in the output merged TARGETS table is currently set by this list:

https://github.com/desihub/fiberassign/blob/master/py/fiberassign/assign.py#L948

Which is a subset of the columns actually read from the target files, which is set by this list:

https://github.com/desihub/fiberassign/blob/master/py/fiberassign/assign.py#L948

I can restore the NUMOBS_INIT column to the larger list, but it will never end up in the output TARGET table unless it is also added to that shorter list.  So:  do you want NUMOBS_INIT in the (now very short) list of columns in the TARGET table?

PR #318, opened 2021-03-18
py/targets: updating for sv2 (will need eventually to make an overall…

araichoor:
… re-writting)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,17 +30,17 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install External Dependencies
-        run: python3 -m pip install wheel pyyaml scipy matplotlib numpy astropy sphinx fitsio healpy
+        run: python3 -m pip install --no-cache-dir wheel pyyaml scipy matplotlib numpy astropy sphinx fitsio healpy
       - name: Install desiutil
-        run: python3 -m pip install git+https://github.com/desihub/desiutil.git@${{ env.DESIUTIL_VERSION }}#egg=desiutil
+        run: python3 -m pip install --no-cache-dir git+https://github.com/desihub/desiutil.git@${{ env.DESIUTIL_VERSION }}#egg=desiutil
       - name: Install desimodel
-        run: python3 -m pip install git+https://github.com/desihub/desimodel.git@${{ env.DESIMODEL_VERSION }}#egg=desimodel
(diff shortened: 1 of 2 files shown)


forero:
@araichoor @sbailey  Maybe it's time to migrate all this code to https://github.com/desihub/desisurveyops ?

tskisner:
Looks like the unit test runner detected that sv2 target masks do not yet exist in desitarget master.  Are they in a separate PR for desitarget?  If so, then we should merge that work first before this one.

geordie666:
They exist:

https://github.com/desihub/desitarget/blob/master/py/desitarget/sv2/data/sv2_targetmask.yaml

and have for at least a few days. So the unit test isn't finding them for some reason.

tskisner:
I'll test this branch locally to see what is going on.

tskisner:
I fixed the issue in desitarget:  https://github.com/desihub/desitarget/pull/686

The sv2 data files were in the repo, but not being installed.

tskisner (commented):
The changes look fine, and are no worse than all the code in there for cmx / sv1.  I do hear @forero 's comment about all of these survey bits becoming out of hand.  Let me go see if there is a quick and easy way to simplify this.  I forget the reasoning why we have these details here rather than just relying on desitarget.

araichoor:
@tskisner : I m finding issues running at NERSC, I may have forgotten lines to add... I hope to report soon; apologies.

tskisner:
No problem, let's not merge this branch for a few hours, I'm working on seeing if we can simplify some things.

araichoor:
ok perfect, thanks for taking care of that; it would be really nice to clean that script; in case, I have the correct lines to add to the current version.

tskisner:
Ok, the only reason for all the complications in `target.py` comes from the fact that for the CMX survey the bits representing how to classify targets as "science" or "standards" were very different.  It looks like for the main survey and svX these bit definitions are the same.  So, I suggest we just have 2 cases:  cmx and everything else.  If that sounds reasonable I can implement that change.

araichoor:
I ve committed a new version of targets.py, with few additional lines to handle sv2; it should be working now.

tskisner:
I could also save that cleanup to post-sv2, which might be better...

araichoor:
yes, maybe better, because we re still running sv1... and it s not top-priority.

tskisner:
Ok, as soon as the tests pass, feel free to merge and delete the branch

PR #321, opened 2021-04-01
Change assignment of leftover fibers

tskisner:
After assigning science targets which have remaining observations, and then ensuring that the per-petal quota of standards and sky are met, there may be some unassigned fibers.  These unassigned fibers are then assigned to science, standards, and sky if possible.

In the current main branch, science targets with less than one remaining obs are never considered for assignment, even in this "last ditch" effort to assign unused fibers.

This PR changes that behavior to allow assignment to science targets with less than one obs remaining during this final attempt to assign the leftover positioners.

### DIFF
--- a/py/fiberassign/assign.py
+++ b/py/fiberassign/assign.py
@@ -328,9 +328,6 @@ def write_assignment_fits_tile(asgn, fulltarget, overwrite, params):
                 ):
                     # This positioner is not moveable and therefore has a theta / phi
                     # position in the hardware model.  Used this fixed fiber location.
-                    print("loc {}, state {} is stuck / broken, using {} / {}".format(
-                        loc, hw.state[loc], hw.loc_theta_pos[loc], hw.loc_phi_pos[loc],
-                    ), flush=True)
                     xy = hw.thetaphi_to_xy(
                         hw.loc_pos_curved_mm[loc],
                         hw.loc_theta_pos[loc],
@@ -353,9 +350,6 @@ def write_assignment_fits_tile(asgn, fulltarget, overwrite, params):
                     phmax = hw.loc_phi_max[loc] + hw.loc_phi_offset[loc]
                     if phmax > np.pi:
                         phmax = np.pi
-                    print("loc {}, state {} is unassigned, using {} / {}".format(
-                        loc, hw.state[loc], thmin, phmax,
-                    ), flush=True)
                     xy = hw.thetaphi_to_xy(
                         hw.loc_pos_curved_mm[loc],
                         thmin,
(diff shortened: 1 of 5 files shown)


araichoor:
thanks a lot @tskisner.
I ve tested it (on a rosette pattern of 9 tiles), and it works as expected, great!
I ve only one suggestion: add this "use_zero_obsremain=True" as an argument of assign.run(), run_assign_full() and run_assign_bytile() (with defaulting it to True).
that way, one can always run the fiber assignment with previous behavior, if desired.

tskisner:
Thanks for testing @araichoor, I can definitely add a switch to those functions so that the old behavior can be enabled for comparison.

araichoor:
I realize that I forgot to check if targets which were receiving a fiber before still do after this change: I notice that ~50/350 TARGET_STATE="CALIB" targets now get one observation less.
Overall, that makes ~790 CALIB fibers, against ~840 before: I don t know if that is expected or not.
And I m not if that is an issue or not: that is ~5% less CALIB only, so I suspect it is ok.

tskisner:
Hello @araichoor , I am not familiar with this `TARGET_STATE` property.  Internally in fiberassign, each target is categorized into one or more types for purposes of assignment (`SCIENCE`, `STANDARD`, `SKY`, `SUPPSKY`, and `SAFE`).  This categorization is done using the bits of DESI_TARGET and definitions from the desitarget package.  The categories of each target are stored in the bits of the `FA_TYPE` column.

Are the CALIB targets you are referring to standards, or something else?

araichoor:
I m not very familiar with those either, but with checking: all the assigned TARGET_STATE="CALIB" targets have FA_TYPE=3, so those should be fiberassign STANDARD, right?
And they have PRIORITY_INIT=-1, which should confirm their standard status.

geordie666:
Everything in the MTL ledger that is labeled `CALIB` should be a standard star, yes. But, not everything that is a standard star will be labeled `CALIB`. This is because some standards are also other target classes (MWS, etc.) and so their state can be governed by a different targets with which they have been merged.

In short: Treat `TARGET_STATE` as informational. It's a useful way for humans to parse how a target's state is changing, but the real information is in `PRIORITY`, `NUMOBS_MORE`, etc.

djschlegel:
Sorry, I had thought we were already using this behavior.  Could we merge this in and start using it?

tskisner:
Yes, I was going to add the options requested by @araichoor and then merge.  Working on it.

araichoor:
@tskisner : have you had a chance to look at this STANDARD/CALIB minor thing?
is it something expected to possibly happen? if yes, I d say it is not an issue (given that it s only 5% less CALIB fibers).
please let me know if there are further checks I could do.

tskisner:
Recall that the old behavior for leftover fibers (i.e. unassigned fibers after 10 standards / 40 sky per petal) was:

1. Assign to science with numobs_more > 0 (this was typically a no-op, since these would have already been assigned.  Only a few corner cases here if a fiber was bumped for sky/standard and then freed up a previously unassignable target)
2. Assign to any standards
3. Assign to any sky

Now we have changed step (1) for these leftover fibers to be allowed to observe *any* science target regardless of numobs_more.  So I would expect fewer "extra" observations of standards (we should always have 10 per petal though).

tskisner:
Ok, merging this now that checks have passed.  Next up is timestamps...

araichoor:
ok great then, that explains why we get slightly fewer STANDARD fibers.
I ve checked, using FA_TYPE=3 to identify standards: I confirm that with the obs_zero branch the petals where there are less standards are petals where there were >10 standards in the master branch run (except for the last petal of the last pass, where there are 9 standards instead of 10)

with the master branch (col1 = passid , col2-11: nstd per petal):
1 [10 10  9 10 10 10 10 10 10  9]
2 [10 10  9 10 10 10 10 10 10 10]
3 [10 10  9 10 10 10 10 10 10 10]
4 [10 10 10 10 10 10 10 10 10  9]
5 [10 10 10 10 10 10 10 10 10 10]
6 [10 10 10 10 10 10 10 10 10 10]
7 [10 10 15 10 10 10 10 10 10 10]
8 [11 10 21 14 10 10 10 10 10  9]
9 [15 10 23 19 10 12 15 10 10 12]

with the obs_zero branch (col1 = passid , col2-11: nstd per petal):
1 [10 10  9 10 10 10 10 10 10  9]
2 [10 10  9 10 10 10 10 10 10 10]
3 [10 10  9 10 10 10 10 10 10 10]
4 [10 10 10 10 10 10 10 10 10  9]
5 [10 10 10 10 10 10 10 10 10 10]
6 [10 10 10 10 10 10 10 10 10 10]
7 [10 10 10 10 10 10 10 10 10 10]
8 [10 10 10 10 10 10 10 10 10  9]
9 [11 10 11 10 10 10 10 10 10  9]

tskisner:
Ok, thanks for that investigation.  Working now on the UTC fixes in desimodel, and then will make a corresponding PR for fiberassign.

araichoor:
awesome, thanks a lot!

PR #328, opened 2021-04-21
bin/fba_launch: proper handling of the ISO time formatting introduced…

araichoor:
… in PR327

Accounts for the change of date format from "YYYY:MM:DDThh:mm:ss" to "YYYY:MM:DDThh:mm:ss+00:00", which were breaking some calls to astropy.time Time().

### DIFF
--- a/bin/fba_launch
+++ b/bin/fba_launch
@@ -945,7 +945,7 @@ def main():
         # AR - mjd (! TBD !)
         d = Table.read(mydirs["too"])
         keep = is_point_in_desi(tiles, d["RA"], d["DEC"])
-        mjd_now = Time(utc_time_now, format="isot").mjd
+        mjd_now = Time(utc_time_now_notz, format="isot").mjd
         keep &= (d["MJD_BEGIN"] < mjd_now + 30) & (d["MJD_END"] > mjd_now - 1)
         log.info(
             "{:.1f}s\tdotoo\tkeeping {}/{} targets in tiles and in the MJD time window: {}, {}".format(
@@ -1991,7 +1991,7 @@ if __name__ == "__main__":
     )
     parser.add_argument(
         "--rundate",
-        help="yyyy-mm-ddThh:mm:ss rundate for focalplane (default=current UTC time)",
+        help="yyyy-mm-ddThh:mm:ss+00:00 rundate for focalplane with UTC timezone formatting (default=current UTC time)",
         type=str,
         default=None,
         required=False,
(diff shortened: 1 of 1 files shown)


bin/fba_launch
```diff
@@ -2067,10 +2067,11 @@ if __name__ == "__main__":
 
     # AR rundate, pmtime
     utc_time_now = datetime.now(tz=timezone.utc).isoformat(timespec='seconds')
+    utc_time_now_notz = datetime.strptime(utc_time_now, "%Y-%m-%dT%H:%M:%S%z").strftime("%Y-%m-%dT%H:%M:%S")
```
sbailey, line 2070: Here you are stripping the timezone information off of the ISO8601 string, and then later leaving it up to astropy.Time to guess that your timestamp is in UTC.  It happens to default to that, but that seems fragile (e.g. it might have defaulted to localtime instead).  Given that we have a timezone-aware ISO8601 string already and are parsing that through a timezone-aware datetime object, let's keep that explicit, e.g.
```
utc_datetime_now = datetime.now(tz=timezone.utc)  # UTC timezone aware datetime object
...
mjd_now = Time(utc_datetime_now).mjd   # parse from datetime object instead of string
```

araichoor: ideally, yes.
I did check that Time() was defaulting to UTC, so that removing the timezone info is ok ok.

note that the two items using Time() transformation don t require a <1 day precision:
- proper motion correction
- MJD window for the ToO.

anyway, the "issue" is that there are three things at play:
- utc_time_now , the time at which the code is called (used for mtl info recording), defined in the code: https://github.com/desihub/fiberassign/blob/ab01a579c0992f71aa44de79c03fb4bfe4b0c87d/bin/fba_launch#L2069
- rundate, which is an argument, defaulting to utc_time_now
- pmtime, time used to correct for proper motion, defaulting to utc_time_now.

the change I made was "mininal" with handling those, with the current code structure.
e.g. for the pmtime: it is a string (argument), changed to time object, to then compute the mjd.
maybe the overall logic/ties between those dates should be re-worked, but I m not sure I can handle that tonight.

araichoor: @sbailey : I ve pushed two commits, treating those UTC time things more properly, as suggested.
the current code works.

noticely:
- now I request that, if provided, args.pm_time_utc_str and args.rundate follow the "YYYY-MM-DDThh:mm:ss+00:00" formatting (ie with time zone, and that time zone has to be UTC)
- so the code will return an error if one of those arguments are "YYYY-MM-DDThh:mm:ss"
is it ok?

PR #333, opened 2021-04-28
Desi fa smallrun improve

araichoor:
Multiple changes in bin/desi_fa_smallrun to prepare for the Readiness Workshop of Apr. 28th, mostly to speed up/clean the code.
Now a full dark run (7 passes) over the NGC/SGC takes 3h/1.5h on two interactive nodes.
Use Dark/Bright and sky dr9-0.57.0 catalogs copied to CSCRATCH, as reading CFS files in an interactive node is 20x slower than normal.

Most important changes:
- update_after_farun() routine faster
- when args.numproc > 1, sky files are now created on-the-fly for each tile, and are not saved
- default tile file: now DESI_ROOT/users/schlafly/tiling/tiles-4112-packing-20210329-decorated.fits; assuming this formatting

Arguments:
- new:
 x numobs_init_notqso and numobs_init_qso: forcing some values for NUMOBS_INIT
 x desi_target_env: possible use of a local DESI_TARGET variable
- rundate:
 x change in the rundate format
 x now defaults to current UTC time
- removed:
 x rmvdarkbgs
 x gray2dark

Minor changes:
- some if args.program == "dark" additions
- some default values for the arguments
- use of fitsio instead of fits to write files
- removal of some photometric columns to make files lighter
- removing cases for temporary versions of catalogs
- more proper log file
- better organized import commands
- black formatting

### DIFF
--- a/bin/desi_fa_smallrun
+++ b/bin/desi_fa_smallrun
@@ -5,26 +5,43 @@ import sys
 from glob import glob
 import numpy as np
 import astropy.io.fits as fits
-from astropy.table import Table
-import fitsio
+from astropy.table import Table, vstack
 from astropy import units
 from astropy.coordinates import SkyCoord
-from desitarget.targetmask import obsconditions
+import fitsio
+from datetime import datetime, timezone
+from time import time
+
+# AR desitarget
+import desitarget
 from desitarget.targets import encode_targetid
+from desitarget.targetmask import obsconditions
 from desitarget.io import read_targets_in_tiles
+from desitarget.geomask import match  # AR to match arrays with no repeats
 from desitarget.mtl import make_mtl, calc_priority
+from desitarget.internal import sharedmem
+
+# AR fiberassign
+import fiberassign
 from fiberassign.scripts.assign import parse_assign, run_assign_full
-from fiberassign.scripts.merge import parse_merge, run_merge
+from fiberassign.utils import Logger
+
+# AR desimodel
+import desimodel
 from desimodel.footprint import is_point_in_desi, tiles2pix
 from desiutil.redirect import stdouterr_redirected
 import healpy as hp
-from desitarget.internal import sharedmem
 from argparse import ArgumentParser
 
-# AR list of priorities (ELG=3000, LRG=3200, QSO=3400)
-# https://github.com/desihub/desitarget/blob/master/py/desitarget/data/targetmask.yaml#L186-L223
-# https://github.com/desihub/desitarget/blob/master/py/desitarget/sv1/data/sv1_targetmask.yaml#L251-L324
-# https://github.com/desihub/desitarget/blob/master/py/desitarget/sv2/data/sv2_targetmask.yaml#L152-L189
+
+# AR/SB assert a date format of "YYYY-MM-DDThh:mm:ss+00:00"
+def assert_isoformat_utc(time_str):
+    try:
+        test_time = datetime.strptime(time_str, "%Y-%m-%dT%H:%M:%S%z")
+    except ValueError:
+        return False
+    # AR/SB it parses as an ISO string, now just check UTC timezone +00:00 and not +0000
+    return time_str.endswith("+00:00")
 
 
 # AR created file names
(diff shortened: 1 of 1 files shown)


sbailey:
Thanks @araichoor .

@schlafly currently this defaults to using your $DESI_ROOT/users/schlafly/tiling/tiles-4112-packing-20210329-decorated.fits .  Could we upgrade that to some place official (e.g. in surveyops)?  I don't want to let purity get in the way of pragmatism, nor do I want to "pollute" surveyops with a long-named tile file that we expect to update soon anyway (?), but normally we wouldn't hardcode a specific user's working directory as a default path. Any recommendations?

araichoor:
related to this tiles file: I notice that is also doesn t exactly follow the same formatting as $DESIMODEL/data/footprint/desi-tiles.fits, noticely:
- IN_DESI is a boolean (integer in desi-tiles.fits)
- no OBSCONDITIONS columns
I don t know if that matters.

djschlegel:
Let's include the nudges to 0.6% of the tile centers to avoid bright stars on the GFAs, as computed by Dustin:
/global/homes/d/dstn/desi-misc/tiles-4112-packing-20210329-decorated-brightest-nudged.fits

The RA,DEC coordinates should be offset by NUDGE_RA/cos(DEC), NUDGE_DEC.  I've confirmed that even after these offsets, all tiles are still strictly non-overlapping within a pass except for a single case.  There is a pair of tiles at 261.95901,75.865997 and 274.75554,75.415001 separated by 3.19 deg; all other tiles are separated by at least 3.30 deg.

Let's also still round to 3 significant digits after the decimal point, ie. turn +0.0001 to +0.001 and -0.0001 to -0.001.

schlafly:
I thought your final tile file was 20210405?

https://data.desi.lbl.gov/desi/users/schlafly/tiling/tiles-4112-packing-20210405-decorated.fits

djschlegel:
Ah right!  Let's ask @dstndstn to run his nudges on that tile file.  Too bad I just spent the time double-checking this one!

araichoor:
I apparently missed that 20210405 version, apologies for any confusion.
Remark: I do see a tiles-4112-packing-20210405-decorated-fixed.fits (latest timestamp) => maybe Dustin should start from this one?

schlafly:
*laugh*, yes, ugh, sorry, the fixed one is correct.
tiles-4112-packing-20210405-decorated-fixed.fits

araichoor:
one additional thought on this tiles file: I see it has dark=15 passes, bright=4 passes, backup=4 passes.
wouldn t we want to add few extra passes w.r.t. the nominal 4 passes in bright/backup, in case they are useful at some point?
I know for instance that Jaime has been exploring 5 passes for the bright program.
for dark, I guess we re good for 10+ years with 15 passes!

schlafly:
I would add that we intend to put an official file in surveyops soon.  That
file is even more different from the desimodel file---it's an ecsv file
missing many of these columns.  i.e., the current sv3 tile file:

I'd rather check in tiles-main there and point to that for this application.

Does that file format contain everything you need, Anand?

I agree we may tweak number of passes, etc., but that can be done by
updating surveyops.

I should also update desimodel accordingly at some point, but I don't know
what all the users of that are.  There will obviously be breakage getting
rid of the traditional fits file and columns and replacing it with this
kind of ecsv file.

>

araichoor:
yes, no need to add extra info from the fiberassign point-of-view.
only a minimal set of columns is required: TILEID, RA, DEC, OBSCONDITIONS, IN_DESI, PROGRAM; and I actually only use RA, DEC (and TILEID, if there; if not, I use CENTERID)
and in fba_launch, I actually re-define IN_DESI=1 or True and the OBSCONDITIONS.
my remark was more about potentially breaking existing code from e.g. surveysim but, as you say, that will happen anyway.

schlafly:
At least surveysim / desisurvey are codes that I spend a lot of time with, so those will handle the format changes gracefully.

I agree that from a surveysim / desisurvey / NTS / survey planning perspective, we want fba_launch to be able to figure out what targets based on PROGRAM and nothing else.  i.e., you are right to set OBSCONDITIONS on your own for fa purposes independently.

I don't have a sense for how fiberassign uses IN_DESI.  In survey planning, we are happy to redefine a tile to be inside / outside of the DESI footprint in order to prevent the NTS from observing it or to allow the NTS to observe it again.  That shouldn't influence how fiberassign treats it.  So I agree that you should be ignoring the value in the tile file.

araichoor:
perfect about surveysim/desisurvey!
about IN_DESI: there is, somewhere burried in fiberassign+desimodel, a rejection of the tile if IN_DESI=0; so I just hard-code IN_DESI=1 in fba_launch, in case a designed tile is outside the desi footprint.
for desi_fa_smallrun, there is no issue, as it s playing inside the desi footprint.

PR #337, opened 2021-05-03
Do STUCK positioners happen to land on good SKY positions for a given tile?

dstndstn:
This PR adds the 'stucksky.py' module, which does the positioner lookups in a set of "skybricks" tiles that summarize whether a chunk of sky is clear of sources.

### DIFF
--- a/py/fiberassign/stucksky.py
+++ b/py/fiberassign/stucksky.py
@@ -1,5 +1,3 @@
-
-
     
 def stuck_on_sky(hw, tiles):
     '''
@@ -8,18 +6,22 @@ def stuck_on_sky(hw, tiles):
     Returns a nested dict:
         stuck_sky[tileid][loc] = bool_good_sky
     '''
-    from scipy.spatial import KDTree
-    import numpy as np
-    import fitsio
     import os
-    from fiberassign.hardware import FIBER_STATE_STUCK
+    import numpy as np
+    from scipy.spatial import KDTree
+    import fitsio
     from astropy.wcs import WCS
+    from fiberassign.hardware import FIBER_STATE_STUCK
+    from fiberassign.utils import Logger
 
-    skybricks_dir = os.environ['SKYBRICKS_DIR']
+    skybricks_dir = os.environ.get('SKYBRICKS_DIR', None)
+    if skybricks_dir is None:
+        log = Logger.get()
+        log.warning('Environment variable SKYBRICKS_DIR is not set; not looking up whether '
+                    'stuck positioners land on good sky')
+        return
     skybricks_fn = os.path.join(skybricks_dir, 'skybricks-exist.fits')
     skybricks = fitsio.read(skybricks_fn, upper=True)
-    #for attr in ['ra1','ra2','dec1','dec2','brickname']:
-    #    setattr(skybricks, attr, skybricks[attr.upper()])
     skykd = _radec2kd(skybricks['RA'], skybricks['DEC'])
 
     tilekd = _radec2kd(tiles.ra, tiles.dec)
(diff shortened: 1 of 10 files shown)


py/fiberassign/stucksky.py
```diff
@@ -0,0 +1,151 @@
...
+    for tile_id, tile_ra, tile_dec, tile_theta, sky_inds in zip(
+            tiles.id, tiles.ra, tiles.dec, tiles.obstheta, sky_indices):
+        stuck_sky[tile_id] = dict()
+    
+        # Stuck locations and their angles
+        stuck_loc = [loc for loc in hw.locations
+                     if ((hw.state[loc] & FIBER_STATE_STUCK != 0) and
+                         (hw.state[loc] & FIBER_STATE_BROKEN == 0) and
+                         (hw.loc_device_type[loc] == 'POS'))]
+        stuck_theta = [hw.loc_theta_pos[x] for x in stuck_loc]
```
dstndstn, line 47: These (loc_theta_pos, loc_phi_pos) coords should get offset by hw.loc_{theta,phi}_offset.  We do that in a later PR.

dstndstn:
oops, I'll fix those test failures!!

sbailey:
Please put the core "are these (ra,dec) locations on sky or not" logic into desitarget so that both desispec and fiberassign can use it, independent of fiberassign hardware classes, focalplane x,y, theta,phi transforms, etc.  Or if you want to move forward with expediency, please at least carve that part of the logic out into an isolated function that we could later move to desitarget and then have fiberassign import it from there.

py/fiberassign/stucksky.py
```diff
@@ -0,0 +1,189 @@
...
+    xyz[:,0] = np.cos(rr) * np.cos(dd)
+    xyz[:,1] = np.sin(rr) * np.cos(dd)
+    xyz[:,2] = np.sin(dd)
+    return xyz
+            
+if __name__ == '__main__':
+    from fiberassign.tiles import load_tiles
+    from fiberassign.hardware import load_hardware
+
+    rundate = '2021-04-22T00:00:00'
```
tskisner, line 185: This was just for debugging the source code by importing the file directly, correct?

src/assign.cpp
```diff
@@ -59,6 +60,26 @@ fba::Assignment::Assignment(fba::Targets::pshr tgs,
...
+                for (auto & st : stuck_sky[tile_id]) {
+                    // st: < loc_id, bool >
+                    int32_t loc = st.first;
+                    bool good_sky = st.second;
+                    if (!good_sky)
+                        continue;
+                    int32_t petal = hw_->loc_petal[loc];
+                    nassign_tile .at(tp).at(tile_id)++;
+                    nassign_petal.at(tp).at(tile_id).at(petal)++;
+                    logmsg.str("");
```
tskisner, line 77: It may not matter here, but when printing debug messages from C++ inside nested loops, I found that string manipulation has a large performance impact.  That was the reason behind the pattern I'm sure you have seen elsewhere in the code:
```
bool extra_log = logger.extra_debug();

<snip>

if (extra_log) {
    logmsg.str("");

    <snip>

    logger.debug_tfg(...);
}
```
Which basically just checks a `bool` and cuts off the strstream manipulation.

tskisner (approved):
Some minor comments, but looks good!

py/fiberassign/stucksky.py
dstndstn: Yep; removed!

src/assign.cpp
dstndstn: Thanks, done in 2221129

py/fiberassign/stucksky.py
```diff
@@ -0,0 +1,189 @@
+import os
+import numpy as np
+
+class Skybricks(object):
+    '''
+    This class handles dynamic lookup of whether a given (RA,Dec) should make a good location for a sky fiber.
+    '''    
+    def __init__(self, skybricks_dir=None):
+        '''
+        Create a Skybricks object, reading metadata.
```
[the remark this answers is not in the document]
dstndstn: Thanks, will check when I send a PR to put this in desitarget.

PR #338, opened 2021-05-04
Add per-slitblock sky fiber limits

dstndstn:
Largely a mechanical change of adding per-slitblock counting in addition to the existing per-petal counting, used to enforce sky fiber spatial uniformity on the focal plane.

### DIFF
--- a/py/fiberassign/stucksky.py
+++ b/py/fiberassign/stucksky.py
@@ -14,9 +14,10 @@ def stuck_on_sky(hw, tiles):
     from fiberassign.hardware import FIBER_STATE_STUCK, FIBER_STATE_BROKEN
     from fiberassign.utils import Logger
 
+    log = Logger.get()
+
     skybricks_dir = os.environ.get('SKYBRICKS_DIR', None)
     if skybricks_dir is None:
-        log = Logger.get()
         log.warning('Environment variable SKYBRICKS_DIR is not set; not looking up whether '
                     'stuck positioners land on good sky')
         return
@@ -80,8 +81,6 @@ def stuck_on_sky(hw, tiles):
         loc_ra  = np.array([r for r,d in loc_radec])
         loc_dec = np.array([d for r,d in loc_radec])
 
-        print('Tile', tile_ra, tile_dec)
-        
         good_sky = np.zeros(len(loc_ra), bool)
         # Check possibly-overlapping skybricks.
         for i in sky_inds:
@@ -90,15 +89,14 @@ def stuck_on_sky(hw, tiles):
                 (loc_ra  <  skybricks['RA2'][i]) *
                 (loc_dec >= skybricks['DEC1'][i]) *
                 (loc_dec <  skybricks['DEC2'][i]))
-            print(len(loc_in), 'fibers overlap brick', skybricks['BRICKNAME'][i])
+            log.debug('%i fibers overlap sky brick %s' % (len(loc_in), skybricks['BRICKNAME'][i]))
             if len(loc_in) == 0:
                 continue
 
             fn = os.path.join(skybricks_dir,
                               'sky-%s.fits.gz' % skybricks['BRICKNAME'][i])
-            # FIXME -- debugging
             if not os.path.exists(fn):
-                print('Missing:', fn)
+                log.warning('Sky bricks: missing file %s' % fn)
                 continue
 
             skymap,hdr = fitsio.read(fn, header=True)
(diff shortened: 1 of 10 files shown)


py/fiberassign/stucksky.py
```diff
@@ -53,13 +53,15 @@ def lookup_tile(self, tilera, tiledec, tileradius,
         good_sky = np.zeros(len(ras), bool)
         # Check possibly-overlapping skybricks.
         for i in sky_inds:
-            # Do any of the query points overlap in the brick's RA,DEC unique-area bounding-box?
+            # Do any of the query points overlap in the brick's RA,DEC
```
dstndstn, line 56: Sorry about these changes, they're just merge debris from changes on multiple branches...

py/fiberassign/stucksky.py
```diff
@@ -116,8 +120,8 @@ def stuck_on_sky(hw, tiles):
 
         # Stuck locations and their angles
         stuck_loc = [loc for loc in hw.locations
-                     if ((hw.state[loc] & FIBER_STATE_STUCK != 0) and
-                         (hw.state[loc] & FIBER_STATE_BROKEN == 0) and
+                     if (((hw.state[loc] & FIBER_STATE_STUCK) != 0) and
```
dstndstn, line 123: (((belts) && ((braces))))

py/fiberassign/stucksky.py
```diff
@@ -116,8 +120,8 @@ def stuck_on_sky(hw, tiles):
...
+                         ((hw.state[loc] & FIBER_STATE_BROKEN) == 0) and
                          (hw.loc_device_type[loc] == 'POS'))]
```
dstndstn, line 125: (((belts) && ((braces))))

src/assign.cpp
```diff
@@ -53,34 +53,56 @@ fba::Assignment::Assignment(fba::Targets::pshr tgs,
...
             for (int32_t p = 0; p < hw_->npetal; ++p) {
                 nassign_petal[tp][tile_id][p] = 0;
             }
+            for (int32_t p = 0; p < hw_->npetal; ++p) {
+                nassign_slitblock[tp][tile_id][p].clear();
+                for (int32_t s = 0; s < hw_->nslitblock; ++s) {
+                    nassign_slitblock[tp][tile_id][p][s] = 0;
+                }
+            }
             tile_target_xy[tile_id].clear();
```
dstndstn, line 70: This change looks messier than it should -- I dropped a level of indentation by changing the "if (tp == TARGET_TYPE_SKY) { ... }" to a check-continue "if (tp != TARGET_TYPE_SKY) continue; ...".

The check to not count ETC fibers is new -- here, I'm doing that by checking their slitblock value, which happens to be -1; perhaps it would be more clear to check explicitly the `hw_->loc_device_type[loc]`.

(In addition, in the `stucksky.py` code, we now only check POS fibers anyway, so ETC fibers should not exist in the stuck_sky[tile_id] mapping; so the ETC check is just defensive coding.)

Really the only per-slitblock change is incrementing the `nassign_slitblock` at the bottom!

src/assign.cpp
```diff
@@ -257,7 +279,40 @@ bool fba::Assignment::petal_count_max(
 }
```
dstndstn, line 281: This basically exactly mirrors a pair of per-petal functions doing the same thing.

src/assign.cpp
tskisner: Thanks for the helpful explanation ;-)

src/assign.cpp
```diff
@@ -913,15 +1013,39 @@ void fba::Assignment::assign_force(uint8_t tgtype, int32_t required_per_petal,
...
     }
 
     gtmname.str("");
     gtmname << "force " << tgstr << ": total";
     gtm.stop(gtmname.str());
 
     logmsg.str("");
-    logmsg << "Force assignment of " << required_per_petal << " "
-        << tgstr << " targets per petal";
+    if (required_per_petal && required_per_slitblock) {
```
tskisner, line 1038: This logical check (and below) will return `true` even if the `required_per_*` values are -1.  Do you mean "`required_per_petal > 0`" and similar?

tskisner (approved):
Just one small comment, but looks good!

src/assign.cpp
dstndstn: Yes, thanks!!  fixed in 44e7cb9

PR #340, opened 2021-05-04
Apply theta-phi offsets when computing locations of stuck positioners

dstndstn:
The desimodel hardware parameters can contain some `theta, phi` values that are (usually only slightly) inconsistent with the `theta_min, theta_max, phi_min, phi_max`.  Previously, this would cause attempts to convert between theta,phi and x,y (esp. of stuck positioners, where the input `theta,phi` values are propagated) to fail, resulting in NaNs in output files, and false-positive collisions with neighbours.

We do a couple of things about this:

- first, we clip these values upon reading them from desimodel;
- second, we also add an `ignore_range` parameter to some of the `theta,phi -- x,y` conversion functions, that prevents it from failing if values are outside the min/max limits.

In addition, in a couple of places we have to add the theta,phi offsets when computing x,y positions.

I also factored out the `get_parked_thetaphi()` function, which computes which `theta,phi` angles we should request in order to "park" a positioner.  (In this instance, we're not trying to avoid collisions with neighbors, just placing the (most relevantly the phi angle) in a mostly retracted position.)

### DIFF
--- a/py/fiberassign/assign.py
+++ b/py/fiberassign/assign.py
@@ -336,8 +336,8 @@ def write_assignment_fits_tile(asgn, fulltarget, overwrite, params):
                     # position in the hardware model.  Used this fixed fiber location.
                     xy = hw.thetaphi_to_xy(
                         hw.loc_pos_curved_mm[loc],
-                        hw.loc_theta_pos[loc],
-                        hw.loc_phi_pos[loc],
+                        hw.loc_theta_pos[loc] + hw.loc_theta_offset[loc],
+                        hw.loc_phi_pos  [loc] + hw.loc_phi_offset  [loc],
                         hw.loc_theta_arm[loc],
                         hw.loc_phi_arm[loc],
                         hw.loc_theta_offset[loc],
@@ -354,7 +354,7 @@ def write_assignment_fits_tile(asgn, fulltarget, overwrite, params):
                     # This is an unassigned, working positioner.  Place it in a folded
                     # state.
                     thmin = hw.loc_theta_min[loc] + hw.loc_theta_offset[loc] + 1e-6
-                    phmax = hw.loc_phi_max[loc] + hw.loc_phi_offset[loc] - 1e-6
+                    phmax = hw.loc_phi_max  [loc] + hw.loc_phi_offset  [loc] - 1e-6
                     if phmax > np.pi:
                         phmax = np.pi
                     xy = hw.thetaphi_to_xy(
--- a/py/fiberassign/assign.py
+++ b/py/fiberassign/assign.py
@@ -353,8 +353,8 @@ def write_assignment_fits_tile(asgn, fulltarget, overwrite, params):
                 else:
                     # This is an unassigned, working positioner.  Place it in a folded
                     # state.
-                    thmin = hw.loc_theta_min[loc] + hw.loc_theta_offset[loc]
-                    phmax = hw.loc_phi_max[loc] + hw.loc_phi_offset[loc]
+                    thmin = hw.loc_theta_min[loc] + hw.loc_theta_offset[loc] + 1e-6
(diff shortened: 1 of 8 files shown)


py/fiberassign/assign.py
```diff
@@ -336,30 +336,33 @@ def write_assignment_fits_tile(asgn, fulltarget, overwrite, params):
                     # position in the hardware model.  Used this fixed fiber location.
```
dstndstn, line 336: This is where we're writing out the results, and we're computing the positions we're going to "request" for a stuck or broken positioner.  This is irrelevant for actually placing fibers, of course, since they can't move anyway, but it does affect the output tables (and plots/QA).

py/fiberassign/assign.py
```diff
@@ -336,30 +336,33 @@ def write_assignment_fits_tile(asgn, fulltarget, overwrite, params):
...
                         hw.loc_theta_min[loc],
                         hw.loc_phi_min[loc],
                         hw.loc_theta_max[loc],
-                        hw.loc_phi_max[loc]
+                        hw.loc_phi_max[loc],
+                        ignore_range=True
                     )
                     empty_x[iloc] = xy[0]
                     empty_y[iloc] = xy[1]
                 else:
```
dstndstn, line 353: Parking an unassigned positioner.  Here the theta/phi offsets were already being applied.

src/assign.cpp
```diff
@@ -1355,8 +1355,8 @@ bool fba::Assignment::ok_to_assign (fba::Hardware const * hw, int32_t tile,
                 loc,
                 tpos,
                 nb,
```
dstndstn, line 1357: This is the most important of the theta/phi offset changes.  This is where we're checking whether a neighbor will collide with a stuck/broken positioner.

src/hardware.cpp
```diff
@@ -131,6 +131,67 @@ fba::Hardware::Hardware(std::string const & timestr,
         loc_phi_max[location[i]] = phi_max[i] * M_PI / 180.0;
         loc_phi_pos[location[i]] = phi_pos[i] * M_PI / 180.0;
         loc_phi_arm[location[i]] = phi_arm[i];
+
```
dstndstn, line 134: I really could combine these two blocks...

src/hardware.cpp
```diff
@@ -131,6 +131,67 @@ fba::Hardware::Hardware(std::string const & timestr,
...
+        if (state[location[i]] & (FIBER_STATE_STUCK | FIBER_STATE_BROKEN)) {
+            if ((loc_theta_pos[location[i]] < loc_theta_min[location[i]]) ||
+                (loc_theta_pos[location[i]] > loc_theta_max[location[i]])) {
+                logmsg.str("");
+                logmsg << "STUCK/BROKEN positioner (loc " << location[i]
+                       << ") theta value is outside of range: theta = "
+                       << loc_theta_pos[location[i]] << " vs range [" << loc_theta_min[location[i]]
+                       << ", " << loc_theta_max[location[i]] << "]; CLIPPING to valid range";
```
tskisner, line 142: We talked about this offline...  Basically this clipping could result in an output RA/DEC of the stuck positioners that did not match the corresponding POS_T / POS_P from the online DB (where those angles may slightly exceed the "targetable" range we use in fiberassign.

tskisner (approved):
Thanks for this work, and the useful offline discussion

py/fiberassign/assign.py
```diff
@@ -1575,6 +1580,24 @@ def merge_results(targetfiles, skyfiles, tiles, result_dir=".",
 
     return
 
+def get_parked_thetaphi(theta_offset, theta_min, theta_max,
+                        phi_offset,   phi_min,   phi_max):
+    # Here we're adding theta/phi offset, and in the
+    # outside_theta_phi_range() function, those offsets get subtracted
```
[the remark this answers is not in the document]
dstndstn: Done; thanks

PR #341, opened 2021-05-07
Skybricks support

sbailey:
Additional skybricks support (assigning stuck positioners to sky locations):
  * adds `etc/fiberassign.module` that defines `$SKYBRICKS_DIR=$DESI_ROOT/target/skybricks/v2` so that when the installed module is loaded (`module load fiberassign/x.y.z`) that variable will be set automatically.
  * updates `fba_launch` to require that `$SKYBRICKS_DIR` is defined.
    * The underlying `fiberassign.assign` code doesn't require that, but the `fba_launch` wrapper requires it to standardize that is how we want to run it for the main survey.
  * if a required environment variable isn't defined, exit with a non-zero return code
  * adds `$SKYBRICKS_DIR` to the dependencies tracked in the output headers

Unrelated changes coming along for the ride:
  * makes `bin/fba_launch` and `bin/fba_sv1` executable
  * changes `--sky_per_slitblock` default to 1 instead of 0 (in the spirit of making default values what we want to use for normal operations)

A test run is available at `/global/cfs/cdirs/desi/users/sjbailey/temp` using
```
source /global/cfs/cdirs/desi/software/desi_environment.sh 21.3
module swap fiberassign/skybricks
module swap desimodel/master
export DESI_SURVEYOPS=$DESI_ROOT/survey/ops/surveyops/trunk
fba_launch --outdir . --tileid 999999 --tilera 150 --tiledec 2 --survey sv3 --program DARK --dtver 0.57.0
```

@anand please take a look, thanks (and thanks for the example `fba_launch` command to run for testing)


sbailey:
Whoops, I meant to mention @araichoor (not @anand who is probably some other random github user, apologies)

araichoor:
I m not sure of how this if condition behaves:
https://github.com/desihub/fiberassign/blob/b9111ebceb6a9c4e1e64f2458541b411f9f06f2f/bin/fba_launch#L1069-L1073
given that the default is now 1 both in fba_launch and scripts/assign.py: wouldn t it be safer/simpler to remove that if condition, and put --sky_per_slitblock at the same level as the other elements of opts?

araichoor:
and an extra minor comment: maybe define args.sky_per_slitblock as a string in fba_launch, as it is passed as a string (as e.g. args.sky_per_petal)

araichoor:
and what about recording SKYBRICKS_DIR in the fiberassign-TILEILD.fits.gz header? (e.g. under keyword SKYBRICK)

sbailey:
Thanks @araichoor .  I agree about putting `--sky_per_slitblock` at the same level as the other elements of opts and made that change.  I left it as an int, so that if someone makes an error like `--sky_per_slitblock yes` it will get caught at the level of argparsing and not later.

I also added a `--log-stdout` option to *not* redirect the log to a file (useful for interactive debugging; default is still to log to a file).  In the process I also found that main() was moving the logfile to the final location while stdout was still redirected to the temp file, so I moved that code to only move the logfile after it was closed.

I tested that with and without logfile redirection, but I'll leave this open a bit longer if there are other comments or objections to that change.

PR #342, opened 2021-05-11
Fba launch io

araichoor:
This PR is a major refactoring of the fba_launch, with factoring out all the functions to a new utility script fiberassign.fba_launch_io.

Though, there is no change on the produced files, except for the following minor bugs corrected in the process:

- ToO original file column values were not propagated in fiberassign-TILEID.fits.gz; this led to zeros in ToO rows in fiberassign-TILEID.fits, noticely the SV3_*_TARGET; now this is corrected; this changes the reported stats in the QA plot and the log;
- it was always reported in the header that the proper-motion correction was applied, irrespective of args.pmcorr="y" or "n"; now it is reported only if arg.pmcorr=="y";

Further minor changes:
- addition of a FASCRIPT keyword in fiberassign-TILEID.fits.gz, to keep track of which script was used to generate the tile; it will be fba_launch for the main survey, but dedicated scripts for dedicated tiles.
- now using desitarget.geomask.match instead of customed defined unq_searchsorted
- replacing get_isaen by desitarget.gaiamatch.gaia_psflike, to enforce consistency (e.g., if AEN cuts for Gaia/EDR3)
- changing the colorbar range for fiber assignment in the QA plot.
- removing divide by zeros in flux to magnitude conversion.
- correctly formatting the sbprof allowed choices

I have checked I can reproduce the output products (modulo the small differences due to the aforementioned bug correction):
- a run from the desi master environment
- one already designed tile (TILEID=36).

### DIFF
--- a/bin/fba_launch
+++ b/bin/fba_launch
@@ -33,16 +33,25 @@ from argparse import ArgumentParser
 # AR goal effective time
 # AR https://desi.lbl.gov/svn/data/surveyops/trunk/ops/config-sv2.yaml
 # AR https://desi.lbl.gov/svn/data/surveyops/trunk/ops/config-sv3.yaml as of Apr. 5th
+# AR main: [desi-survey 2390] Minimum exposure time fraction and requested effective times
 goaltimes_all = {
+    "sv1": {"DARK": 1000.0, "BRIGHT": 150.0, "BACKUP": 30.0},
     "sv2": {"DARK": 1000.0, "BRIGHT": 150.0, "BACKUP": 30.0},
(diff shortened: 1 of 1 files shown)


araichoor:
I just pushed two commits, making fba_launch_io.create_gfa() a bit more general, with renaming it create_target_nomtl().
and totally unrelated (sorry): also allowing SV1 in make_qa.

araichoor:
and a commit to enable args.survey=main or sv1 in fba_launch.
I added "for free" sv1, as it ll be required for some dedicated tiles (e.g. those of today).
as there is no main-mtl file, I couldn t test it actually, but sv3 should be blind to that change.

sbailey (approved):

PR #344, opened 2021-05-11
Gaia-based FLUX_R for GFA_TARGETS extension

ameisner:
@araichoor This updates the GFA_TARGETS catalog FLUX_R column so that it contains a Gaia-based r-band flux rather than a Legacy Survey r-band flux. The reason for doing this is to provide the best possible magnitudes to the ETC for its transparency measurements. Legacy Survey FLUX_R was found to have problematic saturation for relatively bright guide stars in a way that could not be reliably flagged, so we are switching to a setup that uniformly uses a Gaia-based r-band flux for all GFA_TARGETS rows. In detail the Gaia to r-band transformation (created by Rongpu) was trained on Gaia DR2 and DECam r-band.

### DIFF
--- a/py/fiberassign/gfa.py
+++ b/py/fiberassign/gfa.py
@@ -30,29 +30,45 @@ def isolated(ra, dec, mindist=5.0):
     index, separation, dist3d = cx.match_to_catalog_sky(cx, nthneighbor=2)
     return separation.to(u.arcsec) > mindist*u.arcsec
 
-def gaia_synth_r_flux(tab):
+def gaia_synth_r_flux(tab, gaiadr="dr2"):
     '''
     Generate synthetic r band flux based on Gaia photometry
     Args:
         tab: table with columns GAIA_PHOT_G_MEAN_MAG, GAIA_PHOT_BP_MEAN_MAG,
              GAIA_PHOT_RP_MEAN_MAG
+        gaiadr: string, must be either "dr2" or "edr3" (default to "dr2")
+
     Returns:
         array of synthetic Gaia-based r-band fluxes
 
     This code has largely been borrowed from legacypipe/reference.py
-    Uses Rongpu Zhou's Gaia DR2 transformation from (G, BP-RP) to DECam r
+    Uses Rongpu Zhou's transformations from (G, BP-RP) to DECam r
     '''
 
     color = tab["GAIA_PHOT_BP_MEAN_MAG"] - tab["GAIA_PHOT_RP_MEAN_MAG"]
-    # no BP-RP color: use average color
-    color[np.logical_not(np.isfinite(color))] = 1.4
-    # clip to reasonable range for the polynomial fit
-    color = np.clip(color, -0.6, 4.1)
 
-    coeffs = [0.1139078673, -0.2868955307, 0.0013196434, 0.1029151074,
-              0.1196710702, -0.3729031390, 0.1859874242, 0.1370162451,
-              -0.1808580848, 0.0803219195, -0.0180218196, 0.0020584707,
-              -0.0000953486]
+    # the 1.0.0 targeting gfas files have a mixture of zero and NaN
+    # placeholder values for BP and RP
+    badcolor = np.logical_not(np.isfinite(color)) | (tab["GAIA_PHOT_BP_MEAN_MAG"] == 0) | (tab["GAIA_PHOT_RP_MEAN_MAG"] == 0)
+
+    # color clipping parameters to match transform training color range
+    # from Rongpu:
+    color_lo = {"dr2": -0.6, "edr3": -0.5}
+    color_hi = {"dr2": 4.1, "edr3": 4.5}
+
+    # clip to reasonable range for the polynomial fit
+    color = np.clip(color, color_lo[gaiadr], color_hi[gaiadr])
+
+    coeffs = {"dr2": [0.1139078673, -0.2868955307, 0.0013196434, 0.1029151074,
+                      0.1196710702, -0.3729031390, 0.1859874242, 0.1370162451,
+                     -0.1808580848, 0.0803219195, -0.0180218196, 0.0020584707,
+                     -0.0000953486],
+              "edr3": [0.1431278873, -0.2999797766, -0.0553379742, 0.1544273115,
+                       0.3068634689, -0.9499143903, 0.9769739362, -0.4926704528,
+                       0.1272539574, -0.0133178183, -0.0008153813, 0.0003094116,
+                      -0.0000198891]}
+
+    coeffs = coeffs[gaiadr]
 
     synth_r_mag = np.array(tab["GAIA_PHOT_G_MEAN_MAG"])
 
@@ -60,15 +76,21 @@ def gaia_synth_r_flux(tab):
             synth_r_mag += c * color**order
 
     synth_r_flux = np.power(10, (22.5-synth_r_mag)/2.5)
+
+    # -99 is the FLUX_R placeholder that GFA_TARGETS uses
+    synth_r_flux[badcolor] = -99.0
+
     return synth_r_flux
 
-def get_gfa_targets(tiles, gfafile,faintlim=99):
+def get_gfa_targets(tiles, gfafile, faintlim=99, gaiadr="dr2"):
      
     """Returns a list of tables of GFA targets on each tile
 
     Args:
         tiles: table with columns TILEID, RA, DEC; or Tiles object
         targets: table of targets with columsn RA, DEC
+        gaiadr: string, must be either "dr2" or "edr3" (default to "dr2")
+                MAY NOT BE FULLY IMPLEMENTED
 
     Returns:
         list of tables (one row per input tile) with the subset of targets
@@ -170,7 +192,7 @@ def get_gfa_targets(tiles, gfafile,faintlim=99):
         t["FOCUS_FLAG"] = flag
 
         # patch in Gaia-based synthetic r flux for use by ETC
-        t["FLUX_R"] = gaia_synth_r_flux(t)
+        t["FLUX_R"] = gaia_synth_r_flux(t, gaiadr=gaiadr)
 
         gfa_targets.append(t)
 
--- a/py/fiberassign/gfa.py
+++ b/py/fiberassign/gfa.py
@@ -30,6 +30,37 @@ def isolated(ra, dec, mindist=5.0):
     index, separation, dist3d = cx.match_to_catalog_sky(cx, nthneighbor=2)
     return separation.to(u.arcsec) > mindist*u.arcsec
 
+def gaia_synth_r_flux(tab):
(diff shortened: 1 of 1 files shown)


dkirkby:
This is great, thanks @ameisner!

julienguy:
Maybe we should change the 'PHOTSYS' value for those so we know it's not anymore true LS data? (it's currently 'N' or 'S' for northern or southern surveys, maybe change to 'GN' and 'GS')  ...

moustakas:
Let's not expand PHOTSYS beyond one character, please. How about 'T' for
'transformed' or 'B' for 'bright'?

djschlegel:
Or "G" for Gaia-derived.

geordie666:
`PHOTSYS=G` is already used by `desitarget` to mean "this is a Gaia-only target rather than being from the Legacy Surveys." I'm not sure whether that makes `"G"` a good choice for this purpose, or a bad choice.

Pro: These GFAs are now on the Gaia system for r-band.
Con: `"G"` means _from Gaia_, this would instead have the slightly different meaning of "transformed into the Gaia system."

ameisner:
I'm happy to do something like this, but there's no "PHOTSYS" column in either the GFA_TARGETS or PMGSTARS tables. The closest thing I see is "RELEASE" in GFA_TARGETS, but I assume that also applies to other non r-band quantities in those tables.

araichoor:
thanks @ameisner.

I gave a look, with testing that on a TILEID-gfa.fits file, ie what s going into fiberassign:
tmp

the transformation looks good for 16<gaia_g<18, perfect.
though few questions:

1) if no gaia color information, you replace by the mean color, ie 1.4; though, if I m correct, that can estimate a synth_r_mag which is off by 0.5 mag if real_color=3, for instance, right? will that synth_r_mag be used for any calibration ? if so don t we want to just discard those? I ve checked in the desitarget gfa 0.58.0 catalogs: on average, ~1% have color=nan (the "worst" hp pixel catalog has 16%)

2) how holds this transformation for gaia/edr3? is it possible to check? in that spirit, I think we want to pass a gaiadr argument to get_gfa_targets(); because I see that the AEN criterion is also there, which will required an update for that; I could take care of implementing this gaiadr argument.
 
3) what about gfa outside this gaia range? are they used for guiding? with looking at the code, I see that the default faintlim=99 is used; is it ok? because, for gaia_g < 16 or gaia_g > 18, the synth_r_mag can also be ~0.5 mag off (see plot above); maybe that is more for @geordie666.

minor detail: maybe refer Rongpu s email id in the comment ([desi-survey 2361])?

ameisner:
Based on the sample of ~4500 PMGSTARS tables I've been studying, the max value of MAG is exactly 18, so I think that PlateMaker must be making a MAG <= 18 cut.

ameisner:
For G < 16, I think the discrepancy in your plot is due to problems with the Legacy Survey photometry (saturation). This is the main reason why we want to move to Gaia-based FLUX_R, since we would expect Gaia mags to be just fine even when pushing much brighter than the onset of saturation in Legacy Surveys. The sense of the discrepancy shown by the blue points in your right hand plot matches this interpretation about  Legacy Surveys saturation.

ameisner:
I would be fine with just marking the cases with no BP-RP color available as having bad Gaia-based FLUX_R (looks like a -99 value is used for this in GFA_TARGETS). I don't have a strong opinion about this option compared to the existing procedure. I agree that if a Gaia source had a real color of BP-RP = 3 and we guessed 1.4, then that'd lead to a big error in the synthetic FLUX_R. The replacement with 1.4 comes directly from legacypipe (which admittedly may be operating in a limit of huge numbers of stars such that it can get away with not caring about big occasional outliers).

araichoor:
thanks @ameisner for the checks on gaig_g>18 gfas, and the other explanations.
maybe @rongpu could check if that conversion is still valid with gaia/edr3?

rongpu:
@araichoor @ameisner There will be a 1~2% offset if using the DR2 polynomial on EDR3:


So if we are going to use EDR3, please use the following coefficients (which are from fitting EDR3):
```
coeffs = [0.1431278873, -0.2999797766, -0.0553379742, 0.1544273115,
        0.3068634689, -0.9499143903, 0.9769739362, -0.4926704528,
        0.1272539574, -0.0133178183, -0.0008153813, 0.0003094116,
        -0.0000198891]
```
Replacing the values in https://github.com/desihub/fiberassign/blob/5dc094efda58a2deef2a0d2fab7a38971d1cdedc/py/fiberassign/gfa.py#L52


And please also change the clip range from (-0.6, 4.1) (in `color = np.clip(color, -0.6, 4.1)`) to (-0.5, 4.5) where the new polynomial is still well-behaved.

Here is the EDR3 residual with the new coefficients:


And the new transformation:

ameisner:
Thanks @rongpu!! I'll incorporate these new coefficients later tonight.

ameisner:
@geordie666 The placeholder value used for missing BP, RP magnitudes in the targeting "gfas" files appears to have a mixture of two different behaviors. Sometimes NaN values indicate missing BP, RP, but sometimes it's values of 0 instead. The map of HEALPix pixel centers with no NaN's in either BP or RP (red plus marks in the attached image) looks like sky regions outside of Legacy Surveys DR9. Note however that there are also 0-valued BP, RP placeholders sprinkled throughout the DR9 footprint as well (I assume due to rare coverage gaps within the DR9 footprint).

gfas_bp_rp-nans_zeros

Not sure if this is something you're interested in changing on the targeting side? For now, I'll just put in a workaround on my end.

ameisner:
I made a few updates based on this thread:

- no more guessing BP-RP = 1.4; such cases now get assigned to -99 for FLUX_R
- handle missing Gaia BP/RP whether the placeholder value is 0 or NaN
- new gaiadr keyword argument
- if gaiadr="edr3", uses Rongpu's new eDR3 transform and associated color range clipping limits. By default still uses Gaia DR2, since that's what we currently have from the targeting "gfas" files.

There's nothing more that I currently want to do for this pull request. @anand what do you think?

geordie666:
> @geordie666 The placeholder value used for missing BP, RP magnitudes in the targeting "gfas" files appears to have a mixture of two different behaviors. Sometimes NaN values indicate missing BP, RP, but sometimes it's values of 0 instead. The map of HEALPix pixel centers with no NaN's in either BP or RP (red plus marks in the attached image) looks like sky regions outside of Legacy Surveys DR9. Note however that there are also 0-valued BP, RP placeholders sprinkled throughout the DR9 footprint as well (I assume due to rare coverage gaps within the DR9 footprint).
> 
> gfas_bp_rp-nans_zeros
> 
> Not sure if this is something you're interested in changing on the targeting side? For now, I'll just put in a workaround on my end.

@ameisner: This is, I believe, because Gaia officially uses `NaN` for missing data, but the Legacy Surveys instead uses `0` and relies on `ivar=0` to indicate missing data. If you feel strongly about this, I could alter the behavior. But, it is true to the provenance of how quantities are inherited from the different surveys. Let me know if you feel strongly about this, otherwise I'll leave it as-is.

ameisner:
@geordie666 I already updated the code in this pull request to handle the current setup, so I don't think it's necessary to change the targeting "gfas" files. One comment though: in the "gfas" files, the 0 placeholders are actually in the regions that *don't* have Legacy Surveys coverage, whereas I would have expected that those would be the regions that preserve Gaia's NaN placeholder convention. In other words, I think what I'm seeing is effectively the opposite of what you described.

geordie666:
Interesting. I'll definitely look at that when I update the GFAs for EDR3.

araichoor:
sorry, back to this a bit late: the code look fine.
there is just this last point discussed at the beginning: is there a way to record that we changed the FLUX_R from ls to synth_mag? @geordie666 you may be the best one to know what columns we can touch/change.
though I guess it s not a blocking factor: with the fiberassign code version recording in the header, we ll just know that it has been done, and this is for *all* gfa.
so we re done?

geordie666:
I'm fine with no changes other than in the header, or with picking a one-letter value of `PHOTSYS` that isn't "G", "N" or "S".

araichoor:
the change will be in the fiberassign-TILEID.fits.gz header, fine?
about PHOTSYS: as @ameisner said, that column does not exist in the gfa targets catalogs.

geordie666:
I'm fine with the change being in the fiberassign file header. Alternatively, if `REF_CAT` is in the reduced GFAs table, you could make up a special `REF_CAT`. But, I think we're OK, here.

ameisner:
I'm fine with just having the fiberassign version in the headers i.e., no further changes. REF_CAT is there in GFA_TARGETS. In one recent example I see it as exclusively "G2".

I could append "T" or "t" to the end of REF_CAT (standing for "transformed", based on John's earlier suggestion)...does anyone particularly like/dislike that idea?

araichoor:
actually, I don t really know if such a change on REF_CAT can break anything downstream.
I suggest, as everyone is fine with the information being in the header with the fiberassign version, that we stick to that and make no further change, so that we re sure not break anything.
so all good for merging, I think.

ameisner:
Yeah, I did wonder whether changing REF_CAT could break anything with e.g., PlateMaker, so I agree. I won't make any more changes then! Thanks!

geordie666:
> @geordie666 I already updated the code in this pull request to handle the current setup, so I don't think it's necessary to change the targeting "gfas" files. One comment though: in the "gfas" files, the 0 placeholders are actually in the regions that _don't_ have Legacy Surveys coverage, whereas I would have expected that those would be the regions that preserve Gaia's NaN placeholder convention. In other words, I think what I'm seeing is effectively the opposite of what you described.

@ameisner: I finally had time to dig into this. For DR9, the Legacy Surveys preserves `nan` values from Gaia DR2 (meaning "in the LS and detected in Gaia DR2, but Gaia but didn't measure flux"), but also uses `0` (meaning "in the LS but wasn't in Gaia at all"). For example:

```
import fitsio
import numpy as np
tractorcat = fitsio.read("/global/cfs/cdirs/cosmo/data/legacysurvey/dr9/south/tractor/199/tractor-1994m050.fits")                                                                                          
set(tractorcat[np.isnan(tractorcat['gaia_phot_rp_mean_mag'])]['ref_cat'])                                                                                                                                           
Out[]: {'G2'}
set(tractorcat[tractorcat['gaia_phot_rp_mean_mag'] == 0]["ref_cat"])                                                                                                                                                
Out[]: {'  ', 'L3'}
```

When I parsed the entire Gaia database into csv files at NERSC, though, I used `0` to mean missing values instead of `nan`. Outside of the LS footprint, we switch over to my choice of `0` for missing values.

In any case, I can't easily fix this without reprocessing all of my Gaia csv files, and even then there'd be mixed values for entries derived from the LS. So, I'm going to punt on this and just let you handle both `0` and `nan` to mean "missing," given that you've already set that up.

PR #345, opened 2021-05-11
Dc3r2 gama

araichoor:
This PR introduces the first use of a fba_launch-like script for a dedicated program.
Along with that, we implement the possibility that FAFLAVOR is no more set by args.survey and args.program, but defined independently (FAFLAVOR will remain unchanged for sv3 and the main, but will be different for dedicated programs).

### DIFF
--- a/bin/fba_launch
+++ b/bin/fba_launch
@@ -94,8 +94,14 @@ def main():
             )
         )
 
-    # AR faflavor
-    faflavor = "{}{}".format(args.survey, args.program.lower())
+    # AR hdr_survey, hdr_prgrm, faflavor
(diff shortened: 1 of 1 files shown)


sbailey:
Since this is urgent for tonight's observing, let's go ahead and merge+tag, but in the future we should continue to streamline how these special tiles are generated.  e.g. bin/fba_launch and bin/fba_launch_dc3r2_gamma are ~90% identical and hundreds of lines long, making it difficult to tell what is actually different for this tile.  bin/fba_launch_dc3r2_gamma also includes copying hardcoded constants like `goaltimes_all` and `sbprofs_all` which could be moved into a shared module like the work started with PR #342 .

PR #347, opened 2021-05-13
Mtltime timestamp

araichoor:
This PR now defaults the args.mtltime fba_launch argument to the latest timestamp in the $DESI_SURVEYOPS/mtl-done-tiles.ecsv file for the considered args.program.
If no value is found, we default to the current time.

We currently add +1min to the MTL latest timestamp to account for current discrepancy between mtl-done-tiles.ecsv and the ledgers.
To be seen how to change that once the fix is done in desitarget

Besides, we did a black formatting of fba_launch.

### DIFF
--- a/py/fiberassign/fba_launch_io.py
+++ b/py/fiberassign/fba_launch_io.py
@@ -11,6 +11,7 @@ import subprocess
 import sys
 import tempfile
 import shutil
+import re
 
 # time
 from time import time
@@ -141,7 +142,10 @@ def get_program_latest_timestamp(
         if keep.sum() > 0:
             d = d[d["PROGRAM"] == program]
             # AR taking the latest timestamp
-            tm = "{}+00:00".format(np.unique(d["TIMESTAMP"])[-1])
+            tm = np.unique(d["TIMESTAMP"])[-1]
+            # AR does not end with +NN:MM timezone?
+            if re.search('\+\d{2}:\d{2}$', tm) is None:
+                tm = "{}+00:00".format(tm)
             tm = datetime.strptime(tm, "%Y-%m-%dT%H:%M:%S%z")
             # AR TBD: we currently add one minute; can be removed once
             # AR TBD  update is done on the desitarget side
--- a/py/fiberassign/fba_launch_io.py
+++ b/py/fiberassign/fba_launch_io.py
@@ -14,7 +14,7 @@ import shutil
 
 # time
 from time import time
-from datetime import datetime
+from datetime import datetime, timedelta
 
 #
 import numpy as np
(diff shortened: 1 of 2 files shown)


araichoor:
@geordie666 : Aside comment: with checking how the MTL ledgers reading cuts on the isodate, I see that it is now a "<" cut: https://github.com/desihub/desitarget/blob/b86ae8ebf61024e310e5d102723191d5f23d4007/py/desitarget/io.py#L2552-L2557
Would it be possible to make it a "<=" cut ?

py/fiberassign/fba_launch_io.py
```diff
@@ -103,6 +103,56 @@ def get_svn_version(svn_dir):
...
+    # AR no need to check the scnd-mtl-done-tiles.ecsv file,
+    # AR     as we restrict to a given program ([desi-survey 2434])
+    fn = os.path.join(os.getenv("DESI_SURVEYOPS"), "mtl", "mtl-done-tiles.ecsv")
+    if os.path.isfile(fn):
+        d = Table.read(fn)
+        keep = d["PROGRAM"] == program.upper()
+        if keep.sum() > 0:
+            d = d[d["PROGRAM"] == program]
+            # AR taking the latest timestamp
+            tm = "{}+00:00".format(np.unique(d["TIMESTAMP"])[-1])
```
sbailey, line 144: @geordie666 @araichoor will the mtl-done-tiles.ecsv timestamps always *not* have the +HH:MM timezone specifier, or should this check first before adding "+00:00".  Is it guaranteed that if timezone isn't specified, that it means UTC?

If checking, a regex suggestion:
```
import re
if re.search('\+\d{2}:\d{2}$', timestamp) is None:
    #- does not end in +NN:MM timezone
```

geordie666: The timestamps for the Main Survey will have the "+00:00" specifier, so a check like this will be required.

araichoor: good point, thanks; shall I put a commit with that change?

geordie666:
```Would it be possible to make it a "<=" cut ?```. It's not easy to switch this behavior and retain backwards-compatibility. I could add a kwarg specifying that the user wants <= rather than <. But, you could also just add a fraction of a second to the isodate that you pass.

**Comment by geordie666** [comment]
I'll add the "<=" kwarg when I add the LyA decisions for the full Main Survey MTL loop, though.

**Comment by araichoor** [comment]
thanks for adding a "<=" option in reading the mtl; that simplifies things on the fba_launch side (otherwise, I don t know if there corner case issues with adding a fraction of second to args.mtltime when reading the isodate0.

File: py/fiberassign/fba_launch_io.py (hunk 1 above)
Comment (geordie666) [reply py/fiberassign/fba_launch_io.py]: Yep, I think @sbailey's suggested check just before line 144 should work.

Comment (araichoor) [reply py/fiberassign/fba_launch_io.py]: I just pushed the commit for that.

File: py/fiberassign/fba_launch_io.py [hunk 2]
```diff
@@ -103,6 +103,56 @@ def get_svn_version(svn_dir):
...
+
+    # AR check if the per-tile file is here
+    # AR no need to check the scnd-mtl-done-tiles.ecsv file,
+    # AR     as we restrict to a given program ([desi-survey 2434])
+    fn = os.path.join(os.getenv("DESI_SURVEYOPS"), "mtl", "mtl-done-tiles.ecsv")
+    if os.path.isfile(fn):
+        d = Table.read(fn)
+        keep = d["PROGRAM"] == program.upper()
+        if keep.sum() > 0:
+            d = d[d["PROGRAM"] == program]
```
geordie666, line 143: Don't you also need `program.upper()` here? Or just `d[keep]`?

araichoor: nice catch, corrected now; apologies for that, it has been a long, busy day... thanks!

geordie666:
I think this looks OK, now. I'm sure we'll find a few bugs as we work towards the new MTL loop, though!

araichoor:
I pushed a last commit, triggered by a calm re-reading of the function.
I think it should be fine now:
```
>>> from fiberassign.fba_launch_io import get_program_latest_timestamp
>>> get_program_latest_timestamp("DARK")
'2021-05-12T19:16:48+00:00'
>>> get_program_latest_timestamp("BRIGHT")
'2021-05-12T19:16:38+00:00'
>>> get_program_latest_timestamp("BACKUP")
'2021-05-02T18:44:45+00:00'
>>> get_program_latest_timestamp("bla") is None
True

```
should we merge?

PR #348, opened 2021-05-13
Use desimeter for RA,Dec <-> X,Y transformations

dstndstn:
These comments are mostly in the order they appear in the diff below:

- added to hardware.py: radec2xy, xy2radec, xy2cs5 , and funneled all calls through them; they call desimeter

- desimeter requires tile_obstime, tile_obstheta/fieldrot, and tile_obsha, so pass those around everywhere

- add the --ha command-line arg to set the design hour angle

- instead of creating a TargetTree and passing it to TargetsAvailable, we instead call the new `targets_in_tiles()` routine that produces maps from `tileid` to lists: `tile_targetids, tile_x, tile_y`; ie it pre-runs all the radec2xy transformations for each target that appears in each tile, and passes that big map into fiberassign.

- to support targets_in_tiles(), added TargetTree.near_data(), a tree-search routine like near(), except also returning the RA,Dec in addition to the targetid.

- adjust the `Assignment` constructor: it used to re-run a bunch of radec2xy calls; we instead save the x,y values in the TargetsAvailable structure

- tweak the KdTree code to add a `near_with_data` routine that returns the X,Y positions of nearby objects, not just their IDs.  I'm not totally happy with the way I did this (they're probably a mystical templates way of doing this); I'm basically carrying around two vectors of results, and filling one or the other based on a command-line flag, so that I can handle `near` and `near_with_data` in one piece of code.

- TargetsAvailable used to re-run the radec2xy transformations, but by passing in the targets_in_tiles maps, we have those xy values pre-computed.  We use the KdTree near_with_data here to avoid recomputing radec2xy again when finding the targets that are reachable by each positioner.

One thing I did not consider: the fiberassign radec2xy/xy2radec routines are openmp parallelized, while the desimeter ones are in python.  If this branch takes considerably longer to run, we will need to mitigate.  There should be options; it should trivially multiprocess over tiles.


araichoor:
Thanks!
A minor comment about your point: fba_launch designs a single tile at once, there is no more multi-tiles involved.
As the tiles now don t overlap anymore within a pass, if we need to design O(1000) tiles at once (e.g. for the holding pen), we make ntile calls to fba_launch.

sbailey:
after @dstndstn latest commit this looks good.  I also added a `fba_launch --ha HOURANGLE` option that gets passed through to the assignment hour-angle so that this PR is more useful as-is without having to wait for the next one.

PR #350, opened 2021-05-13
Allow setting additional buffers on positioner keep-out polygons

dstndstn:
- add margin-{pos,petal,gfa} command-line args
- (also add missing --ha arg to fba_launch)

- add `expand_closed_curve`, which does the geometry of expanding the polygon.  We apply this to the polygons immediately upon reading them from desimodel.

- this also adds the basically-unrelated inner keepout margin buffer.


sbailey:
tested, looks great.

PR #351, opened 2021-05-14
Add PLATE_{RA,DEC} to fiberassign outputs

dstndstn:
This just puts in the plumbing to haul PLATE_{RA,DEC} around (into and out of the C++ code) and plug them into the outputs.

We also currently carry PLATE_REF_EPOCH around until nearly the end; should reconsider that if we're never going to output it.


sbailey:
This looks fine, though in some places the EPOCH columns/variables are commented out (good, for the record of what we would do to include it), while in other places it is allocated but not used (bad, wasting memory).  Let's just comment it out everywhere (but not fully delete it), and allocate/propagate only the PLATE_RA / PLATE_DEC info.

dstndstn:
You know, thinking on it more, maybe we should reconsider the overall design.

In this version, I passed them into the C++ along with the TARGET_{RA,DEC}, and they get hauled around through the assignment and returned to the Python.  However, I *think* the following are our requirements and goals:

- after *whoever* computes the PLATE_{RA,DEC}, we want to use those coordinates directly in place of TARGET_{RA,DEC}, throughout fiberassign, for computing X,Y positions, collisions, all that.
- we *need* to be able to write out both the original PLATE_{RA,DEC} and TARGET_{RA,DEC} to our output files

We could, therefore, create and hold on to a TARGETID to TARGET_{RA,DEC} map at the Python layer (we could also include any other additional columns we want in the output and do not need at the C++ layer), and refer to that mapping when preparing to write our outputs.  After all, it is the Python layer that is reading the FITS tables of targets and passing them to C++; there's no reason it couldn't also hold on to some of that information!

sbailey:
Good point.  For that matter, does the C++ side need *any* ra,dec values plumbed through it anymore, or does it only need the x,y values?

dstndstn:
superseded by #353

PR #352, opened 2021-05-14
handling no ToO file in create_too()

araichoor:
This PR handles in fba_launch and fba_launch_io.py the case were no ToO file is there.


sbailey:
Thanks.  I tested on several cases new and old with and without ToO and it works as expected.  Merging.

PR #353, opened 2021-05-14
Add PLATE_{RA,DEC}, second attempt

dstndstn:
This is a much more extensive refactor that avoids passing the majority of targeting data down to the C++ layer, instead stashing it in a new Python "TargetTagalong" class.

The TargetTree class gets replaced by the scipy KDTree for querying targets near tiles, and then the htmTree.h data structure is no longer required.

Also:
- add FA_M_{POS,PET,GFA} header values recording the --margin-pos, --margin-petal, and --margin-gfa command-line arguments; use these in the plotting routines.
- allow --margin_pos as well as --margin-pos command-line args, to ease round-tripping through the FAARGS header

I have verified that this produces exactly the same fiberassign table results as master, for tile randomly chosen example tile 001007.  (* The final HDU, showing reachable targets, is written in a different ordering, but has the same set of rows.  All other HDUs appear in the same order and have exactly the same contents.)


araichoor:
thanks a lot for all that work!

minor comments:
- (@djschlegel ) do we want to keep the possibility of applying proper motion correction to PLATE_{RA,DEC} ? if so may want to also add in fiberassign-TILEID.fits.gz PLATE_REF_EPOCH, so that we have a consistent set of information available in case we want to apply proper? at least in ext=1 and 3, where REF_EPOCH is present; that could avoid any confusion.

ideally, to have files as much as consistent with previous ones:
- is there no way to have the POTENTIAL_ASSIGNMENT in the same ordering as previously?
- if not too painful, would it be feasible to add the PLATE_{RA,DEC} columns at the *end*, so that the Nth column in previous files would be the same as in the new files?

and an extra-minor comment: as long as you fixed the "hyphen" vs. "underscore" in the arguments, could you also do that for --log-stdout ?

araichoor:
awesome, thanks Dustin for implementing those suggestions!
about the ordering: I guess it s fine as it is now, but I let @sbailey or @geordie666 comment, if needed.

sbailey:
Given that there have been a bunch of commits after the original post, could @araichoor or @dstndstn re-verify that the outputs are identical to current fiberassign except for POTENTIAL_ASSIGNMENT order?

I like the FIBERASSIGN HDU to be ordered by LOCATION (not TARGETID); let's keep that.

@dstndstn do you expect any memory impact from the tagalog class?

araichoor:
for info I ve checked for one tile design (--tileid 999999 --tilera 200 --tiledec 0 --survey main --program DARK --dtver 1.0.0).
with the plateradec2 branch code:
- I confirm that the only difference in the data is the TARGETID column of the POTENTIAL_ASSIGNMENTS extension;
- except for SKY_MONITOR, `PLATE_RA, PLATE_DEC, PLATE_REF_EPOCH`, if present, are not at the end; I think that is because two sets of columns are "pasted", e.g. for the FIBERASSIGN extension:
https://github.com/desihub/fiberassign/blob/1b475f58d6be2ea9ed717e7a43ee79ab1c323354/py/fiberassign/assign.py#L1443
https://github.com/desihub/fiberassign/blob/1b475f58d6be2ea9ed717e7a43ee79ab1c323354/py/fiberassign/assign.py#L1522-L1527
so `merged_fiberassign_req_columns` appears first, then and `minimal_target_columns`; and `PLATE_RA, PLATE_DEC` appear in `merged_fiberassign_req_columns`; 
- there are some differences in the header due to different modules listed and different columns, not a problem;
- among `PLATE_RA, PLATE_DEC, PLATE_REF_EPOCH`, the following columns are present:
```
FIBERASSIGN ['PLATE_RA', 'PLATE_DEC']
SKY_MONITOR ['PLATE_RA', 'PLATE_DEC']
GFA_TARGETS ['PLATE_RA', 'PLATE_DEC', 'PLATE_REF_EPOCH']
TARGETS ['PLATE_RA', 'PLATE_DEC']
POTENTIAL_ASSIGNMENTS []
```
ie except for the GFA_TARGETS extension, PLATE_REF_EPOCH is not present.
I don t know if that is what we want (maybe, I just don t know...)

sbailey:
It will be a hassle if we start producing fiberassign files that have the columns in a different order because then we can't trivially stack them.  Since this is supposed to be a "no-op" PR that isn't required for resuming operations, let's take the time to get the same columns in the same order as before before merging this.

This implies *not* including them in the SKY_MONITOR, GFA_TARGETS, nor TARGETS HDUs.  IIUC your post correct, @araichoor, those are extra columns that appeared in this PR and aren't present before? 

OK if the rows are in a different order.

sbailey:
Looks good.  I re-confirmed that:
* all HDUs have the same columns in the same order as before
* all HDUs have the same rows (data values) in the same order except POTENTIAL_ASSIGNMENTS which have the same rows in a different order (ok)
* all headers are the same except FA_RUN timestamp (ok), and new keys FA_M_GFA, FA_M_PET, FA_M_POS giving the extra keepout margins (good).

Merging now, thanks.

PR #354, opened 2021-05-14
Nomtl option

araichoor:
This PR adds a args.nomtl option in fba_launch, for the potential use in case of special tiles design.
To do so, we adopt a cleaner approach in fba_launch_io.py.

Before:
- custom_read_targets_in_tiles() -> read desitarget catalogs for no-MTL only
- create_targ_nomtl() -> for no-MTL, calls custom_read_targets_in_tiles()
- create_mtl() -> independent function, reading ledgers + inflating them + other things in common with create_targ_nomtl(), like pmcorr.
fba_launch was using create_targ_nomtl() and create_mtl()

Now:
- custom_read_targets_nomtl_in_tiles() -> cleaned version of custom_read_targets_in_tiles()
- custom_read_targets_mtl_in_tiles() -> to read + inflate the MTL ledgers 
- create_targ() -> allows to read with or without ledgers, calling one of the two functions above.
fba_launch now uses create_targ() only.

I tested that the fiberassign-TILEID.fits.gz. file was similar to the one generated with the master branch for:

MYDATE=2021-05-14T00:00:00+00:00

fba_launch --outdir $OUTDIR --tileid 999999 --tilera 200 --tiledec 0 --survey main --program DARK --dtver 1.0.0 --rundate $MYDATE --mtltime $MYDATE --pmtime_utc_str $MYDATE

fba_launch --outdir $OUTDIR --tileid 999999 --tilera 150 --tiledec 10 --survey main --program BRIGHT --dtver 1.0.0 --rundate $MYDATE --mtltime $MYDATE --pmtime_utc_str $MYDATE

I tried for BACKUP:
fba_launch --outdir $OUTDIR --tileid 999999 --tilera 180 --tiledec 15 --survey main --program BACKUP --dtver 1.0.0 --rundate $MYDATE --mtltime $MYDATE --pmtime_utc_str $MYDATE
but it cannot run (master or branch), as there is no MTL BACKUP files yet.

However, that is a good test case: adding the --nomtl option runs smoothly:
fba_launch --outdir $OUTDIR --tileid 999999 --tilera 180 --tiledec 15 --survey main --program BACKUP --dtver 1.0.0 --rundate $MYDATE --mtltime $MYDATE --pmtime_utc_str $MYDATE --nomtl.


sbailey:
For the record, summarizing from Slack, these produce difference assignments:
```
MYDATE=2021-05-14T00:00:00+00:00
fba_launch --outdir ./withmtl --tileid 999999 --tilera 200 --tiledec 0 --survey main --program DARK --dtver 1.0.0 --rundate $MYDATE --mtltime $MYDATE --pmtime_utc_str $MYDATE
fba_launch --outdir ./nomtl --tileid 999999 --tilera 200 --tiledec 0 --survey main --program DARK --dtver 1.0.0 --rundate $MYDATE --mtltime $MYDATE --pmtime_utc_str $MYDATE --nomtl
```
The input targets are the same TARGETIDs, but there are differences with NUMOBS(_MORE), PRIORITY, and SUBPRIORITY that are surprising but probably have a good explanation.  We're postponing this PR for tonight so that we can study the consequences of mtl vs. nomtl inputs more calmly before implementing the `--nomtl` option.

geordie666:
For the record:

- The MTL files always have a `zcat=0` pass through the MTL code. This can lead to slight but reproducible differences in the ledgers versus the target files.
- The bottom line, here, is that the target files are _not_ guaranteed to be identical to the zero-pass ledgers. The columns related to MTL can be updated when making the ledgers. 

I don't think the `--nomtl` option is advisable. We should always run off the ledgers. The ledgers are relatively quick to make for the target files, so should always exist to be used by fiberassign.

araichoor:
thanks for the precision.
I didn t have time to investigate where the discrepancy between using mtl or not come from.
as you precise, I strongly suspect it is just a feature, coming from supplementary infos in the ledgers, but that should not be a problem as, for the primary+secondary targets, of course, we will always run with mtl.

this function is a more a utility function, to query and write targets inside a tiles file footprint.

for instance, in a survey=main fba_launch call, it is used for the gfa, where there is no mtl.
I guess it can be useful for some special program that will happen in the future.

from a code point of view, it simplifies things, with only one function, create_targ() (instead of two functions currently, create_targ_nomtl() and create_mtl()); it avoids duplication of lines of code.

geordie666:
Yeah, I think it's fine to have the `--nomtl` option, if we're really careful not to use it for the targets.

I looped through all of the target files and ledgers and checked that at least `SUBPRIORITY` was consistent. Discrepancies with that column would have surprised me most. For instance (this isn't fast, but I just ran it in the background while looking at other things):
```
from astropy.table import Table
import numpy as np
import fitsio
import healpy as hp
from desitarget.geomask import match
oc = "bright"
for i in range(hp.nside2npix(32)):
    try:
        a = Table.read("/global/cfs/cdirs/desi/target/catalogs/mtl/1.0.0/main/{}/mtl-{}-hp-{}.ecsv".format(oc, oc, i))
    except FileNotFoundError:
        pass
    pix = list(set(hp.ang2pix(8, a["RA"], a["DEC"], lonlat=True, nest=True)))[0]  
    b = fitsio.read("/global/cfs/cdirs/desi/target/catalogs/dr9/1.0.0/targets/main/resolve/{}/targets-{}-hp-{}.fits".format(oc, oc, pix))
    amatch, bmatch = match(a["TARGETID"], b["TARGETID"])
    aaa = a[amatch]
    bbb = b[bmatch]
    print(i, pix, len(a), len(b), len(aaa), np.all(aaa["SUBPRIORITY"] == bbb["SUBPRIORITY"]))
    if np.all(aaa["SUBPRIORITY"] == bbb["SUBPRIORITY"]) == False:
        raise ValueError
```

araichoor:
@sbailey : what about this PR? shall we still try to work on it, then merge?

araichoor:
I had a look at the created TILEID-targ.fits catalogs with or without the --nomtl option.
there are differences in two columns, OBSCONDITIONS and SUBPRIORITY.
the one in SUBPRIORITY likely explains that the fiber assignment is different.

- SUBPRIORITY: as discovered elsewhere, the SUBPRIORITY is overwritten in the process; I believe that the reason for having different values for the two cases is that the targets are not read in the same order if they are read from the MTL ledgers (i.. without the --nomtl option) or from the desitarget fits catalogs (i.e. with the --nomtl option);

- OBSCONDITIONS (@geordie666): the values are different in the MTL ledgers and in the desitarget fits catalogs, see example below; though I think it does not change the fiber assignment.
```
>>> tid = 39627751832947897
>>> d = read_targets_in_box("/global/cfs/cdirs/desi/survey/ops/staging/mtl/main/dark", radecbox=[199,200,-2,1], quick=False, mtl=True)
>>> d["OBSCONDITIONS"][d["TARGETID"] == tid]
array([63], dtype=int32)
>>> d = read_targets_in_box("/global/cfs/cdirs/desi/target/catalogs/dr9/1.0.0/targets/main/resolve/dark", radecbox=[199,200,-2,1], quick=True)
>>> d["OBSCONDITIONS"][d["TARGETID"] == tid]
array([1])
```

geordie666:
@anand:

- The MTL files always have a `zcat=0` pass through the MTL code. This can lead to slight but reproducible differences in the ledgers versus the target files.
- The bottom line, here, is that the target files are _not_ guaranteed to be identical to the zero-pass ledgers. The columns related to MTL can be updated when making the ledgers. 

In this case, when making the target files, `OBSCONDITIONS` is restricted to the targets that are driving decisions on whether something should be written to `dark` or `bright` output files. But, all of the target bits are retained once the targets have been written, and MTL has traditionally been allowed to update `OBSCONDITIONS`. So, on the zeroth pass through MTL that is done before writing the ledgers, `OBSCONDITIONS` will be updated to match the allowed conditions for all bits that are set for a given target.

In particular, the secondary target bits allow the full suite of `OBSCONDITIONS`:

```
- [SCND_ANY,            62, "Any secondary bit is set",                          {obsconditions: DARK|GRAY|BRIGHT|BACKUP|TWILIGHT12|TWILIGHT18}]
```
So, once a secondary target has been matched to a primary and the target bits have been updated, subsequent runs through MTL (including the zeroth run) will update `OBSCONDITIONS` to match what is allowed for generic secondary targets. I'm guessing the case you posted is something that has `SCND_ANY` set.

In any case, I'm not sure this is a bug. Had we not move to the ledger-based system, `OBSCONDITIONS` would have been updated in this manner too. I could force `OBSCONDITIONS` to always agree between the ledgers and target files. But, I _think_ the mocks still "expect" MTL to be able to update `OBSCONDITIONS`. This is the relevant section of code that allows MTL to update `OBSCONDITIONS`:

https://github.com/desihub/desitarget/blob/9faa019ef6c3fac127356589d68288caeaf74c7d/py/desitarget/mtl.py#L454-L460

That code is likely a holdover from before the bright/dark split. But, I'm not sure we care?

araichoor:
thanks a lot for the detailed explanation; I think we re fine with the way MTL re-write OBSCONDITIONS.
I was just mentioning in case it was not under the radar.
I confirm that the specific object I highlighted has SCND_ANY set.

ps: @anand is another user :)

geordie666:
Oops! Apologies for the noise @anand. I'll try to remember @araichoor in the future.

sbailey:
Given that we don't need `--nomtl` for this week, I'm going to prioritize my attention on other PRs, but briefly:
* if we are going to implement `--nomtl`, then it seems we should put in the full work to make it exactly agree, e.g. perhaps even by generating an MTL on the fly under the hood if that's what it takes;
* otherwise, what is the use case where we would want to use `--nomtl`, knowing that it would produce a different answer than if we had run it through an MTL?  Maybe for "convenience" for a super-custom test tile (or LIGO ToO?) where we just want to minimize the number of steps to get a set of RA,DEC onto fibers?  Even then, maybe the answer is to write that super-custom set of targets into an MTL format instead of a target format.

FWIW, if we had realized 6-months ago that mtl vs. nomtl would produce different results, I would probably push harder on making these agree to avoid the technical debt on having to keep track of why they don't agree.

araichoor:
This PR has been abandonned, and superseeded by further commits.
Closing without merging.

PR #355, opened 2021-05-14
Add PLATE_RA/DEC at end if needed

sbailey:
This PR adds PLATE_RA=TARGET_RA and PLATE_DEC=TARGET_DEC to the output files while merging other target columns, if prior plumbing hadn't already added those columns (like WIP PR #353).  This at least establishes the output format, even if deeper changes are needed to actually use PLATE_RA,PLATE_DEC (which are intended for eventual chromatic offsets).


sbailey:
I have verified that this produces identical results to current master, except for the addition of the two columns.

PR #356, opened 2021-05-14
adding args.hdr_survey and args.hdr_faprgrm arguments, defaulting to …

araichoor:
…args.survey and args.program.lower()


sbailey:
tested that outputs are identical to current master, except for these keywords when requested.  merging without waiting for full test suite.

PR #365, opened 2021-05-25
removal of MWS_FAINT_* targets for MTLTIME<20210524, for backwards co…

araichoor:
…mpatibility

desitarget-1.0.0 catalogs, used for fiber assignment of the main survey untill 20210524, do not have MWS_FAINT_{BLUE,RED} targets included.
This PR is in the perspective of being able to reproduce what we ran with desitarget-1.0.0 with future desitarget catalogs, which will include MWS_FAINT_{BLUE,RED} targets.

When creating the target catalogs with fba_launch_io.create_mtl(), we manually remove any MWS_FAINT_{BLUE,RED} targets if MTLTIME < 20210524.

I think the code should run, though, there is currently no available target files to actually test that (with MTLTIME > 20210524 and with some MWS_FAINT_{BLUE,RED} targets).

### DIFF
--- a/py/fiberassign/fba_launch_io.py
+++ b/py/fiberassign/fba_launch_io.py
@@ -1096,6 +1096,24 @@ def create_mtl(
         )
         d = d[keep]
 
+    # AR mtl: removing by hand MWS_FAINT_* for main if mtltime < 20210524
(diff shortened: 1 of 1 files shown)


sbailey:
After more discussion, we'll handle this on the MTL side instead of the fiberassign side, so closing without merging.

PR #366, opened 2021-05-26
Subpriority

araichoor:
This PR takes advantage of the recently implemented subpriority=True/False in desitarget.io.write_skies and write_targets.
Setting subpriority=False in the calls to those options disables the SUBPRIORITY over-writting that was happing before.

As the consequence, the desitarget catalogs/MTL ledgers original SUBPRIORITY values are preserved down to fiberassign-TILEID.fits.gz.


sbailey:
Looks good, thanks!  Merging now.

PR #367, opened 2021-05-27
add new Gaia-based variability bit to ETC_FLAG

ameisner:
@araichoor @dkirkby this is my attempt at implementing a new ETC_FLAG mask bit to label variable stars based on Gaia. The new bit within ETC_FLAG is 2^5.


sbailey:
@araichoor iterated with @ameisner via email/slack and signed off on this.  Merging now.

dkirkby:
@sbailey @ameisner I don't see any documentation of the new bit in this PR.  Actually, I can't find documentation for the existing bits either ;-)  Can I safely assume that the ETC should only use guide stars with ETC_FLAG==0?

dkirkby:
I'm not sure either - perhaps in the datamodel @sbailey?

PR #369, opened 2021-05-27
Gaia AEN criterion with desitarget.gaiamatch.gaia_psflike()

araichoor:
This PR updates the Gaia AEN criterion to handle EDR3.
Instead of using the hard-coded relation (valid for DR2), we now use desitarget.gaiamatch.gaia_psflike; and the dr, DR2 or EDR3, release is decided from the REF_CAT column (one dr only for the catalog).

### DIFF
--- a/py/fiberassign/gfa.py
+++ b/py/fiberassign/gfa.py
@@ -11,6 +11,7 @@ import numpy as np
 import fitsio
 from astropy.table import Table
 import desimodel.focalplane.gfa
+from desitarget.gaiamatch import gaia_psflike
 
 from ._internal import Tiles
 from .utils import Logger, Timer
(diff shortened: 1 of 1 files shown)


py/fiberassign/gfa.py
```diff
@@ -175,10 +176,8 @@ def get_gfa_targets(tiles, gfafile, faintlim=99, gaiadr="dr2"):
         # AR not passing the Gaia AEN criterion (PM correction done for AEN targets only)
         g = t["GAIA_PHOT_G_MEAN_MAG"]
         aen = t["GAIA_ASTROMETRIC_EXCESS_NOISE"]
-        isaen = np.logical_or(
-            (g <= 19.0) * (aen < 10.0 ** 0.5),
-            (g >= 19.0) * (aen < 10.0 ** (0.5 + 0.2 * (g - 19.0))),
-        )
+        gaiadr = "dr2" if "G2" in t["REF_CAT"] else "edr3"
```
geordie666, line 179: Provided this is only ever used for a full `gfas-*` file, it will work. There are possibly strange cases where somebody passes a single object and it happens to be a Tycho (`T2`) object, but you might not want to bother catching those cases? It's up to you!

geordie666 (approved):
I think this is fine. There's a possible corner-case you might want to catch where somebody passes a single object from an earlier GFAs file and it happens to be one of the rare Tycho objects, I think? It's probably not worth catching that case, though.

I'm happy for you to merge.

araichoor:
thanks!

PR #370, opened 2021-05-28
use np.nan_to_num() to avoid warnings: RuntimeWarning: invalid value encountered in greater

dstndstn:
This is a trivial change that avoids warnings, without changing behavior (since all NaN comparisons come out False, the first of these comparisons would have come out with keep=False even though NaN is not being checked explicitly).

### DIFF
--- a/py/fiberassign/fba_launch_io.py
+++ b/py/fiberassign/fba_launch_io.py
@@ -2035,7 +2035,8 @@ def flux2mag(flux, band=None, ebv=None):
     Notes:
         flux < 0 values are converted to NaN in magnitudes
     """
-    keep = flux > 0
+    # np.nan_to_num: NaN -> 0, so keep=False.
+    keep = (np.nan_to_num(flux) > 0)
     mag = np.nan + np.zeros(len(flux))
     mag[keep] = 22.5 - 2.5 * np.log10(flux[keep])
     if ebv is not None:
@@ -2137,7 +2138,8 @@ def qa_print_infos(
             ["GAIA_PHOT_G_MEAN_MAG)", "min(LS-R-FIBTOTMAG)"],
         ):
             magmin, color = "-", "k"
-            keep = (np.isfinite(mag)) & (mag > 0)
+            # np.nan_to_num: NaN -> 0, so keep=False.
+            keep = (np.nan_to_num(mag) > 0)
             if keep.sum() > 0:
                 magmin = mag[keep].min()
                 if magmin < magthresh:
--- a/py/fiberassign/fba_launch_io.py
+++ b/py/fiberassign/fba_launch_io.py
@@ -2658,13 +2658,13 @@ def plot_colcol_tracer(
             for axis_name, bands in zip(["x", "y"], [xbands, ybands]):
                 mag0 = flux2mag(
                     sample["FLUX_{}".format(bands[0])][sel],
-                    bands[0],
-                    sample["EBV"][sel],
+                    band=bands[0],
+                    ebv=sample["EBV"][sel],
                 )
                 mag1 = flux2mag(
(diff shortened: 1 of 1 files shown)


araichoor:
thanks @dstndstn for that careful check.
though, one question: I was previously checking for flux > 0; your change adds an additional check if flux = nan, right?

but when does that happen? could you provide a simple fba_launch call example?

because this flux2mag function is only applied:
- to ls-dr9 FLUX_* and FIBERTOTFLUX_R quantities, which should be finite values, right?
- for the QA, I create same handy dictionaries for the parent and assign samples (https://github.com/desihub/fiberassign/blob/c3338d91e5f06808cb76f270dc8248920e7f2313/py/fiberassign/fba_launch_io.py#L1812-L1962), with the same number of rows, with filling in the assign dictionary non-assigned rows with zeros: https://github.com/desihub/fiberassign/blob/c3338d91e5f06808cb76f270dc8248920e7f2313/py/fiberassign/fba_launch_io.py#L1883-L1887, so I expect to only have finite values there..

thanks.

dstndstn:
Hi Anand,

Running 
`python -u ../bin/fba_launch --outdir fast --tileid 999999 --tilera 200 --tiledec 0 --survey main --program DARK --dtver 1.0.0`
I see at the end of `fast/fiberassign-999999.log` these warnings:

```
/global/homes/d/dstn/fiberassign/py/fiberassign/fba_launch_io.py:2140: RuntimeWarning: invalid value encountered in greater
  keep = (np.isfinite(mag)) & (mag > 0)
/global/homes/d/dstn/fiberassign/py/fiberassign/fba_launch_io.py:2038: RuntimeWarning: invalid value encountered in greater
  keep = flux > 0
```

araichoor:
mmhh I don t get those warning when running that fba_launch call.
maybe you could point to me the location of the log file, so that I can have a look?

and to be clear : **I m totally fine with the PR,** it s just that I d like to understand how nan can happen there, to be sure it s not of a related bug somewhere.

ps: and thanks for the second commit!

dstndstn:
Sure, ~dstn/fiberassign/py/fast/fiberassign-999999.log

araichoor:
thanks. we re using exactly the same code versions.. this is a bit weird; but I guess it s fine.

py/fiberassign/fba_launch_io.py
```diff
@@ -2137,7 +2138,8 @@ def qa_print_infos(
             ["GAIA_PHOT_G_MEAN_MAG)", "min(LS-R-FIBTOTMAG)"],
         ):
             magmin, color = "-", "k"
-            keep = (np.isfinite(mag)) & (mag > 0)
+            # np.nan_to_num: NaN -> 0, so keep=False.
+            keep = (np.nan_to_num(mag) > 0)
```
sbailey, line 2142: If mag==np.inf, then the previous test would be False, while this new test would be True.  np.isfinite already handles NaN (returns False without complaining), so is this change needed?

dstndstn: Good catch.  np.nan_to_num does accept a value to set pos/neg inf to -- I suppose setting those both to zero would achieve what we want.

This PR was cosmetic, to suppress a warning coming from numpy, because when any of the mags are NaN, the comparison (mag>0) warns, even though we're then going to & it with False.

araichoor:
in this call, mag either comes from:
- GAIA_PHOT_G_MEAN_MAG => should always be a finite number
- flux2mag(d["FIBERTOTFLUX_R"]) => should be a finite number of np.nan
so I don t expect mag=np.inf to happen, in theory..

py/fiberassign/fba_launch_io.py
dstndstn: (I just pushed 0de4b93 to do this)

PR #371, opened 2021-05-28
Only interpret exclusion regions on demand

dstndstn:
The exclusions file can contain regions that are not actually used by the current set of positioners.  This PR adjusts the code so that only the exclusions that are used are interpreted by the code.  (They still get parsed from JSON to Python, but they don't get sent to the C++ code, nor do exclusion polygons get margins added.)

### DIFF
--- a/py/fiberassign/hardware.py
+++ b/py/fiberassign/hardware.py
@@ -232,27 +232,6 @@ def load_hardware(focalplane=None, rundate=None,
     if 'petal' in add_margins:
         margins['petal'] = -add_margins['petal']
 
-
-    for nm, shp in exclude.items():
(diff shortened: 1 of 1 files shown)


tskisner:
@dstndstn , some background information.  Originally there were only a handful of exclusion polygons (many positioners used the same ones), and so we kept around the old ones in the files.  If the focalplane updates from the online DB give us many new exclusions frequently (while keeping the other calibration information the same), then a new focalplane model would not be triggered and these exclusions would keep accumulating.  I don't know if this is happening, but I agree it is fine to filter out the unused ones here.

sbailey:
Thanks.  I confirm the produces the same outputs.  Merging.

PR #374, opened 2021-06-02
For GFA targets, set TARGETID = REFID if TARGETID = -1

dstndstn:
This means they then have unique TARGETIDs, which is handy.

### DIFF
--- a/bin/fba_launch
+++ b/bin/fba_launch
@@ -200,6 +200,7 @@ def main():
             step="gfa",
             start=start,
             add_plate_cols=False,
+            refid_to_targetid=True,
         )
     else:
         log.info("")
--- a/py/fiberassign/fba_launch_io.py
+++ b/py/fiberassign/fba_launch_io.py
@@ -963,6 +963,7 @@ def create_targ_nomtl(
     log=Logger.get(),
     step="",
     start=time(),
+    refid_to_targetid=False,
 ):
     """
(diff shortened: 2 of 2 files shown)


geordie666:
In case it's useful for future reference: Because of the way Gaia sets `SOURCEID`s for DR2/EDR3, the only (nested) `nside=8` HEALPixel that can possibly contain `REF_ID`s that resemble `TARGETID`s is number 4 (i.e. in the file `gfas-hp-4.fits`).

I looped through all of the GFA files and checked that no `TARGETID` in any HEALPixel matches a `REF_ID` from pixel 4. But, I thought that knowing that HEALPixel 4 was the crucial reference for any problematic `SOURCEID`s might save somebody some time in the future. Conveniently, HEALPixel 4 is (just) fully outside of the Legacy Surveys footprint.

sbailey:
Documenting an off-github (slack?) conversation about why this isn't merged: since this seems optional and isn't supposed to make a difference in the end, I suggest not making the change now just to minimize the number of moving parts.  I suggest closing the PR and it remains in the github record if we want to revive this idea later.  Otherwise @dstndstn please re-explain the impact and why this should remain open and/or be merged.

dstndstn:
I think this is probably not the way we want to handle this issue ultimately.
Let's not do anything until after our meeting today.

dstndstn:
do #375 instead

PR #377, opened 2021-06-08
Add script to compare two FITS files containing binary tables.

dstndstn:
This adds bin/fits-compare, a generic script to compare two FITS files containing multiple HDUs of tables.

Also allows md5sum checks of the data sections of files.

### DIFF
--- a/bin/fits-compare
+++ b/bin/fits-compare
@@ -1,0 +1,174 @@
+#! /usr/bin/env python
+
+import os
+import tempfile
+import numpy as np
+
+def main():
+    from argparse import ArgumentParser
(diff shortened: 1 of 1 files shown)


sbailey:
How does the functionality of this differ from fitsdiff, which is included in our standard environment as part of what astropy provides?

dstndstn:
Heh, I had forgotten about fitsdiff, if I ever knew about it.  It's, uh, pretty similar.  This script has md5sum checking of the whole data unit as an option, and it can print the RMS difference of a column that differs.

dstndstn:
byeeeee

Issue #389, opened 2021-09-01
Non-reproducibility if focal plane model files not svn-updated (RUNDATE)


araichoor:
In current (e.g. tag 5.1.1) fiberassign code, the default RUNDATE is set to datetime.now():
https://github.com/desihub/fiberassign/blob/802de4c853c29d104a676d925b74bda177989ab6/bin/fba_launch#L651-L652

However, this leads to non-reproducibility issue if the on-disk focal plane model files are not svn-updated when the fiberassign is run to design the tile.

That did happen for some SV3 tiles (https://desi.lbl.gov/trac/wiki/SurveyOps/OnePercent/fiberassign#Settingstore-runtheintermediateproducts).

A work-around would be to instead default RUNDATE to the latest TIME value in the latest, on-disk desi-state_*.ecsv file.

araichoor:
This is addressed in the merged https://github.com/desihub/fiberassign/pull/390.

PR #390, opened 2021-09-01
Set default RUNDATE to the latest TIME in the desi-state*.ecsv files

araichoor:
This PR addresses https://github.com/desihub/fiberassign/issues/389.

We now set the default RUNDATE value to the latest TIME in the latest, on-disk desi-state_*.ecsv file, instead of using the time when the tile is designed.


schlafly:
Looks good to me.

Issue #393, opened 2021-09-01
Reproducibility in reading the ToO.ecsv ledger


araichoor:
The original pre-shutdown 2020 Main ToO.ecsv file (and SV3) do not have a TIMESTAMP column.

That is potentially problematic to rerun fiber assignment, as the ToO.ecsv file will evolve.
See https://github.com/desihub/desitarget/pull/761.

araichoor:
ah right, sorry I forgot about that TILE/FIBER type (even though you reminded it to me recently...)
thanks a lot for bringing that up!

I should then apply this TIMESTAMP cut only if `too_tile=False` in this create_too() function, ie when designing Main-survey tiles:
https://github.com/desihub/fiberassign/blob/802de4c853c29d104a676d925b74bda177989ab6/py/fiberassign/fba_launch_io.py#L1184-L1186
right?
That should solve the issue with no action needed on your side?

And a clarification: I m a bit confused by your message: there is only one ToO.ecsv file, right (in $DESI_SURVEYOPS/mtl/main/ToO/)? that file will contain TOO_TYPE=TILE or FIBER targets (though for now, only TILE are there).

geordie666:
Correct, I think that if you only apply the `TIMESTAMP` for `TOO_TYPE==FIBER` targets, then everything will be OK.
```
And a clarification: I m a bit confused by your message: there is only one ToO.ecsv file, right (in $DESI_SURVEYOPS/mtl/main/ToO/)? that file will contain TOO_TYPE=TILE or FIBER targets (though for now, only TILE are there).
```
Everything you say in the above is correct (there is also a ToO ledger in `$DESI_SURVEYOPS/mtl/sv3/ToO/`, but that's irrelevant to this discussion.

araichoor:
This is addressed in the merged https://github.com/desihub/fiberassign/pull/394.

PR #394, opened 2021-09-01
MTLTIME argument in create_too()

araichoor:
This PR addresses https://github.com/desihub/fiberassign/issues/393.

It adds a mtltime argument to the `fba_launch_io.create_too()` function.

For potential backward-compatibility in the function calling sequence, we have made this argument optional, defaulting to None; that could be changed.


schlafly:
Looks good except for my comments in the other thread about not really understanding the dual input start/end mjds and the ToO file start/end mjds, which is likely just my misunderstanding!

Issue #395, opened 2021-09-02
ToO: reproducibility of the MJD window


araichoor:
Pinging @sbailey and @schlafly for advice.

For assignment, ToOs are selected based on a cut on `MJD_BEGIN`, `MJD_END`, with defining a desired MJD window.
In the current fba_launch code, we first define `mjd_now` based on `Time.now()`:
https://github.com/desihub/fiberassign/blob/feab4ed8b8ce8f0568c91b91213e698198c4a0ec/bin/fba_launch#L650
Then we define the MJD window (`mjd_min`, `mjd_max`) based on `mjd_now`:
https://github.com/desihub/fiberassign/blob/feab4ed8b8ce8f0568c91b91213e698198c4a0ec/bin/fba_launch#L266-L267.

Using `Time.now()` to define `mjd_now` prevents reproducibility, when re-running fiberassign at a later time.
This is addressed in the fba_rerun3 branch, where we switched to using `args.pmtime_utc_str` to define `mjdnow`:
https://github.com/desihub/fiberassign/blob/2027fbbdf913a1a37526fc3861edc2c826fa8f12/bin/fba_launch#L111-L115
taking advantage that, up to now, `args.pmtime_utc_str` tracks when the code was actually run.

However, in a soon-to-be-merged PR, we introduced a new keyword in the fiberassign-TILEID.fits.gz header, `NOWTIME`, arguing that it is not a safe choice to assume that `args.pmtime_utc_str` tracks when the code was run (https://github.com/desihub/fiberassign/pull/392).

I can imagine two approaches to have things more consistent:
- **approach_1**: we add to fba_launch a `args.nowtime` argument  (defaulting to Time.now()), and tie `mjd_now` to this `args.nowtime`; the minor drawback is that, doing this, `args.nowtime` will track when the code was **originally** run, i.e. for a re-run it will not track when the re-run is done; that seems ok to me;
- **approach_2**: we add to fba_launch two arguments, `args.mjd_min` and `args.mjd_max` (defaulting to `Time.now()-1`, `Time.now()+30)`, and propagate those in the fiberassign-TILEID.fits.gz header; and get rid of `mjd_now`.

Of course, I m open to other solutions.

schlafly:
Can we foresee a future when the ToO requests will be different from Time.now() - 1 and Time.now() + 30?  If so, let's do approach 2.  I hate adding more arguments, but the ToO stuff does sound rather different from just "now."

araichoor:
[...]

PR #396, opened 2021-09-02
ToO: MJD definition and cut

araichoor:
This PR addresses https://github.com/desihub/fiberassign/issues/395.

Previously:
- the `mjd_now` was defined as `datetime.now()`, i.e. when the fiberassign code is run to design the tile, causing a non-reproducibility issue for re-run of a tile;
- up to 2020 pre-shutdown (SV3, Main), the tiles were designed ahead of observations; hence we were defining a MJD window with  `mjd_min=mjd_now-1`, `mjd_max=mjd_now+30`, and selecting ToO targets with: `keep &= (d["MJD_BEGIN"] < mjd_max) & (d["MJD_END"] > mjd_min)`; with now moving to fiberassign on-the-fly, this -1, +30 is unnecessary.

We now use:
- `mjd_now = args.pmtime_utc_str`, i.e. the time used for proper-motion correction, solving the non-reproducibility issue;
- `mjd_min = mjd_max = mjd_now` in the call to `create_too()` in fba_launch, acting similarly to select ToO targets with: `keep &= (d["MJD_BEGIN"] < mjd_now) & (d["MJD_END"] > mjd_now)`, which the desired behavior for fiberassign on-the-fly.

### DIFF
--- a/bin/fba_launch
+++ b/bin/fba_launch
@@ -108,6 +108,13 @@ def main():
     # AR safe: argument dates correctly formatted?
     assert_arg_dates(args, log=log, step="settings", start=start)
 
+    # AR mjd_now: setting to pmtime_utc_str if not set
+    # AR    see https://github.com/desihub/fiberassign/issues/395
+    # AR    (and we just assessed args.pmtime_utc_str has the correct formatting)
+    # AR    (instead of utc_time_now_str, to allow reproducibility)
+    # AR    (purposely not tying to rundate)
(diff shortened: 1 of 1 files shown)


araichoor:
I ve purposely not modified the create_too() calling sequence, keeping `mjd_min` and `mjd_max`, for backward-compatibility.
Besides, that could be useful for special cases handling.

schlafly:
Looks good to me.

Issue #403, opened 2021-09-15
secondary: flux values not propagated


araichoor:
Since the beginning of operations, the fluxes for secondary targets are not propagated in the fiberassign-TILEID.fits.gz.
The initial reason is the following:
- fiberassign first reads the ledgers (which have a minimal set of columns);
- then it reads the desitarget *fits catalogs to add other columns;
- however, the desitarget inflate_ledger() (now match_to_ledger()) routine for adding those columns is not designed for stand-alone fits catalogs, as the secondary ones.

https://github.com/desihub/fiberassign/blob/7cdf3a03deeb2f0dc269f6ef972e194cc7952206/py/fiberassign/fba_launch_io.py#L1365-L1390

The secondary fits catalogs do not have many columns: the only columns, present in the FIBERASSIGN extension of fiberassign-TILEID.fits.gz and populated with zeros, are the flux ones: "FLUX_G", "FLUX_R", "FLUX_Z", "GAIA_PHOT_G_MEAN_MAG", "GAIA_PHOT_BP_MEAN_MAG", "GAIA_PHOT_RP_MEAN_MAG"

Remark for the 2021 pre-shutdown fiberassign-TILEID.fits.gz files have been patched (for another reason, see https://github.com/desihub/fiberassign/pull/375); that patch also fixes that issue for those ([desi-survey 2996]).

araichoor:
Fixed in PR https://github.com/desihub/fiberassign/pull/404.

PR #404, opened 2021-09-15
create_mtl(): add flux columns for secondaries

araichoor:
This PR addresses https://github.com/desihub/fiberassign/issues/403.

The secondary targets fits catalogs are ~230MB/800MB for BRIGHT/DARK:
`232M May 30 07:01 /global/cfs/cdirs/desi/target/catalogs/dr9/1.1.1/targets/main/secondary/bright/targets-bright-secondary.fits`
`823M May 30 07:01 /global/cfs/cdirs/desi/target/catalogs/dr9/1.1.1/targets/main/secondary/dark/targets-dark-secondary.fits`

@schlafly: from small tests on the NERSC logging node, it adds at most ~2s (for DARK) to a `fba_launch` call; I am not sure we can make it faster, unfortunately.

@geordie666: I assumed the following, hoping that is (and will remain!) correct:
- `targets-{bright,dark}-secondary.fits`: I assumed that the TARGETIDs are unique (e.g. no -1);
- BACKUP program: there are no secondaries.

Lastly, as I use `desitarget.mtl.match_ledger_to_targets()`, I proceed only if `desitarget.__version__ >= 1.2.2`, as the pre-1.2.2 `inflate_ledger()` does not work for the stand-alone secondary fits files; but that is ok, as this only concerns SV3.

### DIFF
--- a/py/fiberassign/fba_launch_io.py
+++ b/py/fiberassign/fba_launch_io.py
@@ -21,6 +21,7 @@ from datetime import datetime, timedelta, timezone
 #
 import numpy as np
 import fitsio
+import healpy as hp
 
 # astropy
 from astropy.io import fits
@@ -1296,11 +1297,6 @@ def create_mtl(
     Notes:
         if pmcorr="y", then pmtime_utc_str needs to be set; will trigger an error otherwise.
         for sv3-backup, we remove BACKUP_BRIGHT targets.
-        TBD : if secondary targets, we currently disable the inflate_ledger(), as it
-                seems to not currently work.
-                hence if secondary and pmcorr="y", the code will crash, as the
-                GAIA_ASTROMETRIC_EXCESS_NOISE column will be missing; though we do not
-                expect this configuration to happen, so it should be fine for now.
 
         TBD: the PLATE_{RA,DEC,REF_EPOCH} columns currently simply are copy of RA,DEC,REF_EPOCH
         TBD:    but it prepares e.g. to add chromatic offsets.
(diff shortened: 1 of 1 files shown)


araichoor:
And a last remark: I am not sure how the code behaves if there are no secondaries in the requested tile.
I guess it will crash; actually, with testing a tile at ra=0, dec=-80, it even crashes at the primary target design...

However, that should never happen for the main survey: I visually checked that the secondary targets in BRIGHT/DARK cover the full main footprint:
tmp-BRIGHT
tmp-DARK

py/fiberassign/fba_launch_io.py
```diff
@@ -1388,6 +1386,42 @@ def create_mtl(
                 targdir, tiles=tiles, quick=True, columns=columns + ["TARGETID"]
             )
             d = match_ledger_to_targets(d, targ)
+    # AR mtl: case secondary
+    else:
+        # AR mtl: secondary for backup should never happen, but safe approach still
+        if ("backup" not in mtldir) & (desitarget.__version__ >= "1.2.2"):
```
schlafly, line 1392: I assume the version check is for backward compatibility?  That makes sense.  This looks like a plain alphabetical string comparison test, though, which I don't think is what you want.  You might want something like:
https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python
though I don't think that will handle cases like __version__ == 'master' or something.  Is there a test we can do on a different string to get the behavior you want?

araichoor: thanks for noticing that.

I have used that kind of pure-string comparison for fiberassign and desitarget versions in lots of place in fba_launch_io.py, fba_rerun_io.py, and fba_rerun...
that is working from the tests I have done (even with master); but I acknowledge it will break if we reach one day versions 10.0.0 (hopefully that shouldn t happen!).

I could use this packaging.version.parse() with a test for the "master" case; I can investigate if there are other solutions.
pinging @sbailey in case he is aware of other options.

geordie666:
One comment:

New secondary targets added during the summer shutdown will appear in:

```/global/cfs/cdirs/desi/target/catalogs/dr9/*/targets/main2/secondary/bright/targets-bright-secondary.fits```
```/global/cfs/cdirs/desi/target/catalogs/dr9/*/targets/main2/secondary/dark/targets-dark-secondary.fits```

(i.e. main**2** instead of main). So, you'll also need to check those files, once I've added those targets.

araichoor:
good to know!
I ll add a query to those main2 files when cori is back up (I forgot about today s maintenance...), hoping it won t add another 2s to fba_launch..

PR #406, opened 2021-09-17
Fbarerun bugfix

araichoor:
This PR has two small modifications to bin/fba_rerun:
- fix a bug (discovered when --worldreadable is turned on in the original fiberassign file);
- disable any provided --svntiledir from the original fiberassign file, as we do not want to copy rerun files to svn.


araichoor:
@ashleyjross : I think there should be no problem, but in case you want to make few tests with https://github.com/desihub/LSS/blob/master/Sandbox/validate_lastnight_fa.py, please do!
I plan to merge in something like one hour.

Issue #413, opened 2021-09-28
Set EBV for all fibers


sbailey:
Fiberassign is currently using EBV=0.0 for some targets, apparently including all OBJTYPE='SKY', 'BAD', and '' targets, but also some science targets (OBJTYPE='TGT').  I'm guessing that this is just propagated from input targeting that may be missing EBV columns or have defaults of 0.0, but it would be helpful to catch these cases and use `desiutil.dust.SFDMap` to lookup E(B-V) values.  

This impacts downsteam code that uses the EBV values to interpret extinction for TSNR2 calculations, resulting mis-interpreting these targets as unextincted and thus overexposed.  I'll also file a ticket to catch these on the desispec side, but it would be preferable to have fiberassign output with correct EBV values from the start.

e.g. tile 21311, which in addition to normal SKY and BAD fibers, has a patch of bad imaging:

![image](https://user-images.githubusercontent.com/218471/135024459-b564b744-ae81-46ce-ac2e-ebe4ddc47799.png)

djschlegel:
This happens for all flavors of sky targets (SKY,BAD_SKY,SUPP_SKY) as well as targets that are secondary targets with "OVERRIDE" set to not take the DR9 targeting information.  This should not occur for anything else.

araichoor:
**Foreword:**
The EBV column is not in the columns affected by the patched listed by @dstndstn in [desi-survey 2962].

**TGTs with EBV=0 in fiberassign-TILEID.fits.gz files:**
A complementary way to say this: those are either stand-alone secondaries or ToOs.
Because EBV, which is not in the ledgers, neither is in the static desitarget catalogs for secondaries (and there is no static catalogs for ToOs); so that EBV information just does not exist in desitarget catalogs for those targets.

Below a code snippet assessing that TGTs with EBV=0 are either stand-alone secondaries or ToOs.

```
# gather TARGETIDs from stand-alone secondaries and ToOs
>>> fns = ["/global/cfs/cdirs/desi/survey/ops/surveyops/trunk/mtl/main/ToO/ToO.ecsv"]
>>> fns += sorted(glob("/global/cfs/cdirs/desi/target/catalogs/dr9/1.1.1/targets/*/secondary/*/*targets-*-secondary.fits"))
>>> tids = []
>>> for fn in fns:
>>>     tids += list(Table.read(fn)["TARGETID"])
# main fiberassign files:
[...]

PR #414, opened 2021-09-28
Populate EBV=0 rows in FIBERASSIGN with correct values

araichoor:
This PR addresses https://github.com/desihub/fiberassign/issues/413.
Currently, `EBV=0` in the FIBERASSIGN extension of `fiberassign-TILEID.fits.gz` files for: sky fibers, stand-alone secondary fibers ToO fibers, unassigned fibers.
This complicates the downstream analysis.
Typically, there will be ~800 fibers with `EBV=0` (~700 skys+unassigned, ~100 secondaries); though could be more for tiles with more passes, as more secondaries will be assigned.

We now populate in `assign.merge_results_tile()` the `EBV=0` rows with correct values.
This queries the EBV maps in `$DUST_DIR`.
I ve thus also added `$DUST_DIR` in the list of environmental variables to be defined.

On the NERSC logging node, this takes ~1s.
@schlafly: could you test that at KPNO? hoping that `$DUST_DIR` is already defined, and that the dust maps are there.

Lastly: this adds an astropy dependence in assign.py, I don t know if that matters...

### DIFF
--- a/py/fiberassign/assign.py
+++ b/py/fiberassign/assign.py
@@ -15,6 +15,9 @@ import re
 
 import numpy as np
 
+from astropy import units
+from astropy.coordinates import SkyCoord
+
 import multiprocessing as mp
 from multiprocessing.sharedctypes import RawArray
 
@@ -26,6 +29,7 @@ from types import SimpleNamespace
 import fitsio
 
 from desiutil.depend import add_dependencies, setdep
+from desiutil.dust import SFDMap
 
 import desimodel.focalplane
 
@@ -1087,6 +1091,9 @@ def merge_results_tile(out_dtype, copy_fba, params):
     Returns:
         None.
 
+    Notes:
+        20210928 : we populate EBV=0 rows in FIBERASSIGN with correct values.
+
     """
     (tile_id, infile, outfile) = params
     log = Logger.get()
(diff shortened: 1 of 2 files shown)


schlafly:
It's hard for me to test all the way into the real environment without a tag.  But I did enter the datasystems environment and check that I could instantiate an SFDMap and query it at a lot of positions, which I think is the only real risky part?  Fine from my perspective to merge, and we can do real tests at KPNO from fiberassign/master before cutting the next tag, okay?

araichoor:
Thanks for checking: if you succeed in executing SFDMap, I guess it should be ok.
For what is worth, what I had in mind was to check that in the environment used here:
https://github.com/desihub/desisurvey/blob/4eceb0b67968061ec66fe470c83d1cc3101f6cf0/bin/fba-main-onthefly.sh#L30-L47
the $DUST_DIR is defined, and the dust maps are in $DUST_DIR.

Shall I merge to master now?

schlafly:
Yup, I checked that SFDMap(...).ebv(...) worked in precisely that environment.  I'm okay with your merging to master whenever you're ready.  I don't think we plan to make a new tag or change the tag used in operations especially soon, though, right---that will come when we've decided what to do about the STD ~bug.

araichoor:
Ok, I ll thus merge this PR in few mins.

Issue #410, opened 2021-09-24
STD_WD included in the per-petal STD counting


araichoor:
Apparently, for the main survey, fiberassign considers as "STD", to fulfill the --standards_per_petal requirement: STD_WD, STD_BRIGHT, STD_FAINT.
However, the reduction pipeline is not using STD_WD; so one wants to discard STD_WD from this counting.

My understanding:
- fiberassign.targets.default_main_stdmask() has this list of masks: https://github.com/desihub/fiberassign/blob/33c15ad9904091d185e6329a7f6f6e2575e81ed1/py/fiberassign/targets.py#L230-L237
- then objects passing those masks are defined as TARGET_TYPE_STANDARD, which used by the c++ code.

Removing that line: https://github.com/desihub/fiberassign/blob/33c15ad9904091d185e6329a7f6f6e2575e81ed1/py/fiberassign/targets.py#L235
should address that.
Though, one has to be careful to preserve backward-compatibility.

araichoor:
[discussion potentially also relevant for https://github.com/desihub/fiberassign/issues/411]

The non-trivial point here is to ensure backward-compatibility.
I m not sure to see how we can avoid to add a switch on a date (nowtime? rundate? mtltime? ...).

_Note that we would then need to make sure that the cut off date in the PR is before we actually design the tiles with the new tagged version; or pick a cut off date later in the future._

@sbailey also suggested during yesterday s telecon to make some dates (`rundate`?) a global variable throughout fiberassign.

I ve looked at the code: one way to go would be to propagate the `rundate` down to the `fiberassign.targets.default_main_stdmask()` function, `rundate` being an argument of `fiberassign.scripts.assign.parse_assign()`.

in `fiberassign.bin.fba_run`:
    `args = fiberassign.scripts.assign.parse_assign()` # rundate is one of the arguments

in `fiberassign.scripts.assign.run_assign_bytile()` or `fiberassign.scripts.assign.run_assign_full()`:
    `hw, tiles, tgs, tagalong = run_assign_init(args, plate_radec=plate_radec)`

then the flow of functions successively called in `run_assign_init()`, where one would need to propagate `rundate`:
https://github.com/desihub/fiberassign/blob/33c15ad9904091d185e6329a7f6f6e2575e81ed1/py/fiberassign/targets.py#L1030-L1033
which calls:
[...]

PR #415, opened 2021-10-01
Remove STD_WD from the per-petal STD counting

araichoor:
This PR addresses issue https://github.com/desihub/fiberassign/issues/410.
It discards STD_WD from standard counting in `targets.default_main_stdmask()`.
@sbailey: the fix is simple, however I d like to have advice on the choices I made.

To ensure backward-compatibility, I propagate the rundate argument down to that function (defaulting to None), and add a switch based on a cutoff rundate.
I have currently set this cutoff date to `2021-10-01T19:00:00+00:00`: we would have to merge this PR and make a fiberassign tag before tomorrow night, or change that cutoff date in the code, if merging+tagging later.
Also, I have made the default behavior to discard STD_WD (i.e. STD_WD would be included only if a rundate argument is provided, and if it is before the cutoff date).

In preparation of future similar events, I ve added a `get_date_cutoff()` function; the idea is that it would centralize the cutoff dates (rundate, mtltime, etc) for any future similar changes.
I ve added that function in `fba_launch_io.py`: however proceeding this way adds a dependency to `fba_launch_io.py` in `targets.py`; I had to move around few imports in `fba_launch_io.py` to avoid some circular imports issue, as `fba_launch_io.py` imports some modules, which import `targets.py`.
That is a possible significant drawback: if advised to do so, I can remove that `get_date_cutoff()` function, thus removing any dependence to `fba_launch_io.py` in `targets.py`; the cutoff date would then just appear in `targets.default_main_stdmask()`.

Lastly, I notice that `default_main_stdmask()` function is also used in `fiberassign/qa.py` and indirectly in `fiberassign/vis.py`; apparently that is used in the test suite.
I am not familiar with those scripts, I have not modified those files for now in this PR; I just checked that `python setup.py test` still runs; I *think* that the default `rundate=None` argument just propagates through the functions.
Same, if advised to do so, I could have a closer look to that.

### DIFF
--- a/py/fiberassign/data/cutoff-dates.yaml
+++ b/py/fiberassign/data/cutoff-dates.yaml
@@ -1,2 +1,2 @@
 rundate:
-    std_wd: "2021-10-01T19:00:00+00:00" # std_wd not counted in per-petal-std after that rundate
+    std_wd: "2021-12-01T19:00:00+00:00" # std_wd not counted in per-petal-std after that rundate
--- a/py/fiberassign/data/cutoff-dates.yaml
+++ b/py/fiberassign/data/cutoff-dates.yaml
@@ -1,0 +1,2 @@
+rundate:
+    std_wd: "2021-10-01T19:00:00+00:00" # std_wd not counted in per-petal-std after that rundate
--- a/py/fiberassign/fba_launch_io.py
+++ b/py/fiberassign/fba_launch_io.py
@@ -1499,7 +1499,7 @@ def create_mtl(
         else:
             log.info(
                 "{:.1f}s\t{}\tas desitarget.__version__={} < 1.2.2, we do not add columns from {}".format(
-                    time() - start, step, desitarget.__version__, ",".join(targdir)
+                    time() - start, step, desitarget.__version__, ",".join(targdirs)
                 )
             )
 
--- a/py/fiberassign/fba_launch_io.py
+++ b/py/fiberassign/fba_launch_io.py
@@ -54,11 +54,6 @@ import desimeter
 # fiberassign
 import fiberassign
 from fiberassign.utils import Logger
-# AR not importing the function, to avoid
-# AR    circular imports with targets.py, which has imports from fba_launch_io.py...
-# AR    https://github.com/desihub/fiberassign/pull/415#issuecomment-934940698
-import fiberassign.scripts.assign # used functions: parse_assign, run_assign_full
-import fiberassign.assign # used functions: merge_results, minimal_target_columns
 
 # matplotlib
 import matplotlib.pyplot as plt
@@ -1376,7 +1371,7 @@ def create_mtl(
         20210914 : add equivalent of match_ledger_to_targets() for secondary (except for backup)
         20210917 : add possibility of extra secondary folders, as main2/, main3/, etc (see get_desitarget_paths())
                     for that, changing targdir arguments to targdirs
-        20211018 : minimal_target_columns -> fiberassign.assign.minimal_target_columns
+        20210930 : moving some imports inside this function to avoid circular imports.
     """
     log.info("")
     log.info("")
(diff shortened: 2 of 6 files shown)


araichoor:
in case useful, details about the circular imports issue I encountered (@schlafly):
- I added in fiberassign.targets.py:
    ```
    from fiberassign.fba_launch_io import assert_isoformat_utc, get_date_cutoff
    ```
- however, fiberassign.fba_launch_io.py has (at the top of the script in the master version) imports from fiberassign.scripts.assign.py and fiberassign.assign.py:
    ```
    from fiberassign.scripts.assign import parse_assign, run_assign_full
    from fiberassign.assign import merge_results, minimal_target_columns
    ```
- and fiberassign.scripts.assign.py and fiberassign.assign.py have imports to fiberassign.targets.py; e.g. in fiberassign.assign.py:
    ```
    from .targets import (TARGET_TYPE_SCIENCE, TARGET_TYPE_SKY, TARGET_TYPE_SUPPSKY,
                          TARGET_TYPE_STANDARD, TARGET_TYPE_SAFE, desi_target_type,
                          default_target_masks, default_survey_target_masks)
    ```
I solved that by moving the fba_launch_io imports from scripts.assign and assign inside the fba_launch_io.create_mtl () and fba_launch_io.launch_onetile_fa() functions, where those are actually used.

schlafly:
I played with this for a second.  It looks like you could get around this if you chose by replacing the from x import y calls with basic "import x" followed by calls to "x.y(...)".  I had not appreciated this before, but apparently from x import y requires running module x, but just "import x" has some partial initialization mode where it just figures out the names of everything but doesn't try to run them.
https://dev.to/deepjyoti30/why-you-should-keep-away-from-cyclic-imports-in-python-1n75
But I think your solution is fine as well.

sbailey:
@araichoor that sounds good.  I like the idea of collecting all of these cutoff dates into a single place with comments explaining what they are for.

Let's wrap the access in an function like
```python
def get_rundate_cutoff(cutoff_name):
    """document..."""
    fn = resource_filename("fiberassign", "data/cutoff-dates.yaml")
    with open(fn, "r") as f:
        config = yaml.safe_load(f)
    rundate_cutoff = config["rundate"][cutoff_name]
    return rundate_cutoff
```

So that the N>1 places that need a cutoff date just need to call a function instead of putting together nearly identical `resource_filename`, `yaml.safe_load`, etc. logic.

araichoor:
ok great, I ll do that, thanks for the prompt answer!

and, yes, 100% agreed about making a function to read that file.
one additional remark here: I think I ll create a new file (`py/fiberassign/util.py`?), where I would store such functions.
so far I ve put ~everything in `fba_launch_io.py`, but, because of this circular import stuff, such a function `get_rundate_cutoff()` cannot live in `fba_launch_io.py`.

and I could then move from `fba_launch_io.py` to `util.py` some functions which are general. as e.g.:
https://github.com/desihub/fiberassign/blob/ac6e935614cad8c888f2bd5b93000eb955185292/py/fiberassign/fba_launch_io.py#L75-L89
https://github.com/desihub/fiberassign/blob/ac6e935614cad8c888f2bd5b93000eb955185292/py/fiberassign/fba_launch_io.py#L152-L175
https://github.com/desihub/fiberassign/blob/ac6e935614cad8c888f2bd5b93000eb955185292/py/fiberassign/fba_launch_io.py#L178-L201

schlafly:
+1 for moving code that isn't tightly coupled to fba_launch into something like
https://github.com/desihub/fiberassign/blob/master/py/fiberassign/utils.py
Let's not add a new file util.py in addition to the existing utils.py, though!

araichoor:
Tests are failing because they apparently cannot find the newly added `data/cutoff-dates.yaml` file.
I apologize, I may need help here (@sbailey ?).
In case it's useful:
- in the code, that file is searched with `resource_filename("fiberassign", "data/cutoff-dates.yaml")`;
- I've simply added that file with `git add`, then committed with `git commit`.

araichoor:
I thought I'd be smart with mimicking what is e.g. in desitarget with adding in setup.py `setup_keywords['package_data'] = {'fiberassign': ['data/*',],}`.
But then the checks were not running, with only displaying "Waiting for checks information...".
So, I've reverted that commit; however the checks are not running either now.
I hope I have not break anything, maybe `setup.py` is file needing special treatment...

PR #416, opened 2021-11-20
Skyhealpixs gaia

araichoor:
This PR enables the reading of Gaia-based good sky positions for stuck positioners for the Main BACKUP program.
It is dependent on the twin desitarget PR https://github.com/desihub/desitarget/pull/771.

The chosen approach assumes new maps will be stored in `$SKYHEALPIXS_DIR`.
I expect this environment variable to be defined as `$DESI_ROOT/target/skyhealpixs/v1`.

I have defined a new args.sky_lookup_source argument in both `fiberassign.bin.fba_launch` and `fiberassign.scripts.assign.parse_assign()`.
Allowed values are "ls" and "gaia" and default values are:
- in `fba_launch` : "gaia" if survey=="main" and program=="BACKUP", else "ls".
- in `parse_assign()`: "ls".
That should allow backwards-compatibility.

Then `fiberassign.stucksky.py` will use this argument:
- if sky_lookup_source == "ls", it will read maps in $SKYBRICKS_DIR
- if sky_lookup_source == "gaia", it will read maps in $SKYHEALPIXS_DIR

@sbailey : for the spectroscopic pipeline to know what has been used:
I ve added in the 0-th extension header of fiberassign-TILEID.fits.gz:
- `SKYHEALPIXS_DIR` in the dependences, recording where the Gaia maps are
- a `LKSKYSRC` keyword, which is the args.sky_lookup_source argument:
   - if `LKSKYSRC=="ls"`, then one should use  `$SKYBRICKS_DIR`,
   - if `LKSKYSRC=="gaia"`, then one should use  `$SKYHEALPIXS_DIR`.

Preliminary tests with temporary:
- maps in `/global/cfs/cdirs/desi/users/koposov/sky_mask/gaia_data`
- catalogs in `/global/cscratch1/sd/adamyers/gaiadr2/1.3.0.dev5218/targets/main/resolve/backup/`
work.
But those are limited, as they require hacks in the code.
Once the desitarget backup catalogs + ledgers will be in place, I could proceed to more exhaustive tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,7 @@ jobs:
       DESIMETER_VERSION: 0.6.5
       DESIMODEL_DATA: trunk
       DESIMODEL: ${{ github.workspace }}/desimodel
-      DESITARGET_VERSION: 1.0.0
+      DESITARGET_VERSION: 2.2.1
     steps:
       - name: Cancel Previous Runs
         uses: styfle/cancel-workflow-action@0.8.0
--- a/bin/fba_launch
+++ b/bin/fba_launch
@@ -760,7 +760,7 @@ if __name__ == "__main__":
 
     # AR lookup_sky_source
     if args.lookup_sky_source is None:
-        if (args.survey == "main") & (args.program == "backup"):
+        if (args.survey == "main") & (args.program == "BACKUP"):
             args.lookup_sky_source = "gaia"
             if os.getenv("SKYHEALPIXS_DIR") is None:
                 log.warning("SKYHEALPIXS_DIR not set: stuck positioners will not be used for skies.")
--- a/bin/fba_launch
+++ b/bin/fba_launch
@@ -684,6 +684,13 @@ if __name__ == "__main__":
         action='store_true',
         default=False,
     )
+    parser.add_argument(
+        "--lookup_sky_source",
+        help="'ls' or 'gaia' (default='gaia' if main/backup, 'ls' otherwise)",
+        type=str,
+        choices=["ls", "gaia"],
+        default=None,
+    )
 
     args = parser.parse_args()
     log = Logger.get()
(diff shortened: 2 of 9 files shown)


araichoor:
The above check rightfully fails because the companion PR in desitarget has not been merged yet.
```
ModuleNotFoundError: No module named 'desitarget.skyhealpixs'
```
I guess the proper way to proceed, once we are happy with everything is to first merge the PR https://github.com/desihub/desitarget/pull/771, then this PR.

araichoor:
The desitarget companion PR https://github.com/desihub/desitarget/pull/771 has been merged, and is now present in desitarget/2.2.0 (https://github.com/desihub/desitarget/blob/master/doc/changes.rst#220-2021-11-21).
Though, if my understanding is correct, the fiberassign github tests load desitarget/1.0.0:
https://github.com/desihub/fiberassign/blob/4115bfa6afbf809d819258e06f4e0ad6c551bc33/.github/workflows/test.yml#L16-L21
So we probably also want to update the desitarget version in this test.yml to 2.2.0? (And if so, maybe also bring up-to-date the used desimeter and desiutil versions?)
@sbailey: I am waiting for your instructions here, as I am not familiar with those (note that there also this PR 
also has pending with github tests : https://github.com/desihub/fiberassign/pull/415#issuecomment-974755992).

sbailey:
@araichoor yes please update the desitarget version in .github/workflows/test.yml, and other versions as needed, and add a note in doc/changes.rst that this version of fiberassign now requires desitarget >= 2.2.0 .

In general we try to run the github tests with the *oldest* known working version of dependencies so that cases like this make it clear that we now have a different dependency version requirement.  A nightly cronjob at NERSC does the complimentary test that master/main of everything is compatible.

araichoor:
@sbailey: thanks for the guidelines.
I ve only updated the desitarget version in `.github/workflows/test.yml`, as the test still work with other dependencies, and I am not familiar with all those yet.
I could have a closer look when things are more quiet.

Anyway, now the checks pass, great! I'll merge the PR.

Just, for the record: in the `fba_rerun_io. fba_rerun_fbascript()` routine,  `$SKYHEALPIXS_DIR` is expected to be like `$DESI_ROOT/target/skyhealpixs/v1`. If we decide to pick another path name, I will have to update that in `fba_rerun_io.py. fba_rerun_fbascript()`; but that is not critical, as this function is not used in the daily ops fiberassign reproducibility tests.

PR #417, opened 2021-11-20
Gaia std

araichoor:
This PR assigns the `TARGET_TYPE_STANDARD` to GAIA_STD_BRIGHT and GAIA_STD_FAINT targets.

I have basically duplicated the "stdmask" lines to "gaia_stdmask" lines in `scripts.assign.py` and `targets.py`.
This noticely:
- adds a new `--gaia_stdmask` argument to `scripts.assign.parse_assign()` (defaulting to None, as for `--stdmask`);
- introduces a dependency on MWS_TARGET in `targets.py` (I've tried to protect against the case where no MWS_TARGET column is present, imagining some special fiberassign run).

I did not include `GAIA_STD_WD`, as we are going to remove the WD from the standard counting requirement.

With this PR, the code will assign `TARGET_TYPE_STANDARD` to any STD_BRIGHT, STD_FAINT, STD_WD (will be removed in another PR), GAIA_STD_BRIGHT, and GAIA_STD_FAINT.
But as it happens that:
- the main bright/dark have zero GAIA_STD_* targets,
- the main backup has zero STD_* targets,

so this approach selects STD_* for bright/dark and GAIA_STD_* for backup, without requiring to add a dependency on the program through those codes.

For the same reason it should be backwards-compatible for the main bright/dark observed tiles.
For other surveys (cmx, sv1, sv2, sv3), I ve not really checked, but those surveys already are not backwards-compatible with the current master version anyway.

I ve tested / developed using the temporary main backup catalogs here: `/global/cscratch1/sd/adamyers/gaiadr2/1.3.0.dev5218/targets/main/resolve/backup`.

### DIFF
--- a/bin/qa-fiberassign
+++ b/bin/qa-fiberassign
@@ -81,15 +81,16 @@ for filename in args.tilefiles:
     stdmask = None
     skymask = None
     excludemask = None
+    gaia_stdmask = None
     col=None
     if survey is not None:
             
-        sciencemask, stdmask, skymask, suppskymask, safemask, excludemask = \
+        sciencemask, stdmask, skymask, suppskymask, safemask, excludemask, gaia_stdmask = \
                 default_survey_target_masks(survey)
   
     else:
         fsurvey, col, sciencemask, stdmask, skymask, suppskymask, safemask, \
-                excludemask = default_target_masks(fbtargets)
+                excludemask, gaia_stdmask = default_target_masks(fbtargets)
         survey = fsurvey
 
 
--- a/py/fiberassign/assign.py
+++ b/py/fiberassign/assign.py
@@ -829,26 +829,32 @@ def read_assignment_fits_tile(tile_file):
 
         fsciencemask = None
         fstdmask = None
+        fgaia_stdmask = None
         fskymask = None
         fsuppskymask = None
         fexcludemask = None
         fcol = None
         if survey is not None:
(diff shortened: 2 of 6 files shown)


araichoor:
I had to propagate the changes in several files so that the tests pass (commit https://github.com/desihub/fiberassign/pull/417/commits/3c40f863cc5a468382957472f54c894a737f32d0).
I've done my best, but I am not familiar with those files (except `fiberassign.assign.py`), which are not used by the daily pipeline, so I don't know how careful we should be here.
For instance in `fiberassign.qa.py`, I've chosen the class and type for the Gaia standards, but I'm not sure if my choices matters.

dstndstn (approved):
I really don't know this part of the code well enough to give a critical review.

araichoor:
Ok, no worries, thanks!

araichoor:
I have now re-validated the code with the official 2.2.0 desitarget main/backup catalogs.

Besides, to verify backwards-compatibility, I have rerun a random 10% of the designed main bright/dark tiles with that branch: those pass the reproducibility test (except 6 tiles originally designed with desitarget catalogs 1.0.0, but we know those are not exactly reproducible anyway).

I could not rerun the whole ~1890 designed main bright/dark tiles, because fba_launch takes ages on the cori interactive nodes (a single tile can take >10min! whereas a typical call is <30s).

PR #420, opened 2021-11-23
Fba rerun5

araichoor:
This PR brings few updates to `fba_rerun_io.py` with:
- handling multiple secondary catalogs (main, main2, etc);
- handling negative FIELDROT, HA in `fba_run` call;
- correcting a typo.

### DIFF
--- a/py/fiberassign/fba_rerun_io.py
+++ b/py/fiberassign/fba_rerun_io.py
@@ -126,6 +126,15 @@ def fba_rerun_get_settings(
             )
             sys.exit(1)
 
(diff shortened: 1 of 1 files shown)


araichoor:
Those changes are not critical, in particular they are _not_ used by `https://github.com/desihub/desisurveyops/blob/main/bin/validate_lastnight_fba` (which uses only `fba_rerun_check()` from `fba_rerun_io.py`, which is not changed here), but I happen to notice those, so I ll merge now, along the way with other PRs.

PR #421, opened 2021-11-24
sort fiberassign output hdus GFA_TARGETS and POTENTIAL_ASSIGNMENTS

dstndstn:
GFA_TARGETS by TARGETID
POTENTIAL_ASSIGNMENTS by LOCATION then TARGETID.

### DIFF
--- a/py/fiberassign/assign.py
+++ b/py/fiberassign/assign.py
@@ -533,11 +533,14 @@ def write_assignment_fits_tile(asgn, tagalong, fulltarget, overwrite, params):
 
         fdata = np.zeros(navail, dtype=avail_dtype)
         off = 0
+        # The "FAVAIL" (available targets) HDU is sorted first by LOCATION,
+        # then by TARGETID.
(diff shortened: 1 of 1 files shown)


araichoor:
Apologies @dstndstn, I somehow totally missed that PR (it didn't appear in my emails).
Thanks a lot for that.
The code change look (simple! and) fine to me.

It looks safe to me: @schlafly do you confirm we can merge that without going to the exercise of making a tile at KPNO and NERSC and assert the two files are identical?

I could then create a fiberassign/5.4.1 tag, and use it for observations?

schlafly:
Two cents: I might merge and then wait until we have some reason to cut a tag & install, just to minimize the number of versions of fiberassign in use.  This is good but doesn't actually change assignment so I think we should round it up with other changes.

araichoor:
thanks for the suggestion, I agree, let s proceed that way.
so I ll merge that PR today.

PR #426, opened 2022-04-19
Desi fa smallrun v3

araichoor:
This PR brings several updates to`bin/desi_fa_smallrun`.
Noticely, this PR uses a more mtl-like approach, based on desitarget routines.

This PR version was used to produce the full-survey fiberassign run on 20220227, which provides the expected numbers reported here: https://desi.lbl.gov/trac/wiki/keyprojects/y1kp1?version=50

The list commits names allows one to see the various changed things.

### DIFF
--- a/bin/desi_fa_smallrun
+++ b/bin/desi_fa_smallrun
@@ -1038,10 +1038,10 @@ if __name__ == "__main__":
     )
     parser.add_argument(
         "--program",
-        help="dark or bright (default=dark)",
+        help="dark or bright or backup (default=dark)",
         type=str,
         default="dark",
-        choices=["dark", "bright"],
+        choices=["dark", "bright", "backup"],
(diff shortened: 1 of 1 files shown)


araichoor:
@sbailey : this PR only modifies this stand-alone `bin/desi_fa_smallrun` script, which I use to run the full-survey fiber assignment.
I m thinking to merge it tomorrow morning, as it doesn t impact the survey operations, and I don t expect anyone to dive in that script for a review (and I think I might be the only one to use it!).

the idea is that this version is included in the fiberassign tag we would make, and I could reference to that when I ll write a small document to describe how I generated the numbers in https://desi.lbl.gov/trac/wiki/keyprojects/y1kp1?version=50.

araichoor:
merging now.

PR #428, opened 2022-04-22
query ToO-fiber.ecsv for main if args.too_tile=False

araichoor:
ToOs can have two `TOO_TYPE` values, `"TILE"` and `"FIBER"`.
When designing a regular main survey tiles, only `TOO_TYPE="FIBER"` ToOs are kept.
Up to now, both types were in theory stored in `ToO.ecsv`; though there have no been no `TOO_TYPE="FIBER"` targets so far.

The PR https://github.com/desihub/desitarget/pull/795 introduces the `ToO-fiber.ecsv` file, which will only contains the `TOO_TYPE="FIBER"` targets; the new `ToO.ecsv` will only contains the `TOO_TYPE="TILE"` targets.

This fiberassign PR makes `fba_launch` use `ToO-fiber.ecsv` for main survey tiles if `args.too_tile=False` (i.e. for the "regular" main survey observations), instead ot `ToO.ecsv`.
The motivation is that this `ToO-fiber.ecsv` is -- and hopefully will remain -- much smaller, hence quicker to read.
For instance, currently, there are ~350k rows in `ToO.ecsv` but no `TOO_TYPE="FIBER"` targets: at kpno, `fba_launch` typically takes ~10s -- which is 1/3 of the total time -- to read that file to discover that no targets should be retained.

Comments:
- as, up to now, there has been no `TOO_TYPE="FIBER"` targets in `ToO.ecsv` for the main survey, even ToO-dedicated tiles (run with survey="main" and args.too_tile=True) should in theory be reproducible, because that file will contain the same targets (with TIMESTAMP<mtltime); however, the TIMESTAMP handling in desitarget prevents that, as the original values are overwritten (I tried to rerun tileid=82634, but it doesn t work because of that, and the addition of new ToOs overlapping that tile); the new desitarget PR corrects that, with preserving the original TIMESTAMP;
- the desitarget new approach implies that ToO-dedicated tiles designed in the future would only use `TOO_TYPE="TILE"` targets (whereas in the current design it should use both types, though this has never happened yet as there are no `TOO_TYPE="FIBER"` targets); this comment is made in the PR https://github.com/desihub/desitarget/pull/795 discussion;
- if the `ToO-fiber.ecsv` file size grows again to a large size, a possible refinement in the future would be to isolate targets up to a certain date in the past in a separate file (credits to @geordie666); though that would require a careful handling of reproducibility, which is why it is not implemented here.


sbailey:
@geordie666 @schlafly can you review this?

schlafly:
Thanks Anand.  This looks good to me.  Trying to understand better the one concern I raised before: for old tiles that used the ToO mechanism (M31 tiles... I forget what else?), I imagine we'll pay some ~reproducibility price in that to rerun them now we'd have to add --too_tile?  Or maybe not, since it looks like you already had too_tile as an argument, and now you're just using it more?

geordie666:
I think this looks fine to me, too, pending Eddie's question. I couldn't see any obvious bugs or issues.

geordie666 (approved):

araichoor:
thanks Eddie and Adam for the prompt review!

about the point you raised: 
- as there have been no `TOO_TYPE="FIBER"` targets in the main `ToO.ecsv` up to now, I think this PR changes nothing in the principle;
- but, as I mentioned in the original message, I think that the reproducibility is possibly already broken for some of those tiles.

I v looked at tileid=82634 (m31), here are the details:
this tile was on m31, designed on 20211228; ToOs had then `TIMESTAMP="2021-12-22T15:23:45+00:00"`.
however, some additional targets were added for the design of 82635, which has some overlap with 82634, on 20220120.
and, apparently, the TIMESTAMP for *all* targets has been updated then, to `"2022-01-17T17:28:15+00:00"` (that is one of the things "fixed" in the new desitarget PR).
so when I re-run 82634, the cut on `TIMESTAMP<=mtltime` is ineffective, and some new targets are in the tile footprint, so the fiber assignment is not as the original one.

tileids=80980-81 on 20210708:
I guess those are the tiles Adam refers to; those are two overlapping tiles.
those were designed quite early (e.g. the `ToO.ecsv` didn t any `TIMESTAMP` column then), and with a special script (`$DESI_ROOT/survey/fiberassign/special/20211228/082/fba_launch_scnd_fulltile`), which allowed me to not re-assigned in the second tile targets already assigned in the first tile (typically that would be handled in a cleaner way now with the tertiary program).
I ve not tried to re-run that, but I think that it should be fine with only updating `fba_launch_scnd_fulltile` with the newly added arguments in the `get_desitarget_paths()` function.
Because in the current `ToO.ecsv` catalog, I think no new targets have been added in that footprint since then.

apart from the tiles mentioned above, I ve in my listing of "special" tiles since the start of the main survey, the tiles below, which either are tertiary programs, or test tiles:

- 20210513, tileids=81108-13: 8 specialdark tiles;  [desi-survey 2462] Fiberassign test tiles for tonight;
- 20210529, tileid=80977: 1 specialdark tile; [desi-targets 3684] QSO+ELG tile in Sag stream;
- 20210530, tileids=80978-79: 2 specialbright tiles;  [desi-survey 2666] Request for poor weather tiles;
- 20220324, tileid=82636: 1 specialodin (=tertiary 0)
- 20220419, tileid=82637-82691: 55 specialtertiary1 tiles ("cosmoslowz")

schlafly (approved):

schlafly:
Thanks Anand.  I'm happy.  I was confused because I read the PR as introducing the new argument too_tile, which clearly needs to be set for old dedicated tiles to work.  But I now understand that that argument was always there and is now just being used in more places.  Too bad that reproducibility for these dedicated programs broke anyway, but fortunately that's not where we're trying to be very careful!

araichoor:
exactly, the `args.tootile` was already there; it s just that now it s propagated in the code to reach the `fba_launch_io.get_desitarget_paths()` function, which sets the paths to the desitarget products.
sorry I should have precised that in the beginning :)

I ll have a last, calm look at that, then I ll merge + cut a 5.5.1 fiberassign tag.

araichoor:
or should rather the new tag be 5.6.0? (@sbailey)

schlafly:
I think we think this is fully backward compatible, modulo tiles that we've never been able to reproduce.  I'd vote for 5.5.1 (but I don't think we should spend more than a few minutes trying to decide!).

geordie666:
As this is a backward compatible change rather than a bug-fix — and we're at a version > `1.0.0` — I think `5.6.0` is correct. Rules are [here](https://desi.lbl.gov/trac/wiki/Computing/Software/Guidelines#TagNames).

araichoor:
for the record:
with a last sanity check, I realize that one of the two functions where I ve added an optional `too_tile` argument (defaulting to `False`), `fba_launch_io.get_desitarget_paths()`, is also used in two other scripts.
however, I think no change is needed, so I plan to merge as is.

`fba_rerun_io.fba_rerun_intermediate()`:
that function is part of a suite of functions to rerun the assignment, with some flexibility.
the choice to default to `too_tile=False` sounds ok here as I expect that this function is used for sv3 or main tiles.
I m not sure anyone uses that function with the master version anyway; @ashleyjross @jalasker please let me know if you use that function and something is broken here, I could fix that in another PR (but which wouldn t be needed for observations).
_I emphasize that `fba_rerun_intermediate()` is **not** used by`fba_rerun` in desisurveyops `validate_lastnight_fba`_

`bin/fba_launch_dc3r2_gama`:
that script is an early variation of `fba_launch`, used to design sv1 tiles for the dc3r2 program on 20210511 (tileids=80971-76).
strictly speaking the PR shouldn t change anything here (as it s a survey="sv1" tile, the code will go for the sv1`ToO.ecsv`, but that file doesn t exists..)
I m not sure anyone is going to try to rerun the fiber assignment here anyway..

araichoor:
and thanks @geordie666 for pointing out the rules to pick a tag name.
it s actually after having read those that I thought that 5.6.0 may be more relevant!

Issue #429, opened 2022-04-28
incomplete docstring for load_targets_table


forero:
While trying to update this notebook 

https://github.com/desihub/tutorials/blob/39ab3f17a648322e44c7ab932eead2edaeb335f8/FiberAssignAlgorithms_Part1.ipynb

to work under the 22.2 kernel, I found that the function `load_targets_table` now has three required inputs, instead of the two provided in the notebook. The new input seems to be `tagalong`, but the docstring in `load_targets_table` does not describe what `tagalong` is, making it difficult to correctly patch the notebook.

araichoor:
I think you re correct, the new entry is `tagalong`.
I believe that s been introduced by @dstndstn in this PR https://github.com/desihub/fiberassign/pull/353.
From a quick look at `target.py`, it looks that this newly introduced argument `tagalong` does not appear in the docstr of other functions (e.g. `load_target_file()`).

@dstndstn : do you confirm? if so, one should update the docstr.
If you prefer, I could do that update.
I may just need from your side a generic description of this `tagalong`; i.e. something similar to:
```
tgs (Targets): The targets object to modify.
```

Thanks.

araichoor:
The PR https://github.com/desihub/fiberassign/pull/430 addresses this ticket.

PR #430, opened 2022-04-28
add tagalong argument in the docstr

araichoor:
This PR adds the `tagalong` argument in the docstr of the functions using it in `targets.py` and `assign.py`.

This was raised in https://github.com/desihub/fiberassign/issues/429 (and the `taglong` was introduced in PR https://github.com/desihub/fiberassign/pull/353).

`tagalong` is also used in some `vis.py` functions, but I m not at all familiar with those, and they haven t any docstr yet, so I didn t do any modification there.

### DIFF
--- a/py/fiberassign/assign.py
+++ b/py/fiberassign/assign.py
@@ -138,7 +138,7 @@ def write_assignment_fits_tile(asgn, tagalong, fulltarget, overwrite, params):
     Args:
         outroot (str):  full path of the output root file name.
         asgn (Assignment):  the assignment class instance.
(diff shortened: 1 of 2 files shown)


dstndstn:
Could you maybe drop the "A new one can be created..." sentence except for the one on "load_target_file"?  Because a newly-created one is passed to the `load` functions, then that same one has to be passed to the `write` functions.

araichoor:
Sure, done.
Thanks for the check!

forero:
Thanks @dstndstn @araichoor !

PR #435, opened 2022-08-18
Compile script to run when installing a branch

weaverba137:
@sbailey please review the script itself and verify correct installation.  As far as I can tell the compile command runs successfully, but I'm not sure what commands to call to test the actual compiled code.

Note that `${INSTALL_DIR}` (defined by desiInstall) is used in addition to the Python executable.

See also desihub/desiutil#188.

### DIFF
--- a/etc/fiberassign_compile.sh
+++ b/etc/fiberassign_compile.sh
@@ -6,12 +6,8 @@
 # The single command-line argument is the path to the Python executable
 # that is calling desiInstall.
 #
-# This is meant to be run in ${INSTALL_DIR}, which is automatically
-# defined by desiInstall.
-#
 # As of 2022, setup.py in generally deprecated, so whatever replaces it
 # can go in this script.
 #
 py=$1
-cd ${INSTALL_DIR}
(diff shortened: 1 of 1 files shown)


etc/fiberassign_compile.sh
```diff
@@ -0,0 +1,17 @@
...
+# that is calling desiInstall.
+#
+# This is meant to be run in ${INSTALL_DIR}, which is automatically
+# defined by desiInstall.
+#
+# As of 2022, setup.py in generally deprecated, so whatever replaces it
+# can go in this script.
+#
+py=$1
+cd ${INSTALL_DIR}
```
sbailey, line 16: I have tested this with desihub/desiutil#188 and it works fine, but see the suggestion there to put the responsibility on desiInstall to cd to the top level directory first before calling this script instead of setting an environment variable that the scripts have to know about to be able to find that top-level directory.

sbailey (approved):
Works great, as tested with desihub/desiutil#188 (now merged).

PR #451, opened 2023-02-10
add script and functions to retrieve svn infos

araichoor:
This PR adds a script to retrieve all the fiberassign svn revisions for a set of tiles.

It is a follow-up of https://github.com/desihub/desispec/issues/1834.

Following discussion in that issue, I ve put the script in `fiberassign/etc`.
Besides, I ve factored out two functions in `fiberassign/utils.py`.

The script calls the `svn log` command lots of times to retrieve information: because of that, I kept the calls in a simple `for` loop, so that only one call at a time runs; that way the server is not overloaded.

I m happy to implement any suggested change.

### DIFF
--- a/etc/fba_tiles_svn_revisions.py
+++ b/etc/fba_tiles_svn_revisions.py
@@ -1,0 +1,93 @@
+#!/usr/bin/env python
+
+import os
+import subprocess
+from time import time, sleep
+import numpy as np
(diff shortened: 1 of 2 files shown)


sbailey:
Apologies for letting this PR dangle for so long!  This looks fine, merging now.

Issue #452, opened 2023-03-07
focal plane size mismatch between Megamapper and DESI-I


forero:
I have encountered a small issue while running fiberassign with different focal plane files, specifically Megamapper and DESI-2. When using Megamapper files, I observed that run_assign_full() trims the targets to those within the tile using targets_in_tiles(). However, I noticed that the focal plane size used in the function is hardcoded here: https://github.com/desihub/fiberassign/blob/main/src/hardware.cpp#L195.

As a result, the fiberassign files only include potential targets within the current DESI focal plane angular size, which is problematic for Megamapper, as it has a larger focal plane. I have included a plot below for reference.

I would appreciate guidance on the best approach to running fiberassign when the focal plane size differs from that of DESI-I.

<img width="584" alt="Captura de Pantalla 2023-03-07 a la(s) 12 57 09 p  m" src="https://user-images.githubusercontent.com/1848618/223509435-11829f5c-53d3-434e-ad9c-f1980c228eb8.png">

araichoor:
nice catch!

though I m not really familiar with all the cpp codes... and I ve no time right now to investigate.

in the `hardware.cpp` part of the code you point to:
- I see there are also other parameters hard-coded
- it s written `Hard-coded parameters.  These could be moved to desimodel and passed into this constructor as arguments.`

would that be the way to go? but it could require some work to get things right..

for what is worth, the tile radius in `desimodel` comes from here: https://github.com/desihub/desimodel/blob/575be9489bac29d210adc6683118e9826e6c1a06/py/desimodel/focalplane/geometry.py#L34-L43.
and I think that comes from reading this one: https://desi.lbl.gov/trac/browser/code/desimodel/trunk/data/focalplane/platescale.txt.

forero:
Thanks a lot, Anand! Unfortunately, I'm still running into the issue where the fiberassign function is using the hardcoded angular scale from the C++ code. I have tried using updated platescale.txt files for Megamapper, but the targets are still being trimmed down to the DESI focal plane angular scale.

araichoor:
what I meant is to implement this comment: `These could be moved to desimodel and passed into this constructor as arguments.`, but, as said, I suspect it s quite some re-writing / work.

if I get it correctly, it could be something like:
- have some new megamapper files like `platescale.txt` in the svn repo, 
- update the desimodel codes -- with defaulting to desi for backwards compatibility
- then modify the cpp code to make use of desimodel.

[...]

PR #453, opened 2023-03-10
uses the focalplane radius from DESIMODEL instead of the hardcoded version in fiberassign

forero:
This PR fixes bug #452

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -6,7 +6,9 @@ fiberassign change log
 5.7.2 (unreleased)
 ------------------
 
-* no changes yet
+* uses the focalplane radius from DESIMODEL instead of the hardcoded version in fiberassign (PR `#453`_).
+
+.. _`#453`: https://github.com/desihub/fiberassign/pull/453
 
 5.7.1 (2023-02-06)
 ------------------
--- a/py/fiberassign/targets.py
+++ b/py/fiberassign/targets.py
@@ -46,6 +46,8 @@ from fiberassign.utils import assert_isoformat_utc, get_date_cutoff
 from datetime import datetime
 from astropy.time import Time
 
+import desimodel.io
+
 class TargetTagalong(object):
     '''
     This class holds data from the targeting input files that we want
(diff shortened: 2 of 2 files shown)


forero:
These changes only affect the function `targets_in_tiles()`

dstndstn (approved):
Looks fine to me!

araichoor:
sorry I notice I never hit "send" on a small message I had when you opened the PR.
- I apologize, I don t have time right now to carefully look at that
- though one thing we definitely want to check is that the "default" behavior is not changed (ie it does not change anything for the main survey).

I mention this because I notice that the two radii differ:
with the main branch
```
>>> from fiberassign.hardware import load_hardware
>>> load_hardware().focalplane_radius_deg
INFO: Loaded focalplane for time stamp 2023-03-22 20:35:13.114490+00:00
INFO: Focalplane has 801 fibers that are stuck or broken
1.65
```
with this PR:
```
>>> import desimodel.io
>>> desimodel.io.load_platescale()['theta'].max() 
1.651501393
```

for instance, I could imagine that this change could break the reproducibility of the fiberassign intermediate products (the `TILEID-{targ,sky,gfa,scnd,too}.fits` files, where the files generated with this PR would have more targets (larger radius); I guess the assignment will be the same, as those extra targets won t get assigned, but I think we d want to keep an exact reproducibility of everything if possible.

so maybe propagate in the code some argument that would default to the current behavior; don t know if that s trivial or not...

araichoor:
sorry Jaime for not having been responsive here.
addressing here my own question: I think it s good in terms of reproducibility for the main survey; good!

**details:**

I ve run `fba_rerun --infiberassign $DESI_TARGET/fiberassign/tiles/trunk/011/fiberassign-011946.fits.gz --outdir $OUTDIR`, with `$OUTDIR`:
- main (before this PR): `/global/cfs/cdirs/desi/users/raichoor/tmpdir/fiberassign_platescale_fix/main` 
- branch (ie this PR): `/global/cfs/cdirs/desi/users/raichoor/tmpdir/fiberassign_platescale_fix/branch`

from a quick look, output files are identical, perfect!

I think the reason is that the targets (sky, gfa, too) are queried in `fiberassign.fba_launch_io.py` with `desitarget.io. read_targets_in_tiles()`; and that desitarget function calls `desimodel.footprint.is_point_in_desi()`, which calls `desimodel.geometry.get_tile_radius_deg()`; and that function proceeds a bit differently:
https://github.com/desihub/desimodel/blob/d18fc5a7ff0c7f2c9cafee890c8c0858ad524ed2/py/desimodel/focalplane/geometry.py#L34-L43
```
def get_tile_radius_deg():
    '''Returns maximum radius in degrees covered by the outermost positioner.
    '''
    global _tile_radius_deg
    if _tile_radius_deg is None:
        rmax = get_tile_radius_mm()
        platescale = load_platescale()
        fn = interp1d(platescale['radius'], platescale['theta'], kind='quadratic')
        _tile_radius_deg = float(fn(rmax))
    return _tile_radius_deg
```
and it returns `1.6280324520485583`; which is a smaller radius than both radii mentioned in my previous message (`1.65` and `1.651501393`).
so the selected targets in the intermediate fiberassign files aren't impacted by this, I think.

PR #454, opened 2023-05-08
Add function to Assignment object, check_avail_collisions(), to check for collisions for potential targets

dstndstn:
Add `Assignment.check_avail_collisions(tile, all_matches=False)`, which returns a dict of `(locid, targetid)` to `bitmask`, describing all collisions with STUCK or BROKEN positioners, and petal and GFA edges, for potential assignments on this tile.

Under the hood, this calls a refactored `check_collisions()` function, which is the same function now called during fiber assignment, `ok_to_assign`, so the same collision checks can be performed for potential targets as during fiber assignment.

### DIFF
--- a/collide.py
+++ b/collide.py
@@ -1,0 +1,129 @@
+import numpy as np
+
+from astropy.table import Table
+
+from fiberassign.hardware import load_hardware
+from fiberassign.tiles import load_tiles
+from fiberassign.targets import Targets, TargetsAvailable, LocationsAvailable, create_tagalong, load_target_file, targets_in_tiles
+from fiberassign.assign import Assignment
+
+from fiberassign.utils import Logger
+
+import fitsio
+
+def main():
+
+    # from LSS.mkCat_singletile.fa4lsscat import getfatiles
+    # getfatiles()
+    # return
+    log = Logger.get()
+
+    margins = dict(pos=0.05,
+                   petal=0.4,
+                   gfa=0.4)
+
+    hw = load_hardware(rundate=None, add_margins=margins)
+
+    t = Table.read('/global/cfs/cdirs/desi/survey/catalogs/Y1/LSS/tiles-DARK.fits')
+    print('tiles:', t)
+
+    tile = 1230
+    ts = '%06i' % tile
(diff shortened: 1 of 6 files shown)


dstndstn:
PS, this could potentially make fiber-assignment slower, because previously `ok_to_assign` would return `false` upon finding the first conflict, while the new version returns the bitmask (ie keeps looking upon collisions).  But I don't expect that to be a major effect.

ashleyjross:
@dstndstn do we want to ask anyone in particular to review this before merging?

For anyone reading, we would really like these functions for LSS catalogs.

araichoor:
I m pretty sure it s fine, but I can have a quick look at that this week: the "usual" check I d do would just be to ensure that this is backwards-compatible.
and I can do a small benchmark on the computation speed.

dstndstn:
That would be great, Anand!  Thank you!

ashleyjross:
@araichoor were you able to do your check?

araichoor:
sorry, not yet; will do today, promised!

araichoor:
ok, I ve run a check on reproducibility, and it looks good.
besides, the timing looks ~unchanged (details below).
and I m ~incompetent to review the cpp code, I trust @dstndstn there!

@dstndstn: could you just update the `doc/changes.rst` file?
(I m now trying to have this habit to update this file with each PR, rather than only when creating a tag).
then I can merge it.

details:
I ve run with the main code and with this branch code 300 randomly picked main tiles (100 dark, 100 bright, 100 backup).
I ve checked with `fits.HeaderDiff` and `fits.TableDataDiff` that the 0-th header and all the table extensions are unchanged (apart from few header keyword which are expected to change).
and here s a plot comparing the timings of the call to `run_assign_full`+`merge_results`, showing it s similar:
timings
data are here: https://data.desi.lbl.gov/desi/users/raichoor/fiberassign_dev/dstn-check-collisions/.

dstndstn:
Okay, updates the change log, and had a final look, I think this is good to merge!

araichoor:
awesome, thanks!

PR #457, opened 2023-07-28
Do chmod to _really_ get permissions set.

schlafly:
This adds a couple lines calling os.chmod(...) after os.makedirs(..., mode=...) to really get permissions set.  Apparently os.makedirs(..., mode=...) applies the umask on top of the selected mode, while os.chmod(...) is explicit.


araichoor:
@schlafly: sorry if I ve left that PR in the limbos..
looks good, I m going to merge it, but could you before that update the `docs/changes.rst` file in your branch with mentioning this PR?

schlafly:
I've updated the change log, thanks!

PR #461, opened 2024-02-26
Write speedups for multi-tiles runs

araichoor:
This PR adds a couple of -- simple but efficient -- "tricks" to speed-up the individual `fba-TILEID.fits` file writing:
- allow parallelisation in `write_assignment_fits()`: when N>1 tiles are run at once, those are sequentially written in the current main;
- add fast_match argument in TargetTagalong class: this uses the `desitarget.geomask.match()` function, instead of building a dictionary with a `TARGETID` for each key, which takes a long time when dealing with 10M+ targets.

Those two "tricks" are controlled by two new arguments in `parse_assign()` in `fiberassign/scripts/assign.py`:
- `--write_fits_numproc (default=0)`: number of jobs to be run in parallel when calling `write_assignment_fits()`;
- `--fast_match (default=False)`.

The use case is when one runs N>>1 tiles at once with tens of millions of targets (I developed that to explore the expected fiberassign when the GD1 tiles will be added to the Bright program).
The default values are disabling the "tricks", so that it is backwards-compatible, and this will change nothing in operations.

Test case with the GD1 tiles:
This case had ~800 tiles, and 7M targets plus 43M sky targets, so 50M targets.
The actual part of running the fiberassign is reasonably fast (1h), but the writing part was not.
Here are few timings to write the ~800 `fba-TILEID.fits` files (ie once the fiberassign is computed):
- current main: 6h30;
- with using `--fast_match True --write_fits_numproc 0`:  1h30.
- with using `--fast_match True --write_fits_numproc 64`: 13min
- with using `--fast_match True --write_fits_numproc 128`: 10min (no significant improvement w.r.t. 64 processes).

Note about the `fast_match` trick:
It assumes that the input `TARGETID` are unique.
If that is not the case, results will be wrong (ie the column values will be wrongly filled); but that s probably also the case for the current main... (I don t have clear, in-depth, understanding of the process).
As the goal is to look for speed, no checks are done (uniqueness of input the `TARGETID`, correctness of the matching).
Note to myself: a sanity check is planned to be implemented in `desitarget.geomask.match()`: https://github.com/desihub/desitarget/issues/811; once that is done, I will have to make sure that such check is disable in the call here, to preserve the speed-up.

For sanity, I've rerun 1% of the main tiles (125 backup, bright, dark tiles) with the current main and with turning on the `--fast_match` option, and I get no difference in the output data.

Note about the `write_fits_numproc` trick:
I had to use `multiprocessing.pool.ThreadPool` instead of the "usual" `multiprocessing.Pool`, because the latter was crashing with the Assignment() object `asgn`.
I found this workaround with a quick google search and it worked.

@dstndstn: in case you have time to have a look, I'd appreciate! but there is no rush.

### DIFF
--- a/py/fiberassign/assign.py
+++ b/py/fiberassign/assign.py
@@ -20,6 +20,7 @@ from astropy.coordinates import SkyCoord
 
 import multiprocessing as mp
 from multiprocessing.sharedctypes import RawArray
+from multiprocessing.pool import ThreadPool
 
 from functools import partial
 
@@ -573,7 +574,7 @@ def write_assignment_fits_tile(asgn, tagalong, fulltarget, overwrite, params):
 def write_assignment_fits(tiles, tagalong, asgn, out_dir=".", out_prefix="fba-",
                           split_dir=False, all_targets=False,
                           gfa_targets=None, overwrite=False, stucksky=None,
-                          tile_xy_cs5=None):
+                          tile_xy_cs5=None, numproc=0):
     """Write out assignment results in FITS format.
 
     For each tile, all available targets (not only the assigned targets) and
@@ -593,6 +594,7 @@ def write_assignment_fits(tiles, tagalong, asgn, out_dir=".", out_prefix="fba-",
             properties of assigned targets.
         gfa_targets (list of numpy arrays): Include these as GFA_TARGETS HDUs
         overwrite (bool): overwrite pre-existing output files
+        numproc (int): if >0, runs with numproc parallel jobs (with pool.starmap()); default: 0
 
     Returns:
         None
@@ -610,8 +612,12 @@ def write_assignment_fits(tiles, tagalong, asgn, out_dir=".", out_prefix="fba-",
     tiletime = tiles.obstime
     tileha = tiles.obshourang
 
+
     write_tile = partial(write_assignment_fits_tile,
                          asgn, tagalong, all_targets, overwrite)
+
+    if numproc > 0:
+        all_params = []
 
     for i, tid in enumerate(tileids):
         tra = tilera[tileorder[tid]]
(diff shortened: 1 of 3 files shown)


araichoor:
thanks for the prompt comments!

let me already answer about ThreadPool:
sorry I didn t provide the example of the crash.
here s what I get if I use Pool on a simple test case (i.e. if I use `mp.Pool(numproc)` instead of `ThreadPool(numproc)`):
```
  File "/global/cfs/cdirs/desicollab/users/raichoor/fiberassign-designs/fiberassign-gd1-scnd/test/gd1.py", line 401, in <module>
    main()
  File "/global/cfs/cdirs/desicollab/users/raichoor/fiberassign-designs/fiberassign-gd1-scnd/test/gd1.py", line 382, in main
    run(fbadir, tilesfn, targfn, skiesfn, rundate, args.numproc)
  File "/global/cfs/cdirs/desicollab/users/raichoor/fiberassign-designs/fiberassign-gd1-scnd/test/gd1.py", line 259, in run
    run_assign_full(ag)
  File "/global/homes/r/raichoor/software_dev/fiberassign_write_speedups/py/fiberassign/scripts/assign.py", line 451, in run_assign_full
    write_assignment_fits(tiles, tagalong, asgn, out_dir=args.dir,
  File "/global/homes/r/raichoor/software_dev/fiberassign_write_speedups/py/fiberassign/assign.py", line 655, in write_assignment_fits
    _ = pool.starmap(write_assignment_fits_tile, all_params)
  File "/global/common/software/desi/perlmutter/desiconda/20230111-2.1.0/conda/lib/python3.10/multiprocessing/pool.py", line 375, in starmap
    return self._map_async(func, iterable, starmapstar, chunksize).get()
  File "/global/common/software/desi/perlmutter/desiconda/20230111-2.1.0/conda/lib/python3.10/multiprocessing/pool.py", line 774, in get
    raise self._value
  File "/global/common/software/desi/perlmutter/desiconda/20230111-2.1.0/conda/lib/python3.10/multiprocessing/pool.py", line 540, in _handle_tasks
    put(task)
  File "/global/common/software/desi/perlmutter/desiconda/20230111-2.1.0/conda/lib/python3.10/multiprocessing/connection.py", line 211, in send
    self._send_bytes(_ForkingPickler.dumps(obj))
  File "/global/common/software/desi/perlmutter/desiconda/20230111-2.1.0/conda/lib/python3.10/multiprocessing/reduction.py", line 51, in dumps
    cls(buf, protocol).dump(obj)
TypeError: cannot pickle 'fiberassign._internal.Assignment' object
```

PR #464, opened 2025-01-15
Remove DesiTest

weaverba137:
This PR removes `DesiTest` as requested in desihub/desiutil#215.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -63,7 +63,7 @@ jobs:
             - name: Compile the package
               run: CC=g++ CXX=g++ CFLAGS="-fPIC" LDSHARED="g++ -shared" python setup.py build_ext --inplace
             - name: Run Tests
-              run: pytest --ignore=py/fiberassign/test/simulate.py
+              run: pytest
 
     docs:
         name: Doc test
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,8 +16,8 @@ jobs:
             fail-fast: true
             matrix:
                 os: [ubuntu-22.04]  # , ubuntu-24.04]
-                python-version: ['3.10', '3.11']  # , '3.12']
-                numpy-version: ['<1.23', '<2']
+                python-version: ['3.10']  # '3.11', '3.12']
+                numpy-version: ['<1.23']
                 astropy-version: ['<6.1', '<7']
 
         env:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,7 @@ jobs:
                 astropy-version: ['<6.1', '<7']
 
         env:
-            DESIUTIL_VERSION: 3.5.0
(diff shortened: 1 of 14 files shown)


weaverba137:
Tests are failing due to impossible dependency resolutions, so the test matrix will also need to be updated.

weaverba137:
It looks like it is not possible to compile `fiberassign` on `ubuntu-24.04`, or with Python > 3.10. I think that's beyond the scope of this ticket. I think this is ready for review.

araichoor:
hi @weaverba137, @sbailey 

just checking if this PR is going to be merged soon-ish (checks from my PR are failing because of the issue solved by this PR, I think: https://github.com/desihub/fiberassign/pull/466/checks).
sorry I cannot help with reviewing this one..
thanks!

sbailey:
@weaverba137 this PR removes DesiTest, thus fully dropping support for `python setup.py test` even with older versions of desiutil, but I see that in `.github/workflows/test.yml` you run the tests with
```
python -c 'import fiberassign.test; fiberassign.test.runtests()'
```
instead of `pytest`.  Why is that?  I haven't been able to check if pytest works without NERSC this week because I haven't been able to compile fiberassign on my laptop setup (some issue with pybind11, still investigating).

We could probably merge this anyway, but I'd like to understand whether pytest works and document any gotchas here.

weaverba137:
That is a great question. I will have to do a deep dive into the repository history. For example, the test may have been copied from a Travis test written by someone else.

weaverba137:
The `test.yml` file was added in ead086d, committed by @tskisner, and the test was copied from the `.travis.yml` file that was removed in the same commit.

The test was *changed* in c7ca950, which has the comment "Ensure that we run travis tests with the installed copy of the package."

So I think what's happening here is that `python setup.py test` would not necessarily test the *compiled* version of `fiberassign`, while compiling, installing and *then* running `fiberassign.test.runtests()` definitely would.

sbailey:
Thanks for checking.  Please try to confirm that pytest works for fiberassign, which I think means running
```
python setup.py build_ext --inplace
pytest
```
I'm currently stuck on the first step on my laptop, but it hopefully would work within the github actions container or other places that you might have access to.  If that becomes onerous, we could also probably merge now and revisit pytest next week when NERSC is back, and separately deal with how to in-place compile fiberassign at other locations.

weaverba137:
`pytest` is finding tests in `fiberassign/old`. `pytest` can be configured to ignore that, but perhaps it's time to think about removing unused code.

weaverba137:
Ugh, pytest is interpreting a particular function inside `test/simulate.py` as a pytest fixture, even though the file is explicitly ignored. However, the function in question, `fiberassign.test.simulate.test_subdir_create()` sure looks a lot like a pytest fixture and is imported into other test modules. The function will have to be renamed.

sbailey (approved):
Looks good, thanks for the work to get bare pytest working.

PR #466, opened 2025-01-28
Field rotation updates to avoid hexapod rotation limit errors

araichoor:
This PR is in-development, but I share now, so that we can discuss choices I made.
Once perlmutter is back up, I ll run more sanity checks.

The goal of this PR is to avoid the hexpod rotation limit errors that happen in operations.
Details are presented on DocDB-8931.
In short, for a given tile:
- fiberassign computes `FIELDROT = (RA, DEC, OBSDATE)`, the field rotation (using desimodel routines); so far, `OBSDATE=2022-07-01`, the "mid-survey" date which was defined years ago;
- then PlateMaker computes `PM_REQ_FIELDROT`, the proper field rotation, based on the exact date of observation and HA;
- the sent requested rotation for the hexapod is `ROTOFFST = PM_REQ_FIELDROT - FIELDROT`;
- if `abs(ROTOFFST) > 600 arcsec`, this triggers an hexapod rotation limit error, and we move to another tile.

The approach to solve that is twofold:
(1) change the `OBSDATE` (from `2022-07-01` to `when_we_observe + 1 yr`);
(2) apply a correction on the computed `FIELDROT`.
Based on DocDB-8931, those two actions should narrows the `ROTOFFST` distribution, keeping it within [-400 arcsec, +400 arcsec]; at least for the main survey observations.

Note that there is a subtlety that, during the observation, this rotation is updated every minute; though, hopefully, such additional rotation should be in total smaller than 30-60 arcsec for a 30min exposure, so the system should stay away from the 600 arcsec limit.

I ll discuss the implementation choices in a separate message.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -6,9 +6,11 @@ fiberassign change log
 5.7.3 (unreleased)
 ------------------
 
+* Field rotation updates to avoid hexapod rotation limit errors (PR `#466`_).
 * Fix C++ bug with use of std::abs and improve floating point reproducibility (PR `#470`_).
 * Remove ``DesiTest`` from setup.py and warn about other deprecated features (PR `#464`_).
 
+.. _`#466`: https://github.com/desihub/fiberassign/pull/466
 .. _`#470`: https://github.com/desihub/fiberassign/pull/470
 .. _`#464`: https://github.com/desihub/fiberassign/pull/464
 
--- a/py/fiberassign/data/cutoff-dates.yaml
+++ b/py/fiberassign/data/cutoff-dates.yaml
@@ -1,3 +1,4 @@
 rundate:
     std_wd: "2021-12-01T19:00:00+00:00" # std_wd not counted in per-petal-std after that rundate
     etc_stuck: "2022-04-28T19:00:00+00:00" # etc fibers are treated as stuck fibers after this date
+    obsdate: "2025-03-01T19:00:00+00:00" # obsdate set as rundate + 1yr
--- a/py/fiberassign/data/cutoff-dates.yaml
+++ b/py/fiberassign/data/cutoff-dates.yaml
@@ -1,4 +1,4 @@
 rundate:
     std_wd: "2021-12-01T19:00:00+00:00" # std_wd not counted in per-petal-std after that rundate
     etc_stuck: "2022-04-28T19:00:00+00:00" # etc fibers are treated as stuck fibers after this date
-    obsdate: "2025-03-01T19:00:00+00:00" # obsdate set as rundate + 1yr
+    obsdate: "2025-05-02T19:00:00+00:00" # obsdate set as rundate + 1yr after this date
--- a/py/fiberassign/scripts/assign.py
+++ b/py/fiberassign/scripts/assign.py
@@ -112,7 +112,8 @@ def parse_assign(optlist=None):
     parser.add_argument("--fieldrot_corr", type=bool, required=False,
                         default=None,
                         help="apply correction to computed fieldrot "
-                        "(default: False if rundate<rundate_cutoff, True else")
+                        "(default: False if args.fieldrot is provided or "
+                        "if args.rundate<rundate_cutoff, True else")
 
     parser.add_argument("--dir", type=str, required=False, default=None,
                         help="Output directory.")
@@ -280,10 +281,13 @@ def parse_assign(optlist=None):
         args.obsdate = '{}-{}-{}'.format(year, mm, dd)
 
     # Apply the field rotation correction?
-    # - True if rundate!=None and rundate>=rundate_cutoff
+    # - True if fieldrot=None, rundate!=None and rundate>=rundate_cutoff
     # - False else
     if args.fieldrot_corr is None:
-        _, args.fieldrot_corr = get_obsdate(rundate=args.rundate)
+        if args.fieldrot is None:
+            _, args.fieldrot_corr = get_obsdate(rundate=args.rundate)
+        else:
+            args.fieldrot_corr = False
 
     # Set output directory
     if args.dir is None:
--- a/py/fiberassign/scripts/assign.py
+++ b/py/fiberassign/scripts/assign.py
@@ -14,7 +14,7 @@ import sys
 import argparse
 import re
 
-from ..utils import GlobalTimers, Logger
+from ..utils import GlobalTimers, Logger, get_default_static_obsdate, get_obsdate
 
 from ..hardware import load_hardware, get_default_exclusion_margins
 
@@ -95,9 +95,12 @@ def parse_assign(optlist=None):
                         "Default uses the current date.  Format is "
                         "YYYY-MM-DDTHH:mm:ss+-zz:zz.")
 
-    parser.add_argument("--obsdate", type=str, required=False, default="2022-07-01",
+    parser.add_argument("--obsdate", type=str, required=False,
+                        default=None,
                         help="Plan field rotations for this date (YEARMMDD, "
-                        "or ISO 8601 YEAR-MM-DD with or without time).")
+                        "or ISO 8601 YEAR-MM-DD with or without time); "
+                        "default={} or rundate+1yr, according to rundate".
+                        format(get_default_static_obsdate()))
 
     parser.add_argument("--ha", type=float, required=False, default=0.,
                         help="Design for the given Hour Angle in degrees.")
@@ -259,6 +262,10 @@ def parse_assign(optlist=None):
             except ValueError:
                 args.excludemask = desi_mask.mask(args.excludemask.replace(",","|"))
 
+    # Set obsdate
+    if args.obsdate is None:
+        args.obsdate = get_obsdate(rundate=args.rundate)
+
     # convert YEARMMDD to YEAR-MM-DD to be ISO 8601 compatible
     if re.match('\d{8}', args.obsdate):
         year = args.obsdate[0:4]
@@ -278,7 +285,13 @@ def parse_assign(optlist=None):
     args.margins = dict(pos=args.margin_pos,
                         petal=args.margin_petal,
                         gfa=args.margin_gfa)
+
+    # Print all args
+    for kwargs in args._get_kwargs():
+        log.info("{}:\t{}".format(kwargs[0], kwargs[1]))
+
     return args
+
 
 def run_assign_init(args, plate_radec=True):
     """Initialize assignment inputs.
--- a/py/fiberassign/scripts/assign.py
+++ b/py/fiberassign/scripts/assign.py
@@ -109,6 +109,11 @@ def parse_assign(optlist=None):
                         help="Override obsdate and use this field rotation "
                         "for all tiles (degrees counter clockwise in CS5)")
 
+    parser.add_argument("--fieldrot_corr", type=bool, required=False,
+                        default=None,
+                        help="apply correction to computed fieldrot "
+                        "(default: False if rundate<rundate_cutoff, True else")
+
     parser.add_argument("--dir", type=str, required=False, default=None,
                         help="Output directory.")
 
@@ -264,7 +269,7 @@ def parse_assign(optlist=None):
 
     # Set obsdate
     if args.obsdate is None:
-        args.obsdate = get_obsdate(rundate=args.rundate)
+        args.obsdate, _ = get_obsdate(rundate=args.rundate)
 
     # convert YEARMMDD to YEAR-MM-DD to be ISO 8601 compatible
     if re.match('\d{8}', args.obsdate):
@@ -273,6 +278,12 @@ def parse_assign(optlist=None):
         dd = args.obsdate[6:8]
         #- Note: ISO8601 does not require time portion
         args.obsdate = '{}-{}-{}'.format(year, mm, dd)
+
+    # Apply the field rotation correction?
+    # - True if rundate!=None and rundate>=rundate_cutoff
+    # - False else
+    if args.fieldrot_corr is None:
+        _, args.fieldrot_corr = get_obsdate(rundate=args.rundate)
 
     # Set output directory
     if args.dir is None:
@@ -324,7 +335,8 @@ def run_assign_init(args, plate_radec=True):
                 except ValueError:
                     pass
     tiles = load_tiles(tiles_file=args.footprint, select=tileselect,
-        obstime=args.obsdate, obstheta=args.fieldrot, obsha=args.ha)
+        obstime=args.obsdate, obstheta=args.fieldrot, obsha=args.ha,
+        obsthetacorr=args.fieldrot_corr)
 
     # Before doing significant calculations, check for pre-existing files
     if not args.overwrite:
--- a/py/fiberassign/tiles.py
+++ b/py/fiberassign/tiles.py
@@ -22,6 +22,8 @@ from desimodel.focalplane.fieldrot import field_rotation_angle
 
 import astropy.time
 
+from fiberassign.utils import get_default_static_obsdate
+
 from ._internal import Tiles
 
 
@@ -71,7 +73,7 @@ def load_tiles(tiles_file=None, select=None, obstime=None, obstheta=None,
         obsdatestr = [x.isot for x in obsdate]
     else:
         # default to middle of the survey
-        obsdate = astropy.time.Time('2022-07-01')
+        obsdate = astropy.time.Time(get_default_static_obsdate())
         obsmjd = [obsdate.mjd,] * len(keeprows)
         obsdatestr = [obsdate.isot, ] * len(keeprows)
 
--- a/py/fiberassign/tiles.py
+++ b/py/fiberassign/tiles.py
@@ -22,13 +22,13 @@ from desimodel.focalplane.fieldrot import field_rotation_angle
 
 import astropy.time
 
-from fiberassign.utils import get_default_static_obsdate
+from fiberassign.utils import get_default_static_obsdate, get_obstheta_corr
 
 from ._internal import Tiles
 
 
 def load_tiles(tiles_file=None, select=None, obstime=None, obstheta=None,
-               obsha=None):
+               obsha=None, obsthetacorr=False):
     """Load tiles from a file.
 
     Load tile data either from the specified file or from the default provided
@@ -42,10 +42,16 @@ def load_tiles(tiles_file=None, select=None, obstime=None, obstheta=None,
         obstheta (float):  The angle in degrees to override the field rotation
             of all tiles.
         obsha (float): The Hour Angle in degrees to design the observation of all tiles.
+        obsthetacorr (optional, defaults to False): apply a correction on obstheta,
+            based on DocDB-8931 (bool)
 
     Returns:
         (Tiles):  A Tiles object.
 
+    Notes:
+        2025, Feb.: added the obsthetacorr optional argument; this correction
+            implemented to avoid the hexapod rotation limit error, is
+            presented/described in DocDB-8931.
     """
     # Read in the tile information
     if tiles_file is None:
@@ -98,6 +104,15 @@ def load_tiles(tiles_file=None, select=None, obstime=None, obstheta=None,
     ha_obs = np.zeros(len(keeprows), dtype=np.float64)
     if obsha is not None:
         ha_obs[:] = obsha
+
+    # AR apply correction on the obstheta?
+    # AR sign is "minus" here;
+    # AR see get_obstheta_corr() (and p.5 of DocDB-8931)
+    if obsthetacorr:
+        theta_obs -= get_obstheta_corr(
+            tiles_data["DEC"][keeprows],
+            ha_obs,
+        )
 
     tls = Tiles(tiles_data["TILEID"][keeprows], tiles_data["RA"][keeprows],
                 tiles_data["DEC"][keeprows],
--- a/py/fiberassign/utils.py
+++ b/py/fiberassign/utils.py
@@ -145,28 +145,35 @@ def get_obsdate(rundate=None):
 
     Returns:
         obsdate: "YYYY-MM-DD" format (string)
+        is_after_cutoff: is rundate after rundate_cutoff? (bool)
     """
     obsdate = get_default_static_obsdate()
+    is_after_cutoff = False
     if rundate is None:
-        log.info("rundate={} -> setting obsdate={}".format(rundate, obsdate))
+        log.info(
+            "rundate={} -> (obsdate, is_after_cutoff)=({}, {})".format(
+                rundate, obsdate, is_after_cutoff
+            )
+        )
     else:
         assert_isoformat_utc(rundate)
         rundate_mjd = Time(datetime.strptime(rundate, "%Y-%m-%dT%H:%M:%S%z")).mjd
         rundate_cutoff = get_date_cutoff("rundate", "obsdate")
         rundate_mjd_cutoff = Time(datetime.strptime(rundate_cutoff, "%Y-%m-%dT%H:%M:%S%z")).mjd
         if rundate_mjd >= rundate_mjd_cutoff:
+            is_after_cutoff = True
             yyyy = int(rundate[:4])
             obsdate = "{}{}".format(yyyy + 1, rundate[4:10])
             log.info(
-                "rundate={} >= rundate_cutoff={} -> setting obsdate={}".format(
-                    rundate, rundate_cutoff, obsdate)
+                "rundate={} >= rundate_cutoff={} -> (obsdate, is_after_cutoff)=({}, {})".format(
+                    rundate, rundate_cutoff, obsdate, is_after_cutoff)
             )
         else:
             log.info(
-                "rundate={} < rundate_cutoff={} -> setting obsdate={}".format(
-                    rundate, rundate_cutoff, obsdate)
+                "rundate={} < rundate_cutoff={} -> (obsdate, is_after_cutoff)=({}, {})".format(
+                    rundate, rundate_cutoff, obsdate, is_after_cutoff)
             )
-    return obsdate
+    return obsdate, is_after_cutoff
 
   
 def get_svn_version(svn_dir):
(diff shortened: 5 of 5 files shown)


araichoor:
Tagging @schlafly @geordie666 @ashleyjross for the discussion.

First, if useful, I remind here the flow of function calling, in operations:
- `fba_launch` is called with some arguments, including `TILERA, TILEDEC, HA, RUNDATE`;
- then these arguments are propagated in the call to:
    - `fiberassign.scripts.assign.parse_assign()`
   - `fiberassign.scripts.assign.run_assign_full()`
- `run_assign_full()` calls `run_assign_init()` where the `fiberassign.tiles.load_tiles()` function is called, and where the field rotation is computed.

I ve chosen to tie by default the two actions to a given `cutoff_rundate` (dummy value for now):

https://github.com/desihub/fiberassign/blob/005eae3b863bc0c6bb12eb55c80440d30e9cc569/py/fiberassign/data/cutoff-dates.yaml#L4

I ve chosen the `rundate` quantity to track when the tile is designed, as it avoids to bring in another variable (like `nowtime`), and should be close-ish to the design date for any tiles we will design (in case of perlmutter outages, it could be different by few days, but that s fine here).

The current approach is that, if a `rundate` is provided and later than the `cutoff_rundate`, then:
- `OBSDATE` is set to `RUNDATE + 1 yr`;
- if `FIELDROT` is not provided as an argument, a correction to the field rotation is computed and applied to the computed`FIELDROT`.

On top of modifications in `parse_assign()`, I ve added one argument `--fieldrot_corr`; by default, this argument is set to `True` if `rundate >= cutoff_rundate` and no `--fieldrot` argument is provided.

I ll have to validate/test when perlmutter is back up, but if everything works as it should be, no change would be needed to the `fba_launch` invocation in `fba-main-onthefly.sh`.

And backwards reproducibility should be preserved.
For the LSS ALTMTL runs, if I m correct, no change should be needed: the current processing is by calling `fba_run` for each tile, with providing arguments from the fiberassign header file, including `FIELDROT` and `HA`:
https://github.com/desihub/LSS/blob/4bb4692866ae9305ad2f9f940f1388a594e54baf/py/LSS/SV3/fatools.py#L250-L272.
https://github.com/desihub/LSS/blob/4bb4692866ae9305ad2f9f940f1388a594e54baf/py/LSS/SV3/fatools.py#L444-L467
So the chosen default settings should:
- pick the same `OBSDATE` as for operations;
- not apply any correction on `FIELDROT` (as the header `FIELDROT` value already had that correction computed+applied).

Happy to have feedback / suggestions / etc!

araichoor:
for info:
with the help of @sbailey, I ve re-based from the main branch, pulling in the changes from https://github.com/desihub/fiberassign/pull/464.
now the tests pass, great!

in the very unlikely case that someone has checked out this `hexrot` branch and started to work on it between yesterday and now, just be careful of those changes.

schlafly:
I reviewed the code; it looks great.

Is it the case that you can design a pair of commissioning tiles with this PR, one using the "before" date and one using the "after" date, and try to observe both of these and compare the final hexapod rotations?

araichoor:
thanks for checking the code!
now perlmutter is back up, I ll run few sanity checks, to verify that it doesn t break backwards-comptability.

"_Is it the case that you can design ..._"
=>
I was initially thinking of designing one tile with the main branch, and one with this PR branch (and that tile could be a `specialbackup` tile, in the 8XXXXX).
but what you suggest should also work because:
- the "cutoff date" (Mar. 1) is in the future as of today;
- there s this feature that if `RUNDATE` is in the future, fiberassign will pick the latest focal plane state, so the same fp will be used in both cases.

if for instance I make two designs with `RUNDATE` set to `2025-02-28` and `2025-03-02`, both will pick the same focal plane state (the latest one as of today), and the `FIELDROT` correction will only be applied to `2025-03-02`; the only "drawback" is that those tiles won t be reproducible (because if we rerun after `2025-03-02`, the fp will likely be different than today), but probably we don t care for those test tiles.

schlafly:
I agree that the way you had planned makes sense to me too.  I also agree that reproducibility on this commissioning tile isn't super important.  But I think it would actually be reproducible in a narrow sense since somewhere in there we record the times of the last FP update, I think (where we read the ledger through)?

schlafly:
Hmm, thanks.  So you're saying we don't provide a quantity that is always "this is the date of the last entry we read in the MTL" and "this is the date of the last entry we read in the FP state ledger"?

araichoor:
the infos you mention are generated if no `--rundate` or no `--mtltime` arguments are provided to `fba_launch` (which is the usage in operations); in that case, the fiberassign code will go parse the files, figure out the latest dates, pick those + print infos in the log.

but in the case where e.g. `--rundate XXX` is provided, `fba_launch` doesn t go through this parsing files step, and just uses the input argument.

e.g. you can run in some working folder:
```
godesi main
export SKYBRICKS_DIR=$DESI_ROOT/data/target/skybricks/v3
fba_launch --outdir "./" --tileid 999000 --tilera 0. --tiledec 0. --survey main --program DARK --ha 0. --dtver 1.1.1 --rundate 2025-03-02T17:00:00+00:00 --mtltime 2025-03-02T17:00:00+00:00 --nosteps qa
```
and check the header of `999/fiberassign-999000.fits.gz` or look at `999/fiberassign-999000.log`.

that may be a shortcoming of the code, but that s what s currently in!
though, note that we record in the header the svn version of the `DESIMODEL` and `DESI_SURVEYOPS` folders, so, technically, one can reproduce those, if one checks out those versions...

schlafly:
Got it, thanks.  Yeah, I think it would have been better to store the times of the last ledger entries, but no big deal.

araichoor:
hi, back to this, sorry for the delay.

first, as a sanity check, I ve rerun 50% of the main bright/dark tiles with this PR, to assess this doesn t break backwards compatibility; out of 7.4k tiles, I only get 1 difference in potential assignments, which I ll blame on numerical precision, so all good (files here: `/pscratch/sd/r/raichoor/fiberassign-rerun-main/20250203-hexrot/`).

then, I ve generated the `FIELDROT` values we would have used with this PR for most of expids (I set for this test only cutoff `obsdate: "2025-03-01T19:00:00+00:00"`).

I select the 25,498 expids in `exposures-daily.csv` with:
```
        sel = d["NIGHT"] >= 20201214
        sel &= d["EFFTIME_SPEC"] > 0
        sel &= (d["TILERA"] != 0) | (d["TILEDEC"] != 0)
        sel &= d["TILEID"] != 82248 # no rundate..
```
File here:
```
-rw-rw---- 1 raichoor desi 14M Feb 10 12:12 /global/cfs/cdirs/desi/users/raichoor/hexrot-ha-dec/check-pr466/hexrot-check-pr.fits
```

Because of this slack message, I ve restricted my development work to `NIGHT >= 20220916`: https://desisurvey.slack.com/archives/C02KKL0PBA7/p1733296336897999?thread_ts=1733278555.592099&cid=C02KKL0PBA7.

For those `NIGHT >= 20220916`, things look fine:
- 35 exposures, previously close to that threshold, now have a safer `PR_ROTOFFST`;
- only one expid getting `PR_ROTOFFST` close to the +/- 600 arcsec limit.
hexrot-check-pr-post20220916
it s a backup tile at dec=79:
```
 NIGHT   EXPID  TILEID  FAFLAVOR  TILERA TILEDEC AIRMASS FA_HA ORIG_FIELDROT PR_FIELDROT ORIG_ROTOFFST PR_ROTOFFST
-------- ------ ------ ---------- ------ ------- ------- ----- ------------- ----------- ------------- -----------
20221108 152396  42996 mainbackup 24.871  79.373   1.586   0.0       0.28966     0.37764         251.6       568.3
```

But for the `NIGHT < 20220916` exposures, things are a bit less clear to me:
- 6 exposures, previously close to that threshold, now have a safer `PR_ROTOFFST`;
- but 11 exposures, are close/further than the +/- 600 arcsec limit.
hexrot-check-pr-pre20220916
those are a mix of sv1 and main tiles:
```
 NIGHT   EXPID  TILEID  FAFLAVOR   TILERA TILEDEC AIRMASS FA_HA  ORIG_FIELDROT PR_FIELDROT ORIG_ROTOFFST PR_ROTOFFST
-------- ------ ------ ---------- ------- ------- ------- ------ ------------- ----------- ------------- -----------
20210105  70970  80650  sv1bgsmws   135.0    83.0   1.652    0.0       0.71438     0.77033         352.5       553.9
20210223  77777  80782 sv1backup1   168.8    71.5   1.298    0.0       0.07506     0.11146         461.9       592.9
20210306  79570  80694  sv1lrgqso   162.0    83.0   1.661    0.0       0.30925     0.38879         228.9       515.2
20210306  79571  80694  sv1lrgqso   162.0    83.0   1.675    0.0       0.30925     0.38879         243.5       529.8
20210425  86392  80955 sv1backup1   195.0    80.0   1.626    0.0      -0.18583    -0.10937         250.6       525.9
20211218 114633  41757 mainbackup 123.517  -13.39   1.437    0.0       0.10697     0.06439        -367.4      -520.7
20220210 122045  24423 mainbright 113.715   77.37   1.483 -23.97       0.52074     0.60184         221.2       513.2
20220212 122286  23122 mainbright 147.498  -5.374   1.489 -15.38        0.0672     0.03288        -394.1      -517.7
20220213 122403  20259 mainbright 159.568  -8.842   1.686 -11.41       0.04388     0.00609        -502.2      -638.2
20220313 126045  42640 mainbackup 121.371  78.937   1.469    0.0       0.55208     0.63079         232.1       515.5
20220313 126056  40524 mainbackup 162.088 -15.046   1.538    0.0       0.03952    -0.00463        -418.2      -577.1
```

I don t know to what extent I shall worry about those; as for instance it could be that the PlateMaker modelling was different then.

araichoor:
ok great, then we re "good to go" for designing the test tiles, right?

I have few questions on the choices to make (sorry!):
- first what scheme: we could go with what you suggested, ie pairs of the same tile designed with that fiberassign branch, with different rundates (which would trigger or not the correction), I m fine with that
- how many different sky locations do we want to test? the problematic region is this ra,dec\~150,70 region; shall we pick only one location there? (e.g from some tile that triggered the hexapod error) or more? would we want to also test e.g. ra,dec\~150,-10?
- what HA values shall be picked? e.g. shall I pick a (ra, dec, designha) from tiles-main.ecsv?
- we don't care about what targets are observed, right?
- shall that be a `survey=special` tertiary program? or e.g. `survey=sv2` (which was used for operations tests); provided that I made the PlateMaker test a `tertiary46` program, maybe keep with tertiary?
- else?

araichoor:
thanks.
the simplest on my side is to just make some `fba_launch` call, with e.g. picking main/backup targets, and set in the header `survey=special`, `faprgrm=backup`.
the only "worry" is that the TARGETIDs will be the "real" ones; but maybe that s ok.

if those are somewhat problematic for the data team, we could ask them to set the exposures as "bad", so that they don t appear in any future release...

I ve run here two tiles at ra,dec=113,77: `/global/cfs/cdirs/desi/users/raichoor/fiberassign-designs/fiberassign-hexrot/v0`.
that location corresponds to tileid=5423, which failed for expid=275640, night=20250125.

the first one has a rundate before the cutoff date (ie with no correction):
```
fitsheader -e 0 -k RUNDATE -k FA_PLAN -k FIELDROT 999/fiberassign-999997.fits.gz
# HDU 0 in 999/fiberassign-999997.fits.gz:
RUNDATE = '2025-02-01T19:00:00+00:00'                                           
FA_PLAN = '2022-07-01T00:00:00.000'                                             
FIELDROT=    0.509027994150231                                                  
```
the second one after the cutoff date (ie with the correction):
```
fitsheader -e 0 -k RUNDATE -k FA_PLAN -k FIELDROT 999/fiberassign-9999978fits.gz
# HDU 0 in 999/fiberassign-999998.fits.gz:
RUNDATE = '2025-04-01T19:00:00+00:00'                                           
FA_PLAN = '2026-04-01T00:00:00.000'                                             
FIELDROT=    0.651400057262175                                                  
```

if useful here are the calling commands (not copying here the commands to set the environment):
```
OUTDIR=$PWD

SURVEY=main
PROGRAM=BACKUP
DTCATVER=2.2.0
HDR_SURVEY=special
HDR_FAPRGRM=backup

# ra dec of tileid=5423, which failed for expid=275640, night=20250125
RA=113
DEC=77
HA=0

# before cutoff date
TILEID=999997
RUNDATE="2025-02-01T19:00:00+00:00"
CMD="fba_launch --outdir $OUTDIR --tileid $TILEID --tilera $RA --tiledec $DEC --survey $SURVEY --program $PROGRAM --ha $HA --dtver $DTCATVER --rundate $RUNDATE --hdr_survey $HDR_SURVEY --hdr_faprgrm $HDR_FAPRGRM"
echo $CMD
eval $CMD

# after cutoff date
TILEID=999998
RUNDATE="2025-04-01T19:00:00+00:00"
CMD="fba_launch --outdir $OUTDIR --tileid $TILEID --tilera $RA --tiledec $DEC --survey $SURVEY --program $PROGRAM --ha $HA --dtver $DTCATVER --rundate $RUNDATE --hdr_survey $HDR_SURVEY --hdr_faprgrm $HDR_FAPRGRM"
echo $CMD
eval $CMD
```

is it ok?

geordie666:
@araichoor: I think this plan looks good.

I'm not worried about the `TARGETID` issue as I think we _absolutely_ have to ensure all of these exposures are set as bad anyway — because we really don't want tiles showing up in a release that were run with bespoke versions of fiberassign.

araichoor:
ok perfect, thanks for the feedback!
I ll proceed with "official" designs; I ll do that in this folder: `$DESI_ROOT/survey/fiberassign/special/20250214`.

araichoor:
@schlafly:  the two test tiles have been observed tonight (20250219):
- expid=280217 is tileid=83575, designed without the correction (rundate before the cutoff-date);
- expid=280218 is tileid=83576, designed with the correction (rundate after the cutoff-date).

here are the `ROTOFFST` values:
```
fitsheader -e SPEC -k ROTOFFST /global/cfs/cdirs/desi/spectro/data/20250219/00280217/desi-00280217.fits.fz 
# HDU SPEC in /global/cfs/cdirs/desi/spectro/data/20250219/00280217/desi-00280217.fits.fz:
ROTOFFST=               -323.4 / [arcsec] Rotator offset                        

fitsheader -e SPEC -k ROTOFFST /global/cfs/cdirs/desi/spectro/data/20250219/00280218/desi-00280218.fits.fz 
# HDU SPEC in /global/cfs/cdirs/desi/spectro/data/20250219/00280218/desi-00280218.fits.fz:
ROTOFFST=                 196. / [arcsec] Rotator offset                        
```
and here is how this compares with previous `NIGHT>=20220916` data (for which I ve the original `ROTOFFST` and the one I d expect with this PR):
Screenshot 2025-02-19 at 10 42 43 PM

I am not super-sure of how to interpret the result: I would say that this ok, ie it falls in a region where I expect to find data.
provided that the goal of this test is to verify that the signs in the `FIELDROT` correction are right, we can call this a "half-success", I guess.

in this plot I highlight few things:
- red lines are for the two test tiles;
- I highlight (green) TILEID=42629, which has several exposures falling close to the two test tiles on the right panel; that TILEID has `FA_HA=0`;
- hence I also highlight (yellow) all tiles with `FA_HA=0`, which are either `mainbackup` or `tertiary` tiles.

are we happy with this test?
or would we want to perform another test (e.g. at a different location, or same location with a non-zero `FA_HA`)?
note that those are not too costly observationally: it costed exptime=10min overall, in dark time (expid=280217 started at 22:23, and the next dark tile started at 22:33).
conditions were "dark", survey speed \~85%, `EFFTIME_ETC~35s` (we requested 30s in the sequence).

schlafly:
For me the key is whether we got the sign right.  I think you're saying that the sign was indeed correct, and so I would go ahead and merge?

I hear you that you're calling this a half success because your model isn't terribly good there, possibly because we didn't actually observe the tile at HA = 0, where we planned it.  But I think you've already looked at a lot of data to test the overall trends and we only wanted to confirm that we had the sign of the correction right.

araichoor:
ok, so let s say it s a success then :)
I ll merge tomorrow.

for completeness, let me provide here an updated version of the figure I shared yesterday:
- left plot: yesterday, I wronlgy included the `NIGHT<20220916` data in the `FA_HA=0` sample; now it s removed;
- both plots: I add another subsample (magenta), which are the `FA_HA=0 & DEC>=77` exposures => one sees that those populate the ~same region as our two test tiles, that s kind of re-assuring.
 
Screenshot 2025-02-20 at 3 29 17 PM

schlafly:
I agree that I would hold until we know we can get it in and update the code, or at least be prepared to do a separate PR to update that date if we need to.  @sybenzvi , how long after making a new tag do you think we would need to get something installed at KP?  It needs to go in the datasystems environment.  We could once do that independently of Klaus and Ann, but I don't know if that's now the case.

sybenzvi:
@schlafly, making an update to fiberassign at KPNO should be pretty fast. Compared to the pre-firewall days, it's now a two-step process where we download the code to one machine and then "git clone" it from there to the cluster using `datasystems`. It's not a huge headache; Klaus and Ann showed me how to do it in mid-January and Julien was able to update `desispec` using the method a week and a half ago.

araichoor:
in case it s relevant: note that for fiberassign, there s an extra step, where you ve to build the c++ code.
the typical sequence I do (but I m sure there are people more expert than me here!):
```
git clone https://github.com/desihub/fiberassign
git pull
git checkout main
python setup.py build_ext --inplace
```

sybenzvi:
OK, I can try the build at KPNO (in a safe and isolated spot) to make sure I can get it working, and then when you're satisfied with the updates we can attempt a full upgrade.

schlafly:
Note that we won't use it until we update the NTS to load that version of fiberassign.  So I think you can install pretty safely.

sybenzvi:
@araichoor, after a delay due to access issues at KPNO, I was able to follow Eddie's instructions and run `fba-main-onthefly` using a newer desiconda stack and the `hexrot` branch of fiberassign. I generated ~20 tiles on the DESI cluster and copied them to NERSC. You can access them at
https://data.desi.lbl.gov/desi/users/sybenzvi/fa_holding_pen_20250305/

While this is a small set, it should be big enough to check if there is an obvious problem with reproducibility. Would you mind taking a look? If these files look OK, I'll try to make a greater variety of tiles in different programs so we can be confident in the update.

araichoor:
thanks @sybenzvi.
I ve rerun those 24 tiles at nersc with this `fiberassign/hexrot` branch (and the `godesi main` environment otherwise).
I do find one difference for a potential assignment:
```
cd /global/cfs/cdirs/desi/users/raichoor/fiberassign_dev/hexrot/replay_kpno_20250305/godesi_main
cat fba_nersc_kpno-diff.asc 
# TILEID	EXTENSION	STATUS	TARGETID	FIBER	TARGET_RA	TARGET_DEC	FA_TYPE	PRIORITY_INIT	PRIORITY	DESI_TARGET	MWS_TARGET	BGS_TARGET	SCND_TARGET
009383	FIBERASSIGN	miss	616087861451554913	856	84.47591443016104	-29.358519776650915	4	-1	-1	4294967296	0	0	0
009383	FIBERASSIGN	add	39627109148132500	856	84.44979509592072	-29.352143851237575	1	3000	3000	1179778	0	0	0
```
I m not sure to what extent that s problematic.

could you generate the same 24 tiles at kpno with the `fiberassign/main` branch?

schlafly:
Ugh.  Anand, you've run a lot of tiles at NERSC, so your expectation is that the mismatch has to do with the new desiconda environment and not with the hexrot branch?  In which case we expect that it's related to the issue that they were having with LSS and the wrong desiconda?

araichoor:
thanks.

sorry, I forgot to mention that running with `godesi 23.10` and the `fiberassign/hexrot` branch also returns the same difference for a potential assignment for tileid=9383.
here: `/global/cfs/cdirs/desi/users/raichoor/fiberassign_dev/hexrot/replay_kpno_20250305/godesi_23.10/branch_hexrot`.
this `godesi 23.10` environment is the one the LSS folks use, if I m correct.

but you re right, it s probably a matter of conda environment here.

sybenzvi:
@araichoor , I regenerated the tiles while being a bit more careful about logging everything. On NERSC I posted three different folders:
1. Tiles generated using the [default version of fiberassign](https://data.desi.lbl.gov/desi/users/sybenzvi/fa_holding_pen_default-20250305/) used daily by `planner` with a rather old version of desiconda.
2. Tiles generated using the [current main branch of fiberassign](https://data.desi.lbl.gov/desi/users/sybenzvi/fa_holding_pen_main-20250305/) with a slightly out of date version of desiconda. I discovered there is an exception raised in one tile (9686) that could be due to version mismatches.
3. Tiles generated using the [hexrot branch of fiberassign](https://data.desi.lbl.gov/desi/users/sybenzvi/fa_holding_pen_hexrot-20250305/). It's using the same slightly old version of desiconda and the same exception occurs on tile 9686.

Please check the "default" version and see if the reproducibility mismatch is present. Then we can decide what to do about the software stack and the `hexrot` branch.

araichoor:
awesome @sybenzvi, thanks.

double-checking: files in `fa_holding_pen_hexrot-20250305/` are the same ones as the ones provided yesterday in `fa_holding_pen_20250305/`? or are those some rerun? (to know if I ve to rerun those at nersc or not).

araichoor:
so, if I didn t mess up, I ve tried to rerun @sybenzvi kpno runs at nersc.

all kpno folders have this prefix: `/global/cfs/cdirs/desi/users/sybenzvi/fa_holding_pen_`.
all nersc folders have this prefix: `/global/cfs/cdirs/desi/users/raichoor/fiberassign_dev/hexrot/replay_kpno_`.

here s what I get, where the last, `ndiff` columns reports the number of disagreements between kpno vs. nersc (we obviously want that number to be zero):

| test nb | kpno folder  | kpno conda | kpno fba | nersc folder | nersc godesi | nersc fba | ndiff
| - | - | - | - | - | - | - | - |
| 1 | default-20250305  | 20200924 | 5.6.0 | default-20250305/godesi_23.10/fba_5.6.0 | 23.10 | 5.6.0 | 0 |
| 2 | default-20250305  | 20200924 | 5.6.0 | default-20250305/godesi_main/fba_main | main | main | 2861 |
| 3 | main-20250305  | test-20230727-2.1.1-debian.dev | main | main-20250305/godesi_main/fba_main | main | main | 2548 |
| 4 | hexrot-20250305  | test-20230727-2.1.1-debian.dev | hexrot | hexrot-20250305/godesi_main/fba_hexrot | main | hexrot | 2 |
| 5 | hexrot-20250305  | test-20230727-2.1.1-debian.dev | hexrot | hexrot-20250305/godesi_main/fba_hexrot_godesi23.10 | main | hexrot_godesi23.10 | 2 |

so, except for the test 1, they all fail...
as @schlafly said, it s somewhat expected, ie what we re testing/seeing here is the difference in the conda environment at kpno vs. nersc; not related to this hexrot branch.
right?

araichoor:
and thanks @ashleyjross for providing the LSS environment set up, that s super-useful.

in that context, one question which was raised at some point, but never answered is: why the fiberassign reproducibility test pass in our dailyops operations?
it may be because the fiberassign code used there (`desiproc` user, environment loaded with `godailyops`) was built on Apr. 25, 2024, which is before whatever change was done which broke the reproducibility at nersc (the first `godesi` loading which breaks is `24.4`, where files where built on Apr. 30, 2024).
that s just a guess...

sybenzvi:
> awesome @sybenzvi, thanks.
> 
> double-checking: files in `fa_holding_pen_hexrot-20250305/` are the same ones as the ones provided yesterday in `fa_holding_pen_20250305/`? or are those some rerun? (to know if I ve to rerun those at nersc or not).

Thanks for these check. Just confirming (a little late) I did re-run those "hexrot" tiles today (2023-03-06) but with the identical environment I used yesterday (2025-03-05) at KPNO.

araichoor:
for info, about: "_I discovered there is an exception raised in one tile (9686) that could be due to version mismatches._"

I m not sure what s happening here.
from the log, the error is when reading the ledgers:
```
  File "/software/datasystems/desiconda/test-20230727-2.1.1-debian.dev/code/fiberassign/main/py/fiberassign/fba_launch_io.py", line 1370, in create_mtl
    d = read_targets_in_tiles(
  File "/software/datasystems/desiconda/test-20230727-2.1.1-debian.dev/code/desitarget/main/py/desitarget/io.py", line 3773, in read_targets_in_tiles
    targets = read_targets_in_hp(
  File "/software/datasystems/desiconda/test-20230727-2.1.1-debian.dev/code/desitarget/main/py/desitarget/io.py", line 3330, in read_targets_in_hp
    return read_mtl_in_hp(
  File "/software/datasystems/desiconda/test-20230727-2.1.1-debian.dev/code/desitarget/main/py/desitarget/io.py", line 3239, in read_mtl_in_hp
    mtl = np.concatenate(mtls)
  File "<__array_function__ internals>", line 180, in concatenate
  File "/software/datasystems/desiconda/test-20230727-2.1.1-debian.dev/conda/lib/python3.10/site-packages/numpy/core/_internal.py", line 458, in _promote_fields
    raise TypeError("invalid type promotion")
TypeError: invalid type promotion
```

for what is worth, at nersc, this runs without error:
```
from desitarget.io import read_targets_in_tiles
t = Table.read("/global/cfs/cdirs/desi/survey/ops/surveyops/trunk/ops/tiles-main.ecsv")
t = t[t["TILEID"] == 9686]
hpdir = "/global/cfs/cdirs/desi/survey/ops/surveyops/trunk/mtl/main/dark/"
d = read_targets_in_tiles(hpdir, tiles=t, mtl=True)
```

also, @sybenzvi:
- what desitarget code version is used there? because it looks like the line numbers do not correspond to the current desitarget/main code;
- for completeness, looks like the same error happens for tileids 23030 and 25252.

sybenzvi:
@araichoor, the version of desitarget that I'm using on the cluster is too old. I'm making an update and trying again.

araichoor:
following my last message, @sybenzvi reran with actually using the tagged versions of the code.
I ve redone the rerun at nersc, and it fails in the same way.

| test nb | kpno folder  | kpno conda | kpno fba | nersc folder | nersc godesi | nersc fba | ndiff
| - | - | - | - | - | - | - | - |
| 7 | 5.7.2-20250314  | kpno-20250312-2.2.1.dev | 5.7.2 | 202503014/godesi_23.10/fba_5.7.2 | 23.10 | 5.7.2 | 2861 |

araichoor:
as we re getting closer to merging this branch, a reminder that we need to update the cutoff date, ie for what `rundate > cutoff_date` this behavior is turned on.
the currently implemented one is `"2025-03-01T19:00:00+00:00"`, so definitely needs to be changed.

araichoor:
cutoff date changed to May, 2, as discussed on slack.
merging now.

PR #467, opened 2025-01-29
field rotation fixes with updated tests

sbailey:
This PR is equivalent to #466 (Field rotation updates to avoid hexapod rotation limit errors), but after re-branching hexrot -> hexrot-updatedtests and rebasing off of main after #464 which fixed the testing infrastructure.  Let's re-evaluate which PR to merge after seeing if these tests pass, and/or if there are other updates to the hexrot branch from the #466 discussion.

Heads up @araichoor @weaverba137 @schlafly @geordie666 @ashleyjross

### DIFF
--- a/py/fiberassign/data/cutoff-dates.yaml
+++ b/py/fiberassign/data/cutoff-dates.yaml
@@ -1,3 +1,4 @@
 rundate:
     std_wd: "2021-12-01T19:00:00+00:00" # std_wd not counted in per-petal-std after that rundate
     etc_stuck: "2022-04-28T19:00:00+00:00" # etc fibers are treated as stuck fibers after this date
+    obsdate: "2025-03-01T19:00:00+00:00" # obsdate set as rundate + 1yr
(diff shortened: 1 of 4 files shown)


araichoor:
@sbailey : you ve gone ahead of me!

looks like the tests pass here -- great, thanks @weaverba137 @sbailey!

could we do the same on the hexrot branch, so that everything is in https://github.com/desihub/fiberassign/pull/466? and delete this hexrot-updatedtests branch?
or am I missing something here?

because there probably will be more development/discussion on https://github.com/desihub/fiberassign/pull/466, once perlmutter is back up.

sbailey:
@araichoor I didn't directly rebase your hexrot branch, because it can become a huge mess if one person rebases someone else's branch out from under them.  If you are the only active user/developer of the hexrot branch, I think it is safe for you to rebase it and then force push it back to github.  Let's chat in person before doing any additional commits.

sbailey:
@araichoor and I sorted out the rebasing in #466, so this PR and branch are no longer needed.  I will close it without merging and delete the branch.

PR #468, opened 2025-02-19
handle dark1b/bright1b

araichoor:
This PR goal is to handle the `DARK1B / BRIGHT1B` programs for the DESI-Extension.
It relies on https://github.com/desihub/desitarget/pull/833 (thanks @geordie666!)

As of now, it is WIP, but I start the PR to initiate the discussion (e.g. on choices to be done).
Note that there are few hacks in the code (of course to be removed later):
- hard-coding the path for the `dark1b` test ledgers+static catalogs: `/pscratch/sd/a/adamyers/blub/dr9/2.8.0.dev5597/`;
- commenting out some QA line, where the code currently crashes because of duplicates between the primary/dark1b and secondary/dark.

Some choices I did:
- `get_desitarget_paths()`: for main survey, now always consider 2 sets (e.g. dark+dark1b or bright+bright1b), for both the ledgers and the static catalogs;
- `get_program_latest_timestamp()`: if dark or dark1b, parse both dark AND dark1b tiles (same for bright, bright1b);
- `MTL_{HIGHEST,WANTED,CONTAINS}` columns:
  - I ve added those in the `FIBERASSIGN` and `TARGETS` extensions of the `fiberassign-TILEID.fits.gz` file;
  - so far, I ve not added those to the `fba-TILEID.fits` file; @ashleyjross please let me know if you d like me to
  - also, I think I ve set those columns as "required to be there"; I ll double check with later tests, but I ve not thought hard if that is an issue for backwards-compatibility or not, tagging @ashleyjross again.

I ve run two test tiles, one dark, one dark1b, with this first commit with these commands:
```
fba_launch --outdir /global/cfs/cdirs/desi/users/raichoor/fiberassign_dev/fiberassign_desi1b/20250218 --tileid 999990 --tilera 176 --tiledec 2 --survey main --dtver 1.1.1  --program DARK --nosteps qa

fba_launch --outdir /global/cfs/cdirs/desi/users/raichoor/fiberassign_dev/fiberassign_desi1b/20250218 --tileid 999991 --tilera 176 --tiledec 2 --survey main --dtver 1.1.1  --program DARK1B --nosteps qa
```
In case people would want to poke around with those, and let me know if there s something they re not happy with, I d appreciate.
Because of the duplicates-with-secondary issue, I ve disabled the `qa` here...

New header keywords in the header: `MTL2` and `TARG2` (those will always be for `{BRIGHT,DARK}1B`, irrespective if the requested program is `{BRIGHT,DARK}` or `{BRIGHT,DARK}1B`):
```
fitsheader -e 0 999/fiberassign-999990.fits.gz
[...]
GFA     = 'DESIROOT/target/catalogs/dr9/1.1.1/gfas'                             
MTL     = 'DESIROOT/survey/ops/surveyops/trunk/mtl/main/dark'                   
MTL2    = '/pscratch/sd/a/adamyers/blub/dr9/2.8.0.dev5597/mtl/main/dark1b'      
SCND    = 'DESIROOT/target/catalogs/dr9/1.1.1/targets/main/secondary/dark/targ&'CONTINUE  'ets-dark-secondary.fits'                                             
SCND2   = 'DESIROOT/target/catalogs/dr9/1.1.1/targets/main2/secondary/dark/mai&'CONTINUE  'n2targets-dark-secondary.fits'                                       
SCND3   = 'DESIROOT/target/catalogs/dr9/1.1.1/targets/main3/secondary/dark/mai&'CONTINUE  'n3targets-dark-secondary.fits'                                       
SCNDMTL = 'DESIROOT/survey/ops/surveyops/trunk/mtl/main/secondary/dark'         
SKY     = 'DESIROOT/target/catalogs/dr9/1.1.1/skies'                            
SKYSUPP = 'DESIROOT/target/catalogs/gaiadr2/1.1.1/skies-supp'                   
TARG    = 'DESIROOT/target/catalogs/dr9/1.1.1/targets/main/resolve/dark'        
TARG2   = '/pscratch/sd/a/adamyers/blub/dr9/2.8.0.dev5597/targets/main/resolve&'CONTINUE  '/dark1b'                                                             
TOO     = 'DESIROOT/survey/ops/surveyops/trunk/mtl/main/ToO/ToO-fiber.ecsv'     
[...]
```


araichoor:
@geordie666, one question:

with the on-going discussion in the twin desitarget PR, I realize I may have mis-interpreted one thing.
let s take dark, dark1b for example here.

is the following the approach we want:
- for a dark tile: only query the dark ledgers
- for a dark1b tile: query both the dark and dark1b ledgers.
right?

in my current coding, I query both (dark, dark1b) ledgers for a dark tile, so that s likely a mistake -- I ll correct that tomorrow when nersc is back up.

geordie666:
Yes, that's correct. It's a little hack-y but, now:

DARK means only DARK.
DARK1B means DARK1B and DARK (suitably merged).

araichoor:
perfect, thanks for the confirmation -- and sorry I forgot about that, you surely already explained it to me!

schlafly:
I'm sure I've missed it, but just confirming---
1. when program = dark, we want to continue getting the same results we have always gotten.  The dark1b ledgers are ignored.  It's okay to have some header new header keywords if it makes the code cleaner and it's clear when they should be ignored, but they don't seem desired.
2. when program = dark1b, we do want to read in both sets of ledgers and merge them, and we do need new header keywords describing the new different new sets of MTLs read in. 

My naive reading of this code makes it look like we always read dark/dark1b and bright/bright1b ledgers and merge them, for both the dark and dark1b programs, which wasn't what I thought we wanted.

araichoor:
thanks for the confirmation (and looking at the code!); I confirm the current code is not doing that, ie is wrong for dark tiles.

I ll work on that today, now that nersc is back up, it s an easy fix.

I ll also code the approach discussed here https://github.com/desihub/desitarget/pull/833#issuecomment-2669588857 for secondaries.

araichoor:
with the last couple of commits, the code should now correctly handle the querying of the catalogs, as discussed.
that s both for primary and secondary targets.
as I ve "enabled" secondary targets for `bright1b,dark1b`, but there are none so far, I had to add a check in `fba_launch`, looking for secondaries only if the ledger folder exists.

araichoor:
@schlafly, I think it s ok but double-checking in case:
for `GOALTYPE`, I add two new values, `DARK1B` and `BRIGHT1B`, that s what we want, right?

araichoor:
for info, I ve propagated the dark1b/bright1b cases in few functions used for the `fiberassign-TILEID.png` plot, and also in `fba_rerun_io.py`.

as previously, I ve generated two test tiles here:
```
fba_launch --outdir /global/cfs/cdirs/desi/users/raichoor/fiberassign_dev/fiberassign_desi1b/20250221 --tileid 999990 --tilera 180 --tiledec 10 --survey main --dtver 1.1.1  --program DARK

fba_launch --outdir /global/cfs/cdirs/desi/users/raichoor/fiberassign_dev/fiberassign_desi1b/20250221 --tileid 999991 --tilera 180 --tiledec 10 --survey main --dtver 1.1.1  --program DARK1B
```

as far as I can say, the fiberassign files look ok to me, ie formatted as we want; of course we d to run more massive tests once the targets are available over the full footprint!

araichoor:
I m just a bit suprised by this https://data.desi.lbl.gov/desi/users/raichoor/fiberassign_dev/fiberassign_desi1b/20250221/999/fiberassign-999991.png plot, where one can see that the assigned `LGE` do not have the same z-band mag distribution as the parent `LGE` sample; the assigned ones are brighter (ie the red curve is not \~flat); I m investigating...

fiberassign-999991

araichoor:
mmhh... looks like that the assigned `LGE` are the \~10% ones which also are `BGS_ANY`.

I m investigating, but I suspect it s because I didn t propagate the `DARK1B` in some c++ part of the code.

because those two samples (`LGE` and `LGE & BGS_ANY`) have the same `PRIORITY`, `NUMOBS_MORE` values, the difference being that `LGE` mostly have `OBSCONDITIONS = ['DARK1B']`, and that `LGE & BGS_ANY` have `OBSCONDITIONS=['BRIGHT', 'IGNORE', 'DARK1B']`.

araichoor:
ok found the issue: I had to enable the `LGE` target class in `targets.py`, I totally forgot about that one!
https://github.com/desihub/fiberassign/blob/526e48f9384dc8acba0d7c1adab7e2f966e47e83/py/fiberassign/targets.py#L221-L238
without that, the `LGE-only` targets would be discarded (hence why mostly the `LGE & BGS_ANY` remained).

the results look better now; +1 for qa plots!

fiberassign-999991


I ve moved previous run to https://data.desi.lbl.gov/desi/users/raichoor/fiberassign_dev/fiberassign_desi1b/20250221/v0.
and runs with this last commit are here: https://data.desi.lbl.gov/desi/users/raichoor/fiberassign_dev/fiberassign_desi1b/20250221/v1.

araichoor:
mmhh.. tests are still failing after my small fix in the previous commit.

@geordie666, I m not familiar with those, but could it be that it fails because there are, in the test scripts, lines like:
https://github.com/desihub/fiberassign/blob/77c7d81aa944b04f999876132eb037f37fb923f8/py/fiberassign/test/test_targets.py#L11
which uses `desitarget`, and the needed desitarget version is the one in development? (https://github.com/desihub/desitarget/pull/833)

in which case, there s nothing to do until that https://github.com/desihub/desitarget/pull/833 is merged + a desitarget/tag is created?
then update that line with this new tag:
https://github.com/desihub/fiberassign/blob/77c7d81aa944b04f999876132eb037f37fb923f8/.github/workflows/test.yml#L29
just guessing...

araichoor:
except that "failing checks" issue, I think I m happy with the current version.
as said, I ll perform further testing when new ledgers/targets will be available over a larger area.

schlafly:
I'm embarrassed, but I don't remember what GOAL TYPE controls.  Maybe which variety of effective time is used?  Presumably we want to continue using the the normal LRG effective time (i.e., dark).  That's less obvious for bright1b, but given that we likely don't want to develop a new effective time, probably we continue to use bright?  Or am I missing what GOAL TYPE controls?

araichoor:
I admit, it s not super-clear for me either where `GOALTYPE` is used; I have same sense as you, ie it s probably used to pick which efftime to choose; I would need to dig into history/codes.

I thought it was used online by the ETC to choose the correct way to estimate the efftime, but I don t see any occurence of `GOALTYPE` in the code, apart from a check that the keyword is there: https://github.com/search?q=repo%3Adesihub%2Fdesietc%20goaltype&type=code.

for the spectro. pipeline, from a quick check, it s indeed used to pick which `EFFTIME_SPEC` to use, e.g.:
https://github.com/desihub/desispec/blob/c717572e415314babdf30c09fa2fe4a75af1da4f/bin/desi_tsnr_afterburner#L509-L526

so my 2 cents would be:
- maybe it s better to define `GOALTYPE`=`DARK1B` or `BRIGHT1B` for those two programs, so that we keep the flexibility to define other properties;
- codes like desispec and desietc will probably need to update anyway for the desi-extension, to handle those new programs `DARK1B` and `BRIGHT1B`; then the decision e.g. of what `EFFTIME` to pick for `DARK1B` and `BRIGHT1B` will be done in those codes -- and if the decision is to keep the same definitions as for `DARK` and `BRIGHT`, no problem, but at least it will be explicitly stated that e.g. a "`DARK1B` should be treated as `DARK`" decision is taken...

schlafly:
I hear you, but: conceptually, the dark (extended) LRG and bright BGS targets are still the defining targets of these extension programs.  So I don't think we'd consider using a different GOALTYPE.

araichoor:
allright, thanks for the input; I ve removed the `1B` from the `GOALTYPE`.

araichoor:
that change (setting `desitarget/2.9.0` for the tests) apparently fixed the tests, great.

araichoor:
ok, so for info, I think I m happy with this fiberassign/desi1b pr:

I ve generated (I ignored here the `IN_DESI` status, and added +100000 to the tileids):
- `DARK1B`: with `3.0.0` all the NGC tiles and 10% of the SGC tiles
- `BRIGHT` and `DARK`: with `1.1.1` 10% of the SGC tiles
- `BACKUP`: with `2.2.0` 10% of the SGC tiles

and I ve rerun one `DARK `and one `DARK1B` tiles, to be sure the code is happy running.

files here: `/pscratch/sd/r/raichoor/fiberassign-dark1b/20250428`.

so I ll merge soon.

araichoor:
merging now.

Issue #474, opened 2025-04-15
use freeze_iers() in the scripts


araichoor:
The tile design at kpno was slow as of Oct. 2024.

@schlafly investigated and found that the culprit was the loading of the IERS by `fba_launch`:
"_fba_launch calls astropy which wants to update the IERS (the leap second thing we've been seeing) and that hangs for 30 s while not reaching the outside network (firewalls)._"
(https://desisurvey.slack.com/archives/C01HNN87Y7J/p1729541934260899)

As a temporary fix, @schlafly then "_copied the planner astropy cache to the msdos astropy cache in order to "fix" the IERS issue_" .(https://desisurvey.slack.com/archives/C01HNN87Y7J/p1729606053866689).

A more robust approach is to bypass that loading using the desiutil freeze_iers function.
I m not super-familiar with that, so I d welcome suggestions for how to implement the fix.

What are the impacted `astropy` imports? `astropy.time` and `astropy.coordinates` only? others?

So far, what I could think of (as it s been said it doesn t hurt to have multiple, redundant calls to that):

First, define in `fiberassign.utils`:
```
import desiutil.iers

freeze_iers = desiutil.iers.freeze_iers
```

Then, add at the top of all executables and all fiberassign/*py and fiberassign/scripts/*py scripts which have the impacted `astropy` imports:
```
from fiberassign.utils import freeze_iers
freeze_iers()
```

Would that make sense?
Or is there a more meaningful way to do that?

Also, I was wondering: what about desimeter? Should that be added in there, as fiberassign calls some desimeter functions? Or should the `freeze_iers()` call in fiberassign "propagates" to desimeter?

weaverba137:
This appears to have been fixed by #475. Does it need to remain open?

PR #475, opened 2025-04-16
Freeze iers

araichoor:
and another PR...

this addresses https://github.com/desihub/fiberassign/issues/474, with calling `freeze_iers()`.

I ve chosen the approach I suggested in that ticket, i.e. add `freeze_iers()` in:
- all fiberassign `*py` scripts using `astropy.time `and `astropy.coordinates`;
- also in the executables `fba_launch` and `fba_rerun`.

as a sanity check, I ve rerun 58 dark ngc tiles (using `godesi 23.10` to compile the fiberassign c++ code; but the rerun is done with `godesi main` and this branch).

I m happy to hear feedback/suggestions if there is a better way to proceed, as I m not very familiar with that.

once we ve convergd, I guess we d want to also test at kpno, with the same user as done for operations -- and maybe also beforehand undoing the current temporary fix? (ie the "_copied the planner astropy cache to the msdos astropy cache in order to "fix" the IERS issue_").

### DIFF
--- a/bin/fba_launch
+++ b/bin/fba_launch
@@ -32,6 +32,7 @@ from fiberassign.fba_launch_io import (
     copy_to_svn,
 )
 from desiutil.iers import freeze_iers
+from fiberassign.utils import get_fba_use_fabs
 from argparse import ArgumentParser
 
 # AR allowed steps in fba_launch
@@ -91,6 +92,23 @@ def main():
         log.info("{:.1f}s\trundate\tTIMESTAMP={}".format(time() - start, Time.now().isot))
         args.rundate = get_latest_rundate(log=log, step="rundate", start=start)
         log.info("{:.1f}s\trundate\tsetting args.rundate={}".format(time() - start, args.rundate))
+
+    # AR fba_use_fabs: if None, set according to what is in the data/cutoff-dates.yaml
+    if args.fba_use_fabs is None:
+        log.info("")
+        log.info("")
+        log.info("{:.1f}s\tfba_use_fabs\tTIMESTAMP={}".format(time() - start, Time.now().isot))
+        args.fba_use_fabs = get_fba_use_fabs(args.rundate)
+        log.info("{:.1f}s\tfba_use_fabs\tsetting args.fba_use_fabs={}".format(time() - start, args.fba_use_fabs))
+        if os.getenv("FIBERASSIGN_USE_FABS") is not None:
+            if os.getenv("FIBERASSIGN_USE_FABS") != str(args.fba_use_fabs):
+                msg = "already defined environment variable FIBERASSIGN_USE_FABS={} will be overwritten by args.fba_use_fabs={}".format(
+                    os.getenv("FIBERASSIGN_USE_FABS"),
+                    args.fba_use_fabs
+                )
+                log.warning("{:.1f}s\tfba_use_fabs\t{}".format(time() - start, msg))
+    os.environ["FIBERASSIGN_USE_FABS"] = str(args.fba_use_fabs)
+    log.info("{:.1f}s\tfba_use_fabs\tsetting $FIBERASSIGN_USE_FABS={}".format(time() - start, os.getenv("FIBERASSIGN_USE_FABS")))
 
     # AR mtltime
     # AR if possible: setting to the latest of:
@@ -743,6 +761,12 @@ if __name__ == "__main__":
         action="store_true",
         default=False,
     )
+    parser.add_argument(
+        "--fba_use_fabs",
+        help="value to determine the cpp behavior, see PR470 (default: based on the rundate)",
+        type=str,
+        default=None
+    )
 
     args = parser.parse_args()
     log = Logger.get()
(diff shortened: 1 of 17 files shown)


bin/fba_launch
```diff
@@ -31,6 +31,7 @@ from fiberassign.fba_launch_io import (
     mv_temp2final,
     copy_to_svn,
 )
+from fiberassign.utils import freeze_iers
```
weaverba137, line 34: This should be `from desiutil.iers import freeze_iers`. If there is a specific reason for this import indirection, please describe it in detail. My personal feeling is that this indirection will only cause serious confusion.

weaverba137 (dismissed):
I also have concerns about `freeze_iers()` being called at the module level. That is untested. The standard call is, *e.g.*:

```python
def main():
    """
    """
    freeze_iers()
```
*i.e.* the very first call when entering a script.

araichoor:
thanks @weaverba137 for the feedback, I appreciate!

- for the call in executable scripts, like `fba_launch` or `fba_rerun`: sure, I can move the `freeze_iers` into the `main()` function
- for the other scripts (ie calls at the module level): I admit, I don t really know how to proceed; the reason is that for instance, the lss group (@ashleyjross ), when running `altmtl`, just runs some functions of fiberassign, they don t use `fba_launch`; I d just want to be sure that this `freeze_iers()` is consistently called; that s why I ve added it to the top of the `py/fiberassign/*py` scripts; having that call in each relevant function may be fastidious... so I don t know what the alternative is; maybe just remove those, and let the lss folks have that call in their executable scripts?
- about the "indirect" call in the fiberassign scripts: I admit, I copied  -- or tried to copy -- what @schlafly did in desisurvey; but I m happy to do the more straightforward way.

araichoor:
@weaverba137: I pushed few commits which should address your comments.
as of now, the `freeze_iers()` is only called in the `main()` of the two executable scripts `fba_launch` and `fba_rerun`; and it s directly calling the `desiutil.iers` function.

weaverba137:
Looks good. It's now a minimal change, so I think this can be merged when ready.

sybenzvi:
@anand, I can try to run fiberassign at KPNO, but I'll need to dig a bit to ensure the caching is correct. If you want the same NGC tiles you tested at NERSC, tell me which ones you ran on.

Since we're not in a big rush for this particular fix, I would prefer to wait for Eddie to return from vacation before we make this the official version in the planner account. If there is some (undocumented?) entanglement with the msdos account, then his input is needed.

araichoor:
yes, maybe better/simpler to wait for Eddie for this one.
I guess any few tens of tiles you d run would be fine.
thanks.

schlafly:
I think you could test this as usual. 

The interaction with the msdos account is that apparently astropy downloads the iers file in a user-local cache, which becomes the msdos account when run as the msdos user.  I actually don't know how freeze_iers works under the hood; Stephen might be the best reference.

sbailey:
> The interaction with the msdos account is that apparently astropy downloads the iers file in a user-local cache, which becomes the msdos account when run as the msdos user. I actually don't know how freeze_iers works under the hood; Stephen might be the best reference.

@weaverba137 is the best reference here.  My understanding is that freeze_iers instructs astropy to *not* download the latest file into a local cache, but instead tells astropy to use the desiutil/py/desiutil/data/iers_frozen.ecsv file instead, which is included as part of a desiutil installation.  I defer to @weaverba137 for any corrections/clarifications on that.

weaverba137:
I agree with @sbailey's summary.

schlafly:
Great.  We shouldn't need to worry about any issues with the msdos user's environment in this case.  Anand's patch here just calls freeze_iers() early in main before any astropy iers calculations.  Is there more that we need to think about or check here?

araichoor:
> Right. In other packages, the executable or `main()` function _in that package_ is expected to call `freeze_iers()`, rather than the imported package doing it. So I think the LSS group needs to look into this. Furthermore, the `freeze_iers()` function was written with the expectation that it would be called this way. It is not intended to be called at the module level and therefore at import time.
> 
> >

actually, I realize that, as the `altmtl` uses the `fba_run` executable script, I should also add a call to `freeze_iers()` in there! (@ashleyjross)
I ll do that right now.

araichoor:
on my side, I think the PR is good to go, code-wise.

sybenzvi:
I think we can merge this. The only potential issue could be at KPNO if we try to run the new fiberassign with an older desiutil + astropy, but since all our updates are going to be done manually, I don't think we have to worry about getting caught in an inconsistent state by some `git pull` cronjob. There are cronjobs at perlmutter but all of the "modern" versions of astropy + desiutil are present there.

schlafly:
I can test at kpno, but it's probably easiest if we have everything merged in main for that.  I would go ahead and merge and we can test at KP with the "final" (but untagged) product before tagging?

araichoor:
ok great, thanks for the input.
I ll merge now then.

PR #477, opened 2025-04-24
include cstdint for gcc/13 support

sbailey:
This PR adds "`#include <cstdint>`" which is required to compile fiberassign with gcc/13.2.1 at NERSC.  Without it, we get errors like
```
src/pybind11/attr.h: At global scope:
src/pybind11/attr.h:185:10: error: ‘uint16_t’ in namespace ‘std’ does not name a type; did you mean ‘wint_t’?
  185 |     std::uint16_t nargs;
      |          ^~~~~~~~
      |          wint_t
```
Also see [specex #73](https://github.com/desihub/specex/pull/73).  Credit to @dmargala for debugging.

I plan to self-merge and update at NERSC since the current compilation of fiberassign/main is failing.

### DIFF
--- a/src/_pyfiberassign.cpp
+++ b/src/_pyfiberassign.cpp
@@ -1,5 +1,6 @@
 #include <string>
 #include <sstream>
+#include <cstdint>
 
 #include <pybind11/pybind11.h>
 #include <pybind11/operators.h>


araichoor:
yes, sorry, I didn t think about that when we did the fiberassign merging yesterday.

one curiosity question, though:
as I m on dailyops, I ve been running the fiberassign repro. script today (I think I started around noon).
that s done from `desiproc`, and the used script is `/global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/code/fiberassign/main`, if I m correct.
files there have been updated this night, as expected from the cronjob, and the code has been running smoothly.
but it s the same location as what s used for the main, no?
maybe you fixed the compilation before I ran the code?

sbailey:
The nightly build failed, but that apparently leaves behind the previously working copy (which had last been compiled nearly a year ago!).  I just merged this update at NERSC and did a fresh rebuild.  @araichoor please test.

araichoor:
thanks; I ll check.
I actually misread the timestamps: today is 2025-04-24, but the previous compilation dates from 2024-04-25, so looks similar!

araichoor:
I ve rerun `validate_lastnight_fba --night 20250423` from `desiproc`, and that was successful in reproducing the 16 tiles.
perfect!

PR #478, opened 2025-04-24
get_fba_use_fabs(): add log infos

araichoor:
In the spirit of "let s keep track of key settings in the fiberassign log", this PR adds infos about the `get_fba_use_fabs()` which sets the environment variable introduced in https://github.com/desihub/fiberassign/pull/470 to control the c++ behavior.

I noticed I forgot to add that then...

PR #479, opened 2025-04-24
Doc err fix

sybenzvi:
The docs build is emitting an intersphinx "403 forbidden" warning for https://matplotlib.org/stable/objects.inv, for reasons that are unclear; the objects inventory file exists. Since `sphinx-build` is running with the `-W` option that turns warnings into errors, the CI fails. So this just removes the `-W` option from the docs build.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -95,4 +95,5 @@ jobs:
             - name: Compile the package
               run: CC=g++ CXX=g++ CFLAGS="-fPIC" LDSHARED="g++ -shared" python setup.py build_ext --inplace
             - name: Test the documentation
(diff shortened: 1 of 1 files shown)


araichoor:
thanks!
this error wasn t there yesterday, so maybe it s just some instability on the matplotlib site?

also just checking: we should merge this one, then I should update https://github.com/desihub/fiberassign/pull/478 with this change, and it should be fine?

weaverba137:
Who will take the responsibility to restore the `-W` flag?

sybenzvi:
I'll make a ticket @weaverba137 .

weaverba137:
That's fine, thanx.

PR #481, opened 2025-05-01
Handle exceptions in copy_to_svn

schlafly:
This PR wraps the permissions-changing lines in a try/except block to make them more robust.

### DIFF
--- a/bin/fba_launch
+++ b/bin/fba_launch
@@ -791,7 +791,10 @@ if __name__ == "__main__":
         dirmode = 0o2775 if args.worldreadable else 0o2770
         fn = os.path.join(args.outdir, subdir)
         os.makedirs(fn, exist_ok=True, mode=dirmode)
-        os.chmod(fn, mode=dirmode)  # apparently umask can affect os.makedirs?
+        try:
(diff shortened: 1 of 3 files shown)


schlafly:
Usually the file already has the right permissions, but the wrong owner.  So we don't actually need to do anything, and the chmod is just causing a crash for no reason.  A better way to have written the code would probably be to check if the directory needs to be made, and if so, to make it (rather than rely on exists_ok).  Then only update the permissions if we actually made it (and are therefore the owner).  This was just the simplest thing to avoid the exception causing us to crash.

araichoor:
ok I see, thanks.
so is it ok to merge with these code changes? (if so, I ll update the `doc/changes.rst` file, and merge).

schlafly:
Yes.  That said, I've made these changes "blind" and haven't actually made an SVN directory as another user to make sure that it handles this case gracefully; it's just more work than I wanted to get a new version installed at KPNO to test this well.  But I did test this change which does ~the same thing at KPNO
https://github.com/desihub/desisurvey/pull/168/files

araichoor:
ok perfect; I ve just pushed a commit to update `doc/changes.rst`; I ll merge once the tests have passed.

PR #486, opened 2025-08-30
Rerun main tiles and use of dvs_ro in fba_rerun

araichoor:
This PR adds the script - and required material - to rerun all main fiberassign files, i.e. what I have been using to independently verify that we can indeed reproduce all tiles after critical fiberassign PRs.

The only modification to current scripts mechanics is in `bin/fba_rerun`, to use `/dvs_ro/` for `$DESI_TARGET`, `$SKY_BRICKSDIR`, and `$SKYHEALPIXS_DIR`, so I think it s safe to merge.

The "required material" is a small file listing the early 157 main tiles designed with `dtver=1.0.0`, which are known to be not reproducible -- and hence are excluded from the test here; along with that, two new small functions in `utils.py`:
- `get_main_dtver(tileid, svndir=None)`: retrieve the `dtver` (desitarget catalogs tag) for a main tile;
- `get_main_early_nonrepro_tiles()`: retrieve the 157 early main tiles with `dtver="1.0.0"`. 

the `bin/fba_rerun_main_tiles` script works per program, and optionally per cap.

Here s the help message:
```
fba_rerun_main_tiles 

    Rerun fiberassign for all main PROGRAM tiles.
    We exclude the 157 early tiles designed with DTCATVER<1.1.1, as those are not reproducible anyway.

    fba_rerun_main_tiles --rerundir RERUNDIR --program PROGRAM [--cap CAP] [--njob NJOB] [--frac FRAC] --dvsro [DVSRO]

    --rerundir RERUNDIR  output folder
    --program  PROGRAM   BACKUP, BRIGHT, BRIGHT1B, DARK, DARK1B
    --cap      CAP       ALL, NGC or SGC (default=ALL)
    --njob     NJOB      nb of parallel jobs running (default=128)
    --frac     FRAC      fraction of the tiles to rerun (default=1)
    --dvsro    DVSRO     in DESI_TARGET, SKYBRICKS_DIR, SKYHEALPIXS_DIR, replace /global/cfs/cdirs/desi/ by /dvs_ro/cfs/cdirs/desi/ (y or n; default=y)

    The overall environment must be loaded.
    For instance, run:

=====
source /global/cfs/cdirs/desi/software/desi_environment.sh main
export DESIMODEL=/global/common/software/desi/perlmutter/desiconda/current/code/desimodel/main
export SKYHEALPIXS_DIR=$DESI_ROOT/target/skyhealpixs/v1
====
```

I have so far run those manually, from interactive sessions.
The per-cap run makes it fit into the 4h of those sessions.

Recently, I noticed it was super-slow, because of the reading of the desitarget catalogs.
Replacing "/global/" by "/dvs_ro/" solved that.
As a consequence, this is now done by default in the `fba_rerun` script (though one can desactivate that with the `--no_dvsro` option).

With this "/dvs_ro/" change, quite some processes can be run in parallel; I ve set the default to 128.
For instance, for the `BRIGHT-NGC`, there are 4734 tiles so far, and it takes 1h on an interactive node.

So far, I ve run those tests with launching manually in an interactive session commands like:
```
time fba_rerun_main_tiles --rerundir /pscratch/sd/r/raichoor/fiberassign-rerun-main/20250830-main --program BRIGHT --cap NGC
```
The benefit of that approach is, if the session times out during one call, one can just re-log to another session and re-launch that call (which will overwrite the not-finished results, but the "loss" time is at most 1h), at the cost of some monitoring.

I ve tried to wrap that in a sbatch script and launch that through the regular queue, but my initial test was showing that e.g. this `BRIGHT-NGC` case was taking several hours, hence was much, much slower than the interactive node.
so I didn t push further; maybe someone with better knowledge could work on that.
In case, here s the script I ve set up for that test:
```
#!/bin/bash -xv

#SBATCH --nodes 1
#SBATCH --constraint cpu
#SBATCH --qos regular
#SBATCH --account desi
#SBATCH --job-name fba_rerun_20250829
#SBATCH --output /pscratch/sd/r/raichoor/fiberassign-rerun-main/20250829-main/fba_rerun_20250829-%j.log
#SBATCH --time=24:00:00
#SBATCH --exclusive

source /global/cfs/cdirs/desi/software/desi_environment.sh main

export DESIMODEL=/global/common/software/desi/perlmutter/desiconda/current/code/desimodel/main
export SKYHEALPIXS_DIR=$DESI_ROOT/target/skyhealpixs/v1

OUTDIR=/pscratch/sd/r/raichoor/fiberassign-rerun-main/20250829-main

for PROGRAM in BACKUP BRIGHT BRIGHT1B DARK DARK1B; do
    for CAP in NGC SGC; do
        date
        CMD="time fba_rerun_main --rerundir $OUTDIR --program $PROGRAM --cap $CAP"
        echo $CMD
        eval $CMD
    done
done
```


araichoor:
@ashleyjross @LNapolitano: I m wondering if, for the `altmtl` reruns, you ve looked at replacing `/global/` with `/dvs_ro/`, to speed up the reading of the desitarget catalogs?

For my case here, it provided huge speed increase, as I run lots of parallel processes, which will go read the same large files (e.g. the secondary files; those last days, with `/global/`, each fiberassign run was taking >1min to read the secondary files...); I think the speed increase mostly come with modifying `$DESI_TARGET`; but I also modify `$SKY_BRICKSDIR` and `$SKY_HEALPIXSDIR` for free..

Like these lines:
https://github.com/desihub/fiberassign/blob/7a6ad80174d2578c51f7d4a8270b7c307a06af1d/bin/fba_rerun#L30-L41

Something in this spirit could be added in https://github.com/desihub/LSS/blob/a16dfa40a846647cf532149afdc0a7386da93f22/py/LSS/SV3/fatools.py#L214-L272?

I m mentioning in case.

dylanagreen:
This all looks good. I've added a python only script to rerun all main tiles, which may be easier to long term maintain, but the code to determine which tiles are not reproducible is great!

PR #490, opened 2025-09-16
numpy 2, importlib, test matrix coverage

### LINKED ISSUE #484: NumPy 2.0 compatibility
Expand testing to ensure compatibility with NumPy >= 2.0.

sbailey:
This PR expands the test matrix version coverage, addresses various deprecation warnings raised with numpy 2.x, and replaces `pkg_resources` with `importlib.resources`.

Closes #484.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,13 +15,26 @@ jobs:
         strategy:
             fail-fast: true
             matrix:
-                os: [ubuntu-22.04]  # , ubuntu-24.04]
-                python-version: ['3.10']  # '3.11', '3.12']
-                numpy-version: ['<1.23']
-                astropy-version: ['<6.1', '<7']
+              include:
+                - os: ubuntu-latest
+                  python-version: '3.13'
+                  numpy-version: '<2.3' # Numba does not yet support NumPy 2.3
+                  astropy-version: '<8.0'
+                - os: ubuntu-latest
+                  python-version: '3.12'
+                  numpy-version: '<2.2'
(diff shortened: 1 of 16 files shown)


.github/workflows/test.yml
```diff
@@ -15,14 +15,27 @@ jobs:
...
+                  astropy-version: '<6.1'
+                - os: ubuntu-latest
+                  python-version: '3.10'
+                  numpy-version: '<2.0'
+                  astropy-version: '<6.1'
 
         env:
-            DESIUTIL_VERSION: 3.4.3
-            DESIMODEL_VERSION: 0.19.2
+            DESIUTIL_VERSION: 3.5.0
```
weaverba137, line 37: Although it won't make a huge difference for the tests, this is not the latest version of desiutil.

Also, you're continuing to install desiutil from GitHub instead of PyPI. If that's necessary for the test, please add a comment as to why.

py/fiberassign/assign.py
```diff
@@ -74,7 +74,7 @@
     ("LAMBDA_REF", "f4"),
     ("PETAL_LOC", "i2"),
     ("DEVICE_LOC", "i4"),
-    ("DEVICE_TYPE", "a3"),
+    ("DEVICE_TYPE", "S3"),
```
weaverba137, line 77: Just a comment. I don't remember `aN` being a type descriptor in NumPy.  Clearly `S3` is the better choice.

weaverba137 (approved):
One comment about desiutil version in tests, but otherwise this looks fine to me.

PR #491, opened 2025-09-25
Minor Enhancements and Bug Fixes to fba_replay.

dylanagreen:
This PR makes (minor) changes to fba_replay_kpno.sh. The script is designed to load logs and fiberassign output files copied from KPNO, rerun the fiberassignment using the currently initialized environment (and associated desi* package versions, which may be different than the original) to either confirm or deny reproducibility. I have been using the script to rerun fiberassign to check the impact of the MTL read_two_ledgers bug. 

First, this PR fixes a bug that originally hardcoded the subdirectory, meaning any subdirectory passed to the script was ignored, and the script attempted to run only over the subdirectory "100". 

Second, I've added some quality of life toggles as command line arguments to the script:
`--rerun`: toggles running the actual fiber assign rerun
`--compare`: toggles running the comparison between the new fiberassignment and the KPNO fiber assignment.
`--concat`: toggles concatenating all of the individual diff files into one global file.

These toggles were motivated by the large volume of tiles necessary to rerun. In this regime it was necessary to run in two jobs, instead of running all the tiles (or attempting to) in one job. The script as originally coded, for example, only concatenates the diffs for the tile(s) run in this specific job/run, but with the command switch you can now concat the two runs together by passing only the `--concat` flag.

Alternatively, one could run only the `--rerun` flag if they only wanted to replay, but did not necessitate any information about the difference to the KPNO output (for example, if a user has their own pipeline for comparing fiber assign). 

The behavior of these toggles is as such:
- If no toggles are passed, the script assumes that all three steps are requested. This means that the script is backwards compatible (modulo the bug fix) with previous versions: if you run the script with the exact command before and after this PR the results are identical.
- If one or more toggles are passed, then only those specific steps are run.

### DIFF
--- a/etc/fba_replay_kpno.sh
+++ b/etc/fba_replay_kpno.sh
@@ -14,6 +14,9 @@ usage () {
     --subfolder     SUBFOLDER     optional; if set, only rerun that subfolder (e.g. 001, for 1??? tileids)
     --rerundir      RERUNDIR      the local folder where outputs will be stored
     --njob          NJOB          nb of parallel jobs running (default=8)
+    --rerun                       optional; toggles whether to do the rerun or not
+    --compare                     optional; toggles comparing rerun fiberassign output to original KPNO output
+    --concat                      optional; toggles concatenating all the diff files into one large diff file
 
     Should be run on one or multiple interactive (or debug) nodes.
     If the reproducibility is successful, then the \$RERUNDIR/fba_nersc_kpno-diff.asc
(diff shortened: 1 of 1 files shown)


sbailey:
Thanks, this looks good.  I think we could merge and deploy this, but I see you @dylanagreen have the PR in a "draft" state so I won't merge it yet in case you had other commits pending etc.  Please confirm and merge when ready.

@araichoor thanks for your continued support of DESI work even from your new position!

dylanagreen:
@sbailey Thanks for looking at this already, I have it in draft state in case I thought of something else that might be useful to implement during my rerun, but the rerun is complete now and the changes are stable so let's say it's ready to merge.

Issue #493, opened 2025-11-20
fiberassign tests in Github Actions are pinned to uninstallable versions of desimodel and desitarget.


dylanagreen:
While addressing #488, and testing coverage reports in my own fork of fiberassign, I ran into a problem created in recently updated pip versions (I assume). In the pinned versions of desimodel and desitarget used for the fiberassign test suite, the associated setup.py files imported a setup function from desiutil. A recent change in pip changed the way the environment is setup while installing packages, and these packages will fail to install because the desiutil package will not be found. See e.g. this failed workflow run: https://github.com/dylanagreen/fiberassign/actions/runs/19520877880/job/55883868850

I can reproduce this problem locally, for example, with a minimal environment setup with
```
python -m venv /path/to/venv/
source /path/to/venv/bin/activate
python -m pip install desituil
python -m pip install --no-cache-dir git+https://github.com/desihub/desimodel.git@0.19.3#egg=desimodel
```

which will fail because desiutil is not found. You can confirm, however, that it exists in the environment with `python -c "import desiutil"` which will run without failing. 

Installing the main/bleeding edge versions of desimodel and desitarget succeeds, because @weaverba137 removed this troublesome import from desitarget in https://github.com/desihub/desitarget/commit/100ebf8776f7f4c03a6d4aeafd4f5b9633a2d9b8, and from desimodel in https://github.com/desihub/desimodel/commit/330f7a30bf2a056aadcdcfff67415d2208b1d669.

But the problem still remains for fiberassign, namely, that the next time the github action runs the test suite it will fail (as evidenced by the fact that it did so on my fork). 

Is it strictly necessary to pin these packages to these specific versions? The fiberassign tests all pass on desi main (i.e. source desi main and running pytest on fiberassign passes all tests), so from an outside perspective the pinning seems like it could be updated to be at least to a tagged version of the package *after* these changes were made, but I don't know if these pinned versions were picked for historical reasons. 

If we are ok with updating the pins, then that will necessitate tags in desimodel and desitarget that exist after the above commits, which I think currently neither package has. In my opinion both packages **should** have been tagged after the pyproject.toml was added, as even if it was a small code change it is a significant enough change to *how the package is installed* that it should he been bookmarked. If we can make those new tags and update the pins then when I add test coverage (when I get it working) I will include updated pinned package versions. 

Pinging @sbailey for his input as well.

Issue #488, opened 2025-09-03
Add coverage test reports


weaverba137:
Although this package has a unit test suite and GitHub Actions setup, unit test coverage is not measured. This is trickier with C++ packages, of course.

PR #494, opened 2025-11-20
Add coverage analysis to github action

dylanagreen:
This commit fulfills #488 and adds coverage testing to the github action. This is done as follows:

- Uses pytest-cov to generate the coverage for the python code, in the usual .coverage file.
- C++ coverage is generated using mostly build in tools. The C++ code is built with an additional flag `--coverage` which builds .gcno files alongside the compiled files. Then when the C++ code is run, the built binary generate .gcda files which record which lines were run during that binary execution. These .gcda and .gcno files are collated into a single coverage report using `lcov` (which we have to install into the Ubuntu VM).
- Instead of using coverage.py or any other external package this PR uses the official Coveralls Github Action to upload the results to Coveralls. This allows us to upload the python `.coverage` file and the C++ `coverage_cpp.info` files to Coveralls, which then handles collating that data on their side.

In addition, this PR had to add the `--no-build-isolation` flag to the `desimodel` and `desitarget` installs, in order to get around #493. 

A coverage badge has also been added to the README.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -12,6 +12,10 @@ fiberassign
     :target: https://fiberassign.readthedocs.io/en/latest/
     :alt: Documentation Status
 
+.. image:: https://coveralls.io/repos/github/desihub/fiberassign/badge.svg
+:target: https://coveralls.io/github/desihub/fiberassign
+
+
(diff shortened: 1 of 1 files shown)


sbailey:
The coverage badge is now showing a number, so perhaps there was just some processing delay, but the "target:" line remains surprising to me. Cosmetic, but if it can be removed that would be nice.

Issue #501, opened 2026-03-27
fba_plot is broken for tiles from before the 1B program.


dylanagreen:
Encountered while trying to plot a calibration field tertiary tile. Nominally there's no reason why the tile should fail through fba_plot, but it does because it can't find the field MTL_HIGHEST, which was added when we added the 1B program. But the calibration tile targets obviously never had that. Any tile before the 1B program was added also don't have an MTL_HIGHEST column.

```fba_plot /global/cfs/cdirs/desi/target/fiberassign/tiles/trunk/083/fiberassign-083043.fits.gz  --out_dir .
INFO: Plotting 1 fiberassign tile files
INFO: Creating ./fiberassign-083043.pdf
INFO: rundate = None is after rundate_cutoff_std_wd = 2021-12-01T19:00:00+00:00, so STD_WD are *not* counted as TARGET_TYPE_STANDARD
multiprocessing.pool.RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/conda/lib/python3.10/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
  File "/global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/conda/lib/python3.10/multiprocessing/pool.py", line 48, in mapstar
    return list(map(*args))
  File "/global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/code/fiberassign/main/py/fiberassign/vis.py", line 354, in plot_assignment_tile_file
    read_assignment_fits_tile(infile)
  File "/global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/code/fiberassign/main/py/fiberassign/assign.py", line 875, in read_assignment_fits_tile
    targets_data[col][:nrawtarget] = fbtargets[col][:]
ValueError: no field of name MTL_HIGHEST
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/code/fiberassign/main/bin/fba_plot", line 16, in <module>
    main()
  File "/global/common/software/desi/perlmutter/desiconda/20240425-2.2.0/code/fiberassign/main/bin/fba_plot", line 11, in main
    run_plot(args)
[...]

PR #502, opened 2026-03-30
Only try load MTL_* columns if desitarget >= 2.9.0

dylanagreen:
This fixes the bug identified in #501 by checking the fiberassign headers for the desitarget version. 

This PR fixes the bug by checking the fiberassign header for the `desitarget` version, and then comparing whether its before or after 2.9.0. 2.9.0 is when the MTL* columns were added to desitarget, so those columns do not exist in the fiberassign file before that version. They do, however, exist afterwards.

Behaviour for fba_plot should remain the same for all tiles after desitarget 2.9.0. For tiles before, this PR restores the ability to plot those tiles by no longer attempting to load the MTL* columns. 

I tested this both on the tertiary tile and early BRIGHT tile identified in the ticket:

`fba_plot /global/cfs/cdirs/desi/target/fiberassign/tiles/trunk/023/fiberassign-023093.fits.gz  --out_dir .`
`fba_plot /global/cfs/cdirs/desi/target/fiberassign/tiles/trunk/083/fiberassign-083043.fits.gz --out_dir .`

As well as a new DARK tile:

`fba_plot /global/cfs/cdirs/desi/target/fiberassign/tiles/trunk/004/fiberassign-004170.fits.gz  --out_dir .`

all of which work and plot as expected.

### DIFF
--- a/py/fiberassign/assign.py
+++ b/py/fiberassign/assign.py
@@ -812,7 +812,7 @@ def read_assignment_fits_tile(tile_file):
         # MTL_HIGHEST, MTL_CONTAINS, and MTL_WANTED. So we check the
         # desitarget version, and if it's below 2.9.0, when those columns
         # were added, we don't try and load those columns.
-        keys_to_check = [s for s in header.keys() if "DEPNAM" in s]
+        keys_to_check = [s for s in header if "DEPNAM" in s]
         version = "0.0.0"
         for k in keys_to_check:
             if header[k] == "desitarget":
--- a/py/fiberassign/assign.py
+++ b/py/fiberassign/assign.py
@@ -808,9 +808,25 @@ def read_assignment_fits_tile(tile_file):
                 if col in fbsky.dtype.names:
                     fiber_data[col][npos:] = fbsky[col]
 
-        full_targets_columns = [(x, y) for x, y in
-                                merged_targets_columns.items()]
-        full_names = [x for x, y in merged_targets_columns.items()]
+        # DG Tiles assigned before the 1b program don't have the columns
+        # MTL_HIGHEST, MTL_CONTAINS, and MTL_WANTED. So we check the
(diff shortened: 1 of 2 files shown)


dylanagreen:
I think this change should be generally safe, the affected function is only called from visualization and QA code, *not* in any actual fiberassignment, and all tests still pass even with this change.

py/fiberassign/assign.py
```diff
@@ -808,9 +808,25 @@ def read_assignment_fits_tile(tile_file):
...
                     fiber_data[col][npos:] = fbsky[col]
 
-        full_targets_columns = [(x, y) for x, y in
-                                merged_targets_columns.items()]
-        full_names = [x for x, y in merged_targets_columns.items()]
+        # DG Tiles assigned before the 1b program don't have the columns
+        # MTL_HIGHEST, MTL_CONTAINS, and MTL_WANTED. So we check the
+        # desitarget version, and if it's below 2.9.0, when those columns
+        # were added, we don't try and load those columns.
+        keys_to_check = [s for s in header.keys() if "DEPNAM" in s]
```
geordie666: It's a small thing, but do you really need the `.keys()` here? I'd think that `[s for s in header if "DEPNAM" in s]` should work because the keys are the iterable for a dictionary.

geordie666 (approved):
This looks good to me. I've independently confirmed that `2.9.0` is the relevant `desitarget` version for this purpose.

Please remember to update the `doc/changes.rst` file before merging.

I added one minor didactic comment, but feel free to ignore it.

py/fiberassign/assign.py
dylanagreen: Good catch, I went ahead and removed it.

I've only ever iterated over a fits header using `.keys()` since I didn't realize it had inherited the functinoality to do so without it from the base dictionary object (since the header is a `fitsio.header.FITSHDR` object). You learn something new every day!

Issue #503, opened 2026-03-31
Update fba_launch_io.py to handle secondary targets in the 1B era


geordie666:
@dylanagreen: On yesterday's survey ops telecon Stephen and I hashed out a plan to handle secondary targets on `1B` tiles. The basic idea is to add targets from `DARK1B` and `BRIGHT1B` secondary ledgers to `DARK` and `BRIGHT` tiles, in addition to adding them to `DARK1B` and `BRIGHT1B` tiles.

This chosen approach will require some updates to `fba_launch_io.py`, I think, which I'm going to try to hash out here:

- [x] The `1B` primary ledgers will need to read the `1B` secondary ledgers. If I understand how [Anand set the code up](https://github.com/desihub/fiberassign/blob/b66d17fd96f5661c28af26e4d3606c8c73d551ad/py/fiberassign/fba_launch_io.py#L770-L775) this should happen automatically, although we should test that is indeed the case. I will need to add [the `main4` static target files](https://github.com/desihub/fiberassign/blob/b66d17fd96f5661c28af26e4d3606c8c73d551ad/py/fiberassign/fba_launch_io.py#L999-L1002) for the new secondary targets, although I _think_ `fiberassign` will detect those automatically.

- [ ] As a side development, we'll now need to also be able to read secondary targets on `BACKUP` tiles. I think this will require updates at least [to read the secondary `BACKUP` ledgers](https://github.com/desihub/fiberassign/blob/b66d17fd96f5661c28af26e4d3606c8c73d551ad/py/fiberassign/fba_launch_io.py#L810) and to find the [static target files](https://github.com/desihub/fiberassign/blob/b66d17fd96f5661c28af26e4d3606c8c73d551ad/py/fiberassign/fba_launch_io.py#L983).

- [ ] We'll probably need a bunch of changes to have `DARK` and `BRIGHT` read the `DARK1B` and `BRIGHT1B` ledgers. This will look something like Anand's [reverse version of those updates](https://github.com/desihub/fiberassign/blob/b66d17fd96f5661c28af26e4d3606c8c73d551ad/py/fiberassign/fba_launch_io.py#L803-L808). This development can be a little slower as it won't be needed until we have new `DARK` and `BRIGHT` tiles in place from the DR11 imaging.

We can likely turn these features on in stages, with the first task being ASAP, but the next two tasks being less time-critical.

[...]

PR #504, opened 2026-04-06
Secondary Ledger Changes I: Enable Secondary Ledgers for DARK1B/BRIGHT1B 

dylanagreen:
In accordance with doing the updates in #503 piecemeal, this PR addresses Update 1: The 1B primary ledgers will need to read the 1B secondary ledgers.

Originally it was expected that this would be handled automatically int he code. In practice this required some small minor edits, namely the following two changes:

- Check if `mydirs["scnd"]` exists as a file, and if not, use `mydirs["scnd2"]` as the "primary" secondary targets file.
  - To find the secondary target files the code scans the directories in the`$DESI_TARGET/catalogs/dr9/{dtver}/` directory. It takes the secondary targets saved in `$DESI_TARGET/catalogs/dr9/{dtver}/main` as the "base" file, and scans the tree for main* for additional files. 
  - Previously, the code checked all secondary target files for existence after the base one (i.e. if it finds `main2` it checks if the secondary file there exists before adding it to the list of files to scan later), but never checked the base original file. We now also check the first found file exists. This has no effect on DARK/BRIGHT tiles, for whom the base secondary targets file always exists in the original `main` dir. The main effect is for 1B ledgers, where the targets file will be added to `main4`. 
  - We could also add the secondary targets files to `main` instead, but this is more robust anyway, because we _should_ check that the secondary file in `main` actually exists.
- If no targets are loaded in `create_mtl` exit that function early and inform fba_launch that you did so. 
  - If no targets are loaded, then the `desitarget` function that actually creates the augmented mtl does not create a file at all: https://github.com/desihub/desitarget/blob/81b91948d1d30498d21c7178d931a2b79bd02ba2/py/desitarget/io.py#L614-L616
  - Fiberassign then dies when it tries to use shutil to move the expected augmented mtl file out of its temporary directory to a new name (`[tileid}-scnd.fits`): https://github.com/desihub/fiberassign/blob/b66d17fd96f5661c28af26e4d3606c8c73d551ad/py/fiberassign/fba_launch_io.py#L1708
  - Now we don't die, and explicitly tell `fba_launch` that we did not find any targets. This way we also prevent a crash later, when we would also try move `{tileid}-scnd.fits` out of the temporary directory to the output directory. 
  - It was never expected to load *0* targets from an input target file / mtl. In this case it does happen when you try and reproduce tiles from before enabling the secondary 1b ledgers, and all the 1b secondary targets would get ignored. 

With these changes, fiberassign can now correctly use secondary ledgers in the 1b program.

### DIFF
--- a/bin/fba_launch
+++ b/bin/fba_launch
@@ -329,7 +329,7 @@ def main():
                 for key in sorted(list(mydirs.keys())):
                     if (key[:4] == "scnd") & (key != "scnd") & (key != "scndmtl"):
                         targdirs.append(mydirs[key])
-                create_mtl(
+                status = create_mtl(
                     mytmpouts["tiles"],
                     mydirs["scndmtl"],
                     args.mtltime,
@@ -344,6 +344,10 @@ def main():
                     step="scnd",
                     start=start,
                 )
+                # DG - We exited early for some reason, most likely because
+                # there were no targets in the secondary MTL.
+                if status == 1:
+                    expected_files["scnd"] = False
         else:
             expected_files["scnd"] = False
             log.info("")
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -7,8 +7,10 @@ fiberassign change log
 ------------------
 
 * Don't load MTL* columns for pre-1B tiles in fba_plot. (PR `#502`_).
+* Enable running fiberassign with secondary ledgers for 1B programs. (PR `#504`_).
 
 .. _`#502`: https://github.com/desihub/fiberassign/pull/502
+.. _`#504`: https://github.com/desihub/fiberassign/pull/504
 
 5.9.0 (2026-03-17)
 ------------------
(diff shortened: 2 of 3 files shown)


dylanagreen:
Converting this to "ready for review" given the tests I've done over the past few days:

- scron runs every morning with `validate_lastnight_fba` pointing to this branch of fiberassign on my local copy. With all other aspects of the DESI environment remaining at the original/default values, fiberassign passes reproducibility. To zeroth order this means that even if we merge today and don't activate any 1B secondary ledgers, the changes in this PR will not break any reproducibility for tiles we currently have observed.
- I also tested 1B tiles using a custom SURVEYOPS location that includes 1B secondary ledgers, courtesy of @geordie666. In this test:
  - All 1B tiles I tested with dates BEFORE the "start date" of the 1B secondary targets correctly reproduce the fiberassignment (this means that fiberassign correctly ignored all those targets)
  - All 1B tiles with dates AFTER the start date of the 1B secondary targets FAIL reproducibility. This is the expected state, because it means that fiberassign correctly read in those targets, and used them in assignment. Reproducibility fails because we're including targets not included in the original survey assignments.

Given this I'm comfortable saying that the changes in this PR do what we expect/want them to do.

PR #505, opened 2026-04-27
Secondary Ledger Changes II: Enable Secondary Ledgers for BACKUP

dylanagreen:
Following #504, and addressing more aspects of #503 piecemeal, this PR addresses enables using secondary targets on BACKUP tiles. This was a simple change after enabling the 1B survey secondaries, since most of the work required for this was done there. This PR simply removes any blocking `if` statements that prevent searching for and loading secondary target files and MTLs on backup tiles. 

Testing is still in progress, but that said, these changes do not affect the reproducibility of the CURRENT tiles in the CURRENT DESI software stack (in other words, since the actual on disk secondary files/ledgers do not exist for BACKUP tiles, this PR does nothing to change any reproducibility of any tiles).

PR #506, opened 2026-05-07
Secondary Ledger Changes III: Read 1B secondary ledgers on 1A tiles

dylanagreen:
Following https://github.com/desihub/fiberassign/pull/505, and addressing the last aspects of https://github.com/desihub/fiberassign/issues/503. This PR branches off the aforementioned PR, and the aforementioned PR should be  merged _before_ this PR.

In this PR fiberassign now reads the 1b secondary ledgers when generating the 1a tile assignments, to preempt the possibility that we may add more 1a tiles but want to include the 1b secondary targets. This required:

- Return both 1a and 1b secondary MTL directories from `get_ledger_paths()`.
- Return the set of secondary targeting files for both 1a and 1b targets in `get_desitarget_paths()`.
- Account for the fact that `os.path.exists(dirs["scndmtl"])` might be false if the scndmtl is a list of two items, so instead check for the existence of both scndmtls, and keep any that exist and fail if none of them do.
- Account for > 2 targeting files with exactly 2 mtl directories when assembling the mtl to make the targeting file.
- Add a more robust check for if this is a secondary file or not in `create_mtl` to handle that `mtl_dir` might be a list. 
- Add a robust check to write the correct ledger paths to the output file, since scndmtl might be a list now.

I _think_ that this is everything necessary to do what we want. Testing is still in progress, and this PR will move to "ready to merge" when I'm confident it works as intended.

### DIFF
--- a/bin/fba_launch
+++ b/bin/fba_launch
@@ -316,15 +316,26 @@ def main():
     # AR if not backup
     if ("scnd" in steps):
         if ("scnd" in list(mydirs.keys())):
-            # AR handle dark1b/bright1b, where there is no secondary for now
-            # AR but could be in the future
-            # AR assume here that if there are some secondary, the ledgers should be there
-            if not os.path.isdir(mydirs["scndmtl"]):
+            # DG - For 1A targets we know load the 1b secondary mtls, so we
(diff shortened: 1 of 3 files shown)


dylanagreen:
Superseded by #509

PR #508, opened 2026-05-13
Read Static Target Files from Multiple Data Releases.

dylanagreen:
This PR adds capability to read static target files from multiple data releases. This is achieved by passing in a comma separated string of data releases. For example `--dr dr9,dr11`. This string is then split, and fiberassign checks for all relevant target and secondary target files in all input data releases. `fba_launch` restricts the inputs to be either `dr9` or `dr11` or some combination of the two. 

The changes in this PR are backwards compatible with all previous tiles. However, note the following:
- For sky and GFA targets, fiberassign will use specifically the **first** data release input, rather than loading both and concatenating them. So, if the input is `--dr dr9,dr11` then `fba_launch` will use the sky and GFA targets from `dr9`. 
- Internally, I've had to refactor and entirely rewrite how secondary target files are detected. This will break PRs #505 and #506, which both will have to be rewritten almost entirely.
- fba_launch now does a sanity file existence check on static target files, which it didn't before, but should preempt any future crashes. 
- I've made sure that the correct values of the `--dr` param are written to the fiberassign header so that `fba_rerun` correctly reruns the old tiles. This is necessary for `validate_lastnight_fba`. 

I've tested this on current dr9 only tiles, and fiberassign correctly reproduces the old tiles. I have yet to test any tiles that have dr11 targets, and I additionally have not tested if it reproduces old tiles if you accidentally pass in dr11 in addition to dr9. Further testing is required before this is ready to merge.

### DIFF
--- a/bin/fba_launch
+++ b/bin/fba_launch
@@ -316,15 +316,26 @@ def main():
     # AR if not backup
     if ("scnd" in steps):
         if ("scnd" in list(mydirs.keys())):
-            # AR handle dark1b/bright1b, where there is no secondary for now
-            # AR but could be in the future
-            # AR assume here that if there are some secondary, the ledgers should be there
-            if not os.path.isdir(mydirs["scndmtl"]):
+            # DG - For 1A targets we know load the 1b secondary mtls, so we
+            # need to check both scnd mtls, and then we only keep the ones that exist
+            # (in case the 1b ledgers aren't in place for some reason)
+            to_check = mydirs["scndmtl"]
+
+            if not isinstance(to_check, list):
+                to_check = [mydirs["scndmtl"]]
+
+            scnd_exists = []
+            for fname in to_check:
+                if os.path.isdir(fname):
+                    scnd_exists.append(fname)
+
+            if len(scnd_exists) == 0:
                 expected_files["scnd"] = False
                 log.info("")
                 log.info("")
                 log.info("{:.1f}s\tscnd\tno secondary ledgers {}".format(time() - start, mydirs["scndmtl"]))
             else:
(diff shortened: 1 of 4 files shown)


dylanagreen:
Updates since the last comment:

- fiberassign now (optionally) takes in multiple `--dtver` arguments. If only one is passed, that `--dtver` is used for all `--dr` values. If multiple, there must be exactly one `dtver` for each `dr`.
- Skies, GFAs and suppsky now all search all passed in `dtver` `dr` combinations and merges the target list. 

Still only tested on dr9 tiles at the moment, will test on dr11 tiles as soon as we have some MTLs to test with.

dylanagreen:
Updates since the last comment:

- fiberassign now (optionally) takes in multiple `--gaiadr` arguments. If only one is passed. it is required to be the same length as `--dr` (if you pass more than one `dr`, you must also pass more than one `gaiadr`.) We can relax this a little in the future if we want. 
- Skies, GFAs and suppsky now all search all passed in `dtver`+`dr` (sky, gfa) or `gaiadr`+`dtver` (suppsky) combinations and merges the target list. 

Testing locally on tiles with both dr9 and dr11 tiles succeed as expected.

dylanagreen:
Signing off. I'm comfortable with the results of my tests. Note that once this PR is merged, we **must** proceed with updating **both** MTLs and targeting files, as in previous versions of fiberassign if no targets end up being passed into assignment then fiberassign will fail. I'm comfortable with the tests I've run, but there could always be something we didn't anticipate in production. I hope not.

geordie666:
This looks good to me. I've been following along with the new changes in the background, and I think this is ready to merge.

PR #509, opened 2026-06-01
Secondary Ledger Changes II/III: Enable Secondary Ledgers for BACKUP, 1B Secondaries on 1A

dylanagreen:
Following https://github.com/desihub/fiberassign/pull/504, and addressing more aspects of https://github.com/desihub/fiberassign/issues/503 piecemeal, this PR addresses enables using secondary targets on BACKUP tiles, and also turns on 1B secondaries on 1A tiles. 

The following is a list of what was required to change under the hood to enable this:
   - Return both 1a and 1b secondary MTL directories from get_ledger_paths().
   - Return the set of secondary targeting files for both 1a and 1b targets in get_desitarget_paths().
   - Account for the fact that os.path.exists(dirs["scndmtl"]) might be false if the scndmtl is a list of two items, so instead check for the existence of both scndmtls, and keep any that exist and fail if none of them do.
   - Account for > 2 targeting files with exactly 2 mtl directories when assembling the mtl to make the targeting file.
   - Add a more robust check for if this is a secondary file or not in create_mtl to handle that mtl_dir might be a list.
   - Add a robust check to write the correct ledger paths to the output file, since scndmtl might be a list now.
   - Remove blocking statements that exclude BACKUP tiles.

This PR supersedes #505 and #506, combining both into a single PR that is compatible with ledger and target list changes in #508.

### DIFF
--- a/bin/fba_launch
+++ b/bin/fba_launch
@@ -316,15 +316,26 @@ def main():
     # AR if not backup
     if ("scnd" in steps):
         if ("scnd" in list(mydirs.keys())):
-            # AR handle dark1b/bright1b, where there is no secondary for now
-            # AR but could be in the future
-            # AR assume here that if there are some secondary, the ledgers should be there
-            if not os.path.isdir(mydirs["scndmtl"]):
+            # DG - For 1A targets we know load the 1b secondary mtls, so we
+            # need to check both scnd mtls, and then we only keep the ones that exist
+            # (in case the 1b ledgers aren't in place for some reason)
+            to_check = mydirs["scndmtl"]
+
+            if not isinstance(to_check, list):
+                to_check = [mydirs["scndmtl"]]
+
+            scnd_exists = []
+            for fname in to_check:
+                if os.path.isdir(fname):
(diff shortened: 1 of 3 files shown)


dylanagreen:
I tested this branch this morning on reproducing current main survey results from:
20250101: 40 tiles
20250202: 29 tiles
29250303: 11 tiles
20250415: 15 tiles
20250506: 33 tiles
20250606: 30 tiles
20250707: 2 tiles
20250808: 27 tiles
20250909: 29 tiles
20251013: 9 tiles
20251111: 20 tiles
20251212: 33 tiles

This is a total of 278 tiles. 

Some tiles (notably, the 2 tiles on 20250707 and 3 tiles on 20250909) filed reproducibility, but I have confirmed that these tiles also fail with `source /global/common/software/desi/desi_environment.sh main-2.2.0` so these failures are not exclusive to this PR. They should probably be investigated, and will be noted on a ticket to `desisurveyops` at a later date. 

The 273 other tiles succeed. Proceeding with merging.

PR #513, opened 2026-08-04
Script to patch wrong TARGETIDS in a calibration tile.

dylanagreen:
Pretty simple PR, just a script used to patch some incorrect TARGETIDs in a calibration tile using the mapping to the original DESI_TARGETID.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -59,10 +59,7 @@ jobs:
             - name: Install Python dependencies
               run: |
                 python -m pip install --no-cache-dir --upgrade pip setuptools wheel pytest
-                python -m pip install --no-cache-dir pyyaml requests scipy matplotlib healpy
-                python -m pip install --no-cache-dir --upgrade "numpy${{ matrix.numpy-version }}"
-                python -m pip install --no-cache-dir --upgrade "astropy${{ matrix.astropy-version }}"
-                python -m pip install --no-cache-dir fitsio
+                python -m pip install --no-cache-dir pyyaml requests scipy matplotlib healpy "numpy${{ matrix.numpy-version }}" "astropy${{ matrix.astropy-version }}" fitsio
                 python -m pip install desiutil==${DESIUTIL_VERSION}
                 python -m pip install --no-build-isolation --no-cache-dir git+https://github.com/desihub/desimodel.git@${{ env.DESIMODEL_VERSION }}#egg=desimodel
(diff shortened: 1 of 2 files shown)


sbailey:
Thanks. Please add a header comment to this file providing context including a link to desihub/desitarget#892 and an example commandline that was run at NERSC (e.g. because I don't know "where the tertiary data is stored")

dylanagreen:
Thanks @sbailey, I add the example call with a generic path to the fiberassign file and then the tertiary data I used at NERSC from the DESI environment (in $DESI_SURVEYOPS/tertiary).

forero:
Hi @dylanagreen. Thanks for doing this work.

As a test I copied the fiberassign file from `/global/cfs/cdirs/desi/target/fiberassign/tiles/trunk/083/` into a local directory to run the script. It ran without any problem.  

But then, I was left wondering how I could check that the script did the right thing. I tried running the script again on the patched file, hoping it would tell me there was nothing to patch, but the script ran again producing the same output on screen. So, I couldn't check the new file in that naive way.  Running the script twice actually repatches things, introducing new targetids again. 

So many questions for a simple script!  

- is there a way to check at the beginning of the script that the fiberassign file actually needs patching?
- is there a way you could add some new code at the end of the current script that checks that the patched file is actually what we wanted?

forero:
The bug that prompted this script (https://github.com/desihub/desitarget/issues/892) originated from RA/Dec checks. The checks on the unpatched/patched file could most probably also involve RA/Dec checks.

dylanagreen:
@forero I'm not sure that there's an easy way to check without loading the `ztile-special-bright` file or otherwise loading all three lower tiles (83060-83062). However, it is a good point that I should proivde some indication that a tile _was_ patched, so I will do what I did for the previous patching and add a comment to the FITS header. To zeroth order the script and check for existence of the comment and if so indicate it was already patched and don't patch a second time.

forero:
@dylanagreen  adding a comment to the FITS's header is good solution. Thanks!

sbailey:
With the advice of Dylan and the help of Claude, I verified that
* Unpatched 83063 has TARGETID with different RA,DEC than tiles 83060, 83061, 83062, 83064.
* Patched 83063 is now consistent with the others
* Only the TARGETID column changed, plus COMMENT header
* Attempting to repatch an already patched file will stop

I also added https://github.com/desihub/desitarget/issues/892 to the header comments for more context than just "TARGETIDs Patched on ..." (our future selves will be thankful).

